PR #1278, opened 2013-12-09
Implement the public user agent methods in runtime.

ShDavidliu:
The user agent is not supported in crosswalk runtime. This is
to implement the public user agent methods for crosswalk, it adds
the corresponding user agent description according to the specific
platform type.

 Based on ZTE975 Build/JDQ39, the user agent string is listed as follows:
 Mozilla/5.0 (Linux; Android 4.2.2; ZTE 975 Build/JDQ39) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/32.0.1700.14 Mobile Crosswalk/3.32.49.0 Mobile Safari/537.36

BUG=https://crosswalk-project.org/jira/browse/XWALK-595

### DIFF
--- a/runtime/common/xwalk_content_client.cc
+++ b/runtime/common/xwalk_content_client.cc
@@ -26,8 +26,12 @@ std::string XWalkContentClient::GetProduct() const {
 }
 
 std::string XWalkContentClient::GetUserAgent() const {
-  // TODO(hmin): Define user agent for xwalk.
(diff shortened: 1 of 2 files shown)


fujunwei:
Did the user agent need `'Mobile'`  before `'Crosswalk'`? It caused Sencha Touch did not determined correct device type on tablet devices because of the  `'Mobile'` (https://crosswalk-project.org/jira/browse/XWALK-4318).

PR #1487, opened 2014-01-15
[Tizen] Implement date / time chooser proxy

alexshalamov:
Date / Time chooser proxy is a DBus object proxy that communicates
with dialog launcher service.


darktears:
Let's close this.

PR #1849, opened 2014-04-22
[Android] make_apk.py fails when dir name same as app name

PeterWangIntel:
To create a folder with random name as intermedia result, when it happens that
the target folder we're going to copy files to has the same name of source
folder of those files.

BUG=https://crosswalk-project.org/jira/browse/XWALK-1378


lincsoon:
Since make_apk is not used anymore, close this issue

PR #1899, opened 2014-05-05
[Android] Add locale support based on the language of running Android.

sqliu:
This is the first step of enabling locale pak support, including:
(1) Add repack script and actions to repack locale paks (eg, zh-CN.pak)
(2) Enable loading locale pak based on system language, this will fallback
    to default locale packed in xwalk.pak when the locale pak is not provided.
(3) Enable XWalkCoreShell to support this. One same test is openning
     http://www.w3schools.com/html/tryit.asp?filename=tryhtml5_input_type_color
    and checking the webkit string ('submit') based on Android's language.

The change does not affect other components (runtime, library...), and supporting
for other components will be added in future patches.

BUG=https://crosswalk-project.org/jira/browse/XWALK-1632


lincsoon:
Close expired pull request

PR #1923, opened 2014-05-08
[Android] Add test cases for loadAppFromManifest().

wuhengzhi:
This patch is to add test cases for loadAppFromManifest().
It contains several URL schemes like 'http:', 'https:', 'file:'
and others.


kenchris:
what is up with this?

PR #1929, opened 2014-05-08
[Android] Add test cases for stopLoading().

wuhengzhi:
This patch is to add test cases for stopLoading().
Open an inexistent url, then execute stopLoading(), the load will be
stopped, title is empty.


kenchris:
What is up with this?

PR #2022, opened 2014-05-29
[Android] Add test case for getCurrentIndex().

jinbang:
This patch is to add test case for getCurrentIndex().
Load two urls, navigate to back url, get the current index and compare it with the expected result.
Navigate to forward url, get the current index and compare it with the expected result.


lincsoon:
Close obsolete pull request

PR #2277, opened 2014-08-04
[Android] If there is no 'version' feild from users, the packaging will ...

Harry110:
...exit

BUG=XWALK2008


lincsoon:
make_apk is not used anymore. Close this issue

PR #2389, opened 2014-09-04
[Android] User TextureView as the default rendering backend view.

sqliu:
There are several issues with SurfaceView rendering backend, including:
(1) Can't be animated/transformed/translucent.
(2) Multiple SurfaceView overlapped in the same ViewGroup will cause undefined behavior.

Those are not issues for Chrome for Android, since Chrome for Android uses only
one SurfaceView and everything is painted/compsited inside the SurfaceView.
But for Crosswalk, we may have multiple SurfaceViews for each embedding App, so the
drawbacks of SurfaceView may cause some bugs.

This CL makes TextureView as the default rendering backend, which can overcome the
drawbacks of SurfaceView. Eventhough the TextureView increase more memory usage.

To switch back to SurfaceView,set the ANIMATABLE_XWALK_VIEW preference value to be
false before using XWalkView.

BUG=

### DIFF
--- a/app/android/runtime_activity/src/org/xwalk/app/XWalkRuntimeActivityBase.java
+++ b/app/android/runtime_activity/src/org/xwalk/app/XWalkRuntimeActivityBase.java
@@ -33,6 +33,8 @@ public abstract class XWalkRuntimeActivityBase extends Activity {
 
     private boolean mRemoteDebugging = false;
 
+    private boolean mUseAnimatableView = false;
+
     private AlertDialog mLibraryNotFoundDialog = null;
 
     private XWalkRuntimeExtensionManager mExtensionManager;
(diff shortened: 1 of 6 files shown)


FrigoEU:
I've finally got my first XWalk application running, and the first thing I see is this terrible flashing-resizing of content whenever the keyboard opens/closes. I found the underlying issue here: https://crosswalk-project.org/jira/browse/XWALK-2012 , which points to this PR . 

Am I reading this correctly that the new default render target is TextureView, even though the warping that is happening because of it can't be fixed? Or am I missing something and I can actually make the TextureView stop doing this warping?

Hope this message is not coming over as complaining, I'm just having difficulty understanding that TextureView is accepted as the best render target even though the whole app gets warped every time the keyboard opens or closes. I must be doing something wrong!

baleboy:
@FrigoEU thanks for the feedback. Would you mind taking the discussion to crosswalk-help@lists.crosswalk-project.org ? That way more people will be able to participate. Thanks!

PR #2433, opened 2014-09-17
[Android] Fix XWalk extension implementation layer include voilation

js0701:
@hmin @wang16  @halton 
Extension system should avoid to include runtime headers
BUG=XWALK-1598
The Android Extension system and runtime part are including headers files from each other. Thus they are coupled with each other. The purpose for this commit is to decouple them by add a middle layer interface called XWalkExtensionManager


lincsoon:
@halton Can we close this pull request since it's obsolete?

halton:
Agree.

PR #2518, opened 2014-10-17
[Tizen] Implement Tizen privilege

xuzhang:
* Implement XWALK-1489: parse Tizen privilege item from config.xml
* When installing applicaiton, installer should set permissions into
  cynara db. Also installer should set SMACK label for resource files.
  When updating and removing applicaitons, the permissions need updated.
  This patch is picked from PR#2397.


rakuco:
Closing, Tizen is out of scope now.

PR #2566, opened 2014-11-04
[Android] Fix crash when send SMS with XWalkView in a Fragment.

shawngao5:
Root cause: Xwalk messaging extension filters all SMS_SENT intents, and
assume every intent to have an extra data, "instandid". In the hybrid app,
when developer sends an SMS_SENT intent without this extra data, the app
will get crashed.

Fix: Add a UUID to identify the intent sent by messaging extension.

BUG=XWALK-2709


sahar176:
FYI, I'm still having this issue on the latest stable crosswalk 10 build

PR #2707, opened 2014-12-06
Add support for file-picker using WebUI

joone:
BUG=XWALK-950

### DIFF
--- a/runtime/browser/ui/xwalk_web_dialog_view.cc
+++ b/runtime/browser/ui/xwalk_web_dialog_view.cc
@@ -1,0 +1,30 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/runtime/browser/ui/browser_dialogs.h"
+
+#include "ui/views/controls/webview/web_dialog_view.h"
+#include "ui/views/widget/widget.h"
+#include "xwalk/runtime/browser/ui/webui/xwalk_web_contents_handler.h"
+
+namespace xwalk {
+
+// Declared in browser_dialogs.h so that others don't need to depend on our .h.
+// Copied from chrome/browser/ui/views/chrome_web_dialog_view.cc
+gfx::NativeWindow ShowWebDialog(gfx::NativeWindow parent,
+                                content::BrowserContext* context,
+                                ui::WebDialogDelegate* delegate) {
+  views::WebDialogView* view =
+      new views::WebDialogView(context, delegate, new XWalkWebContentsHandler);
+
+  views::Widget* widget = (parent) ?
+      views::Widget::CreateWindowWithParent(view, parent) :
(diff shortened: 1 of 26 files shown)


PeterWangIntel:
@joone @kenchris @pozdnyakov 
I'm investigating how to implement File Picker for Crosswalk Linux and has found this patch. I have two questions:
(1) Are there some special reasons that we must implement File Picker by web UI?
     I'm not sure if it breaks our design rule to implement File Picker by libGTK.
(2) Do we still have plan to update this patch?
     Means, if we must implement File Picker by web UI, can I take this patch as base (or reference) to implement File Pciker for Crosswalk Linux?

joone:
@PeterWangIntel  I'm sorry for the late reply.

1) Because Tizen3.0 doesn't support Gtk+. If it is okay to use the Gtk+ file-picker in Crosswalk for  Linux desktop (e.g. Ubuntu, Fedora), you may be able to use Gtk+ file-picker.
2) Yes, I will update the patches. They landed in ozone-wayland so I'm planning to use the file-picker implemented in ozone-wayland.

Thanks!

PeterWangIntel:
@joone

PeterWangIntel:
@joone 
I've found the root reason of problem (1) I mentioned, look at my comment for "FilePickerWebDialog::OnDialogClosed"

joone:
@PeterWangIntel  @kenchris @pozdnyakov Please take a look at the latest CL. There is no dependency on Chromium side any more.

runtime/browser/ui/xwalk_web_dialog_view.cc
```diff
@@ -0,0 +1,33 @@
...
+#include "xwalk/runtime/browser/ui/webui/xwalk_web_contents_handler.h"
+
+namespace xwalk {
+
+// Declared in browser_dialogs.h so that others don't need to depend on our .h.
+// Copied from chrome/browser/ui/views/chrome_web_dialog_view.cc
+gfx::NativeWindow ShowWebDialog(gfx::NativeWindow parent,
+                                content::BrowserContext* context,
+                                ui::WebDialogDelegate* delegate) {
+  views::Widget* widget = NULL;
```
pozdnyakov: views::Widget* widget = (parent) ? views::Widget::CreateWindowWithParent(view, parent) : views::Widget::CreateWindow(view);

runtime/browser/ui/linux_webui/OWNERS
```diff
@@ -0,0 +1 @@
+joone.hur@intel.com
```
pozdnyakov: well, shouldn't it be proposed on Crosswalk owners ML?

pozdnyakov:
looks good overall

runtime/browser/ui/linux_webui/OWNERS
joone: I don't know, but we need the owners file as the project grows

runtime/browser/ui/xwalk_web_dialog_view.cc
joone: Nice comment!

joone:
@pozdnyakov  Can you commit this CL?

PR #2717, opened 2014-12-11
[Tizen] Drop the Shared process model

pozdnyakov:
This change drops the Shared process model. The app id is
obtained from the execution binary path (i.e. symbolic link).
The 'xwalk-launcher' and 'xwalkctl' utils are not build any more.
Dbus service files are not installed from the Crosswalk rpm file.
The code clean-up is not yet complete.

BUG=XWALK-3203


joone:
@pozdnyakov I tried to open a url with xwalk as follows:
$ xwalk http://google.com

It doesn't launch a webview.
It seems to be removed url handling in ApplicationSystemTizen::LaunchFromCommandLine.

Could you check this?

PR #2718, opened 2014-12-12
[Android] For shared mode, loading classes and resources seperatedly

wang16:
To get rid of flag Context.CONTEXT_IGNORE_SECURITY, load classes of
crosswalk library directly from the apk file. In addition, verify
the pubkey of that apk file before trying to load it.
For resources, still get them by createPackageContext. But now it
doesn't need flag for including code and ignore security check
anymore.

TODO(wang16): add the real comparison of pubkey after finalized the
keystore used to publish crosswalk library.

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
@@ -2,16 +2,53 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.xwalk.core.internal;
 
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.RandomAccessFile;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.security.cert.Certificate;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.HashMap;
+import java.util.Set;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.zip.ZipEntry;
+
+import dalvik.system.DexClassLoader;
 
 import android.content.Context;
 import android.content.pm.PackageManager;
+import android.content.pm.PackageManager.NameNotFoundException;
+import android.content.pm.Signature;
+import android.widget.Toast;
 
 /**
  * This class is used to encapsulate the reflection invoking for bridge and wrapper.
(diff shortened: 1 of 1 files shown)


wang16:
@terriko , I've fill the TODOs here to do additional checks against Android known exploits. Could you please help take a look?

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
```diff
@@ -2,16 +2,53 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+/*
```
terriko: Is this piece of code dual-licensed ASF and the BSD-style license in the LICENSE file, or was this license addition done in error?

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
```diff
@@ -117,6 +154,145 @@ static boolean shouldUseLibrary() {
...
+                rsaMatched = true;
+            }
+        }
+        apkFile.close();
+        // Crosswalk Library apk should only contain one certificate.
+        // Multiple signature introduces fake id risk.
+        if (rsaCount != 1) return false;
+        // For master key exploit 8219321, it's using multiple entries with same name to bypass system checking.
+        // If there is no duplicate entries, we are safe.
+        if (hasDuplicateEntries) return false;
```
terriko: Wouldn't any potentially problematic duplicate entries also be caught by the rsaCount?

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
wang16: I've leveraged several lines of code from aosp, so i put the license header here as well.

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
wang16: Are you talking about having duplicate entries of RSA files in the apk? If so, it will be considered tampered for sure. Either `rsaCount` or `hasDuplicateEntries` will detect it. But `rsaCount` go first, anyway it doesn't affect the result.

terriko: It's the fact that it doesn't affect the result that makes me wonder why it's being done.  it seems like this only time this would trigger is if there were two identically-named non-RSA files and one good RSA file.  Is there some sort of attack vector that uses that pattern, or is this code just being redundantly cautious?

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
terriko: You may need to check with our license lawyers to make sure this will be ok before it gets merged, then.

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
wang16: Duplicated entries are used to attack Android apk verification process. By duplicated entries, they have exactly same name (including path) in the zip file's structure. Because before some version of android, the code was using hashmap to keep entry info, so apparently there is exploit. So we are checking no duplicated entries here to prevent this type of attack.

For the RSA, it's a different story, one APK file could contains multiple RSAs with different name. In such case, there is risk to be attacked by FAKE ID (https://bluebox.com/technical/android-fake-id-vulnerability/) , so here we are checking that the target apk file only contains one RSA.

terriko: Yeah, I'm familiar with the attack, but I'm not convinced that checking for non-RSA duplicates is useful when we're already aborting if there's more than one RSA file, and we're only checking signatures in RSA files.  

Do we know for sure that there's no legit case in which there would be more than one RSA signature?  We were wondering if it's possible to have an apk legitimate signed by more than one source, such as a store where both the store and the application developer sign the apk.

wang16: The duplicates entries are not only limited to RSA files. And the check here for duplicated entries are not against RSA file either. It's prevent duplicate `classes.dex`, `libxwalkcore.so` and etc.

For legit multiple signatures, we can extends the implementation to multiple, just that we must know exactly how many RSAs will be there and the content of each of them.

terriko: aha, that makes much more sense.  thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
terriko: (Note for security reviewers coming to look at this code before it gets merged: Francesco is attempting to clear this with the appropriate person but has not heard back yet)

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
```diff
@@ -117,6 +154,145 @@ static boolean shouldUseLibrary() {
...
+
+    private static boolean checkFileNameLength(String apk) throws IOException {
+        RandomAccessFile raf = new RandomAccessFile(apk, "r");
+        // Following code are partly picked from AOSP.
+
+        // Scan back, looking for the End Of Central Directory field. If the zip file doesn't
+        // have an overall comment (unrelated to any per-entry comments), we'll hit the EOCD
+        // on the first try.
+        // No need to synchronize raf here -- we only do this when we first open the zip file.
+        long scanOffset = raf.length() - 22; // ENDHDR
```
pdxJayhawk: java.util.zip.ZipFile.ENDHDR

wang16: the purpose here is to read the zip file in the most "raw" way, to prevent any potential bugs in using utils.

pdxJayhawk: I don't think this justifies writing a novel parser.  Novel parsers generally have more bugs than established and mature utility libraries (of which java.util would qualify).  I strongly suggest using established and widely used helper classes wherever possible to avoid writing your own.

I think the best solution here is to trust the Android PackageManager and avoid this logic entirely (which will also solve the license concerns).

wang16: > I don't think this justifies writing a novel parser. Novel parsers generally have more bugs than established and mature utility libraries (of which java.util would qualify). I strongly suggest using established and widely used helper classes wherever possible to avoid writing your own.
> 
> I think the best solution here is to trust the Android PackageManager and avoid this logic entirely (which will also solve the license concerns).

The apk file is placing under system folder, which means is already passed Android PackageManager's security check. Here is some additional check against the bugs Android PackageManager have. So it doesn't need to even parse the contents out, the whole parsing is checking for some special check point to make sure the apk file is not using Android exploit to bypass Android security check.

wang16: BTW, the check here is to make sure the file name lengths for each entry is the same in central directory and local file ref. There is no such API in java.util at all to expose these details.
Well-known utils also has bugs, like Android PackageManager, If you think current parsing has exploit, just point it out.

terriko:
Current status: I don't think the security team can sign off on this as is.   The biggest problem right now is the amount of security-relevant code that will need to be thoroughly tested and maintained, as well as the fact that we're duplicating code from libraries and thus will have to maintain this code ourselves if future exploits are found.  The best way to resolve this and get this feature ready to release ASAP is to replace these checks is to replace this code with a much smaller set of checks that works on the assumption that the package manager can be trusted and uses the package manager to check the signature against our key.

terriko:
Here's a reworking to use PackageManager for the cert check:  https://github.com/crosswalk-project/crosswalk/pull/2860

baleboy:
I suppose we should abandon this PR, @lincsoon @terriko ?

terriko:
Yes, this can be safely abandoned.

PR #2733, opened 2014-12-16
[Crosswalk-Mini][Android] LZMA compression support

d0ngjun:
Use lzma compress libxwalkcore.so to reduce apk size.


halton:
> BTW, assets can't be used in Android Library project, you probably need to use res/raw instead.

@d0ngjun is working on this.

> use_lzma should not be a compile flag. Compressed and Uncompressed binaries should be both contained in packaging tool. 
For xwalk_core_library, we probably need to create xwalk_core_library_compressed besides existing xwalk_core_library.

Good suggestion and we ever think about that. But that would introduce several flags to use lzma or not:
* Test shell build time flag (current use_lzma):  for XWalkCoreShell and XWalkEmbeededClientShell 
* make_apk flag( to be added): for app developer. Default can be yes.
* Embedding API usage: for ant or Eclipse (not sure how to do that, but definitely can archive)

Clearlly, that will bring complexity. I doubt the value of output of this effort spending. Can we assume the CrosswalkMini customer only want to lzma currently. And report a feature request for future work?

wang16:
Another issue comes to me is that. We are not carrying native libraries anymore. Therefore, GooglePlay can't tell the APK is for arm or ia. We should think this through.

halton:
> Another issue comes to me is that, we are not carrying native libraries anymore, Therefore, GooglePlay can't tell the APK is for arm or ia. We should think this through.

Are you talking about the shared mode code that @lincsoon offered? If so, I'd like to resolve that in master branch, this branch only care about the embeeded mode and embedding api.

wang16:
> Are you talking about the shared mode code that @lincsoon offered? If so, I'd like to resolve that in master branch, this branch only care about the embeeded mode and embedding api.

Not about shared mode. For each app created with compressed libxwalkcore.so. It doesn't contain a native library from a normal application's perspective. Even if developers upload both apks for ARM and IA to PlayStore, playstore will take both apks as non-native-binary included apks. So it will always offer the one with the higher version code to end user regardless of CPU architecture match.

d0ngjun:
@halton @wang16 @lincsoon please help review, thanks.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,720 @@
...
+        @Override
+        protected Void doInBackground(Void... params) {
+            try {
+                XWalkCoreWrapper.decompressXWalkLibrary();
+                // TODO: use publishProgress to update percentage.
+            } catch (Exception e) {
+                Log.w("XWalkActivity", "decompress library failed: " + e.getMessage());
+            }
+
+            return null;
```
lincsoon: doInBackground should pass the result of decompressXWalkLibrary to onPostExecute by return statement

runtime/android/core_shell/src/org/xwalk/core/xwview/shell/XWalkViewShellActivity.java
```diff
@@ -359,4 +324,132 @@ public void onXWalkViewCreated(XWalkView view) {
...
+        XWalkViewShellActivity mActivity;
+
+        StartupDecompressDialog(XWalkViewShellActivity activity) {
+            super(activity);
+            mActivity = activity;
+
+            OnClickListener positiveListener = new OnClickListener() {
+                @Override
+                public void onClick(DialogInterface dialog, int id) {
+                    mActivity.finish();
```
lincsoon: The asyncTask is still running in background until completion even if the activity has been finished. It's necessary to call asyncTask.cancel(). And do not forget to check isCancelled() for every loop in doInBackground()

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,250 @@
...
+        Method method = null;
+        XWalkCoreStatus status = XWalkCoreStatus.NOT_FOUND;
+
+        try {
+            clazz = mBridgeLoader.loadClass(BRIDGE_PACKAGE + "." + XWALK_VIEW_DELEGATE);
+            method = clazz.getMethod("loadXWalkLibrary", Context.class);
+            method.invoke(null, xwalkApp);
+
+            status = XWalkCoreStatus.MATCHED;
+        } catch (UnsatisfiedLinkError | InvocationTargetException e) {
```
lincsoon: UnsatiffiedLinkError means native library is built for ARM but running on IA image. There's no need to decompress and should return false directly.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -79,24 +79,44 @@
...
-                System.load("/data/data/" + context.getPackageName() + "/lib/" + library);
+        if (sLibraryLoaded || (context == null)) return;
+
+        if (XWalkLibraryCompressed(context)) {
+            String lib = PathUtils.getDataDirectory(context.getApplicationContext());
+            if (lib != null) {
+                for (String library : MANDATORY_LIBRARIES) {
+                    System.load(lib + "/" + library);
+                }
+            }
```
lincsoon: If the so file is compressed, is it necessary to call System.load for each file? Why not just throw an exception?

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,720 @@
...
+    private XWalkCoreWrapper mCoreWrapper;
+    private LinkedList<Object> mReservedObjects;
+    private LinkedList<ReflectMethod> mReservedMethods;
+
+    public enum XWalkLibraryStatus {
+        MATCHED,
+        NOT_FOUND,
+        NEWER_VERSION,
+        OLDER_VERSION,
+        COMPRESSED,
```
lincsoon: The XWalkLibraryStatus filed means the result of jar version check. Jar version and whether native so is compressed are two different things. It's better to treat them differently

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,250 @@
...
+
+    public static XWalkCoreWrapper getInstance() {
+        return instance;
+    }
+
+    public static void initEmbeddedMode(XWalkCoreListener coreListener) {
+        reset(coreListener, null);
+        check();
+        init();
+    }
```
lincsoon: Why remove the assert statement

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
d0ngjun: according to http://developer.android.com/reference/java/lang/Runtime.html#loadLibrary(java.lang.String), seems 'not found' could trigger this exception also.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
d0ngjun: This means so files are compressed, so loadLibrary method should load the decompressed library from getDataDirectory.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: So how about letting loadXWalkLibrary throw a different exception if ARM library encounters IA image? What's your opinion @wang16 ?

wang16: OK to me.

xingnan:
Hi guys, do we have the data about decompression time on different devices?

halton:
@xingnan, welcome back. The decompression time various on different devices, but basically better CPU and more memory have small time. So far the HongMi is around ~20s, which is kind of not acceptable. We're improving it.

xingnan:
@halton Great, thanks for information :)

wang16:
Could you do a git rebase -i to merge the fixup into shared mode commit? LGTM after that.

d0ngjun:
@wang16 since lincsoon does not provide a single commit for this bug fix, rebase means I need to resolve a lot of conflict here, so I just use this fixup to remind myself resolve it later.

d0ngjun:
@wang16 please wait a few minutes for code rebase & test.

wang16:
Seems latest binary is still 13M with this commit? @halton

halton:
Yes, I aware of that, same with my local build. @d0ngjun is working on that.

halton:
Seems it is a buggy: it was using non-lzma libxwalkcore.so when first-run ninja. But when run ninja second time, it will use lzma libxwalkcore.so.

PR #2745, opened 2014-12-17
[Runtime] Enable to support service port in WARP mode rules

takethathe:
Since the Chromium/Blink Access White List can not support to check service port pattern
in access rule, so need to treat it in runtime. This patch add a checking of
service port in WARP mode.

BUG=XWALK-3132


takethathe:
@pozdnyakov 
I've changed the patch according to your comments :)

PR #2761, opened 2014-12-23
Revert "[Tizen] Refactor XWalkLauncher for usage of RESET event."

rakuco:
This reverts commit a32332dcc4aa34590e953d84535992e9b823c95a.

This change has several problems, some of which are evident when
Crosswalk is built in debug mode.
- Lack of a base::AtExitManager in the main function.
- The dbus::Bus is not properly shut down.
- Most importantly, the interaction between the D-Bus thread and the
  main thread is incorrect: XWalkLauncher and its dbus::Bus are created
  in the main thread, so many D-Bus methods end up being called there
  instead of in the IO thread. This causes many issues, including the
  crash reported in XWALK-3231.

Reverting this commit causes xwalk-launcher to work again, as otherwise
it would crash almost all the time.

BUG=XWALK-3231


tiwanek:
@rakuco 
Either way, this needs to be implemented for app-control and app debug mode. If not in xwalk-launcher, then in xwalk after dual process architecture introduction.

PR #2762, opened 2014-12-24
[Android] Fix app crash when two pages load each other for more than 512 times

fujunwei:
The crash is caused by local reference table overflow, there is  a thread hold the input stream reference by analysing MAT.
Threads attached through JNI must call DetachCurrentThread before they exit. So call DetachFromVM after AttachCurrentThread which wraps the DetachCurrentThread function.

BUG=XWALK-3096


baleboy:
This should have made it to Crosswalk 10 as well, but somehow we missed it. @fujunwei could you submit?

fujunwei:
I will merge it to Crosswalk-10.

PR #2766, opened 2014-12-26
Reenable SIMD.js support after the M40 rebase.

rakuco:
Commit 8c5d886 ("DEPS.xwalk: Roll v8-crosswalk (3bfc597 -> 11bb7b4)")
correctly rolled v8-crosswalk, but forgot to uncomment out the piece of
code that enables SIMD.js support by default.

BUG=XWALK-3210


huningxin:
@rakuco , LGTM. Thanks for this patch!

PR #2768, opened 2014-12-29
[Android] Fix file://*/*.xhtml will be opened in default browser.

junmin-zhu:
Because the minitype of xhtml file is "application/xhtml+xml".
It's intercepted by XWalkNavigationHandler to not handle in Crosswalk.
The default behavior for xhtml should be the same as html file, that
is loaded in the Crosswalk no matter with file schema or http schema.

BUG=XWALK-2912


wang16:
LGTM, thanks for the fix.

rakuco:
Please note that this was merged after crosswalk-11 was branched, so you need to cherry-pick it into -11 if necessary.

junmin-zhu:
@rakuco Thanks for reminder. It's available at https://github.com/crosswalk-project/crosswalk/pull/2779.

PR #2776, opened 2015-01-04
Add the check for return value of GetString().

wuhengzhi:
This patch is to add the check for return value of GetString() in
widget_handler.cc and xwalk_content.cc.

CID=220010

Related to XWALK-2928.
(cherry picked from commit 0bc7870d527aebed3fd6b8c5108f8e931094ec4c)


wuhengzhi:
@rakuco Please this patch to crosswalk-10, thanks.

PR #2777, opened 2015-01-04
[Android] Update project.properties for xwalk core library

wang16:
Require android-21 to compile to avoid build error.

BUG=https://crosswalk-project.org/jira/browse/XWALK-3238


wang16:
@fujunwei @lincsoon, PTAL, thanks.

fujunwei:
Thanks for your explain. LGTM, thanks.

PR #2778, opened 2015-01-05
[Runtime] Change the default start files priority

takethathe:
According to W3C Widget specific (http://www.w3.org/TR/widgets/), when a start
page is not given by config.xml, the UA should locate default start file by a
fixing order instead, which is given by "default start files
table" (http://www.w3.org/TR/widgets/#default-start-files-table) from top to
bottom. This patch will solve a mismatch with the spec.

BUG=XWALK-3264

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -39,8 +39,8 @@ namespace widget_keys = application_widget_keys;
 
 namespace {
 const char* kDefaultWidgetEntryPage[] = {
-"index.html",
 "index.htm",
+"index.html",
 "index.svg",
 "index.xhtml",
 "index.xht"};


takethathe:
@pozdnyakov 

rakuco:
The spec doesn't seem to really say that that table is supposed to be looked up in that order (it just seems to be alphabetically sorted), but I guess it doesn't hurt to do that.

takethathe:
@rakuco 
The spec http://www.w3.org/TR/widgets/#step-8-locate-the-start-file said: For each file name in the default start files table (from top to bottom) ..., if use this rule, it does have a fixing order to look up the default start page.

pozdnyakov:
lgtm

rakuco:
Ooops, I see.

Please backport to crosswalk-11 if necessary.

PR #2779, opened 2015-01-06
[Android] Fix file://*/*.xhtml will be opened in default browser.

junmin-zhu:
Because the mime type of xhtml file is "application/xhtml+xml".
It's intercepted by XWalkNavigationHandler to not handle in Crosswalk.
The default behavior for xhtml should be the same as html file, that
is loaded in the Crosswalk no matter with file schema or http schema.

This commit is backported from 6056a6f27

BUG=XWALK-2912


junmin-zhu:
@wang16 , @rakuco , this is backported  patch from 6056a6f. Thanks.

wang16:
LGTM, thanks.

PR #2780, opened 2015-01-07
[Android] Shared mode implementation

lincsoon:
Shared mode allows multiple xwalk apps to share one xwalk runtime
library. Each xwalk app is bundled with a thin reflection layer
instead of the whole runtime library, whereas there is a library
apk to be installed on device for xwalk apps to use.

BUG=XWALK-2948


runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
```diff
@@ -1,4 +1,3 @@
-// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
wang16: Mis operation?

runtime/android/core_shell/src/org/xwalk/core/xwview/shell/XWalkViewShellActivity.java
```diff
@@ -109,6 +110,7 @@ private void unregisterTracingReceiver() {
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        XWalkCoreWrapper.initEmbeddedMode();
```
wang16: I suggest to add a FIXME or TODO here to remove this unexpected call of internal methods.

tools/reflection_generator/java_method.py
```diff
@@ -731,20 +773,20 @@ def GenerateWrapperInterface(self):
...
-      return '%s%s%s%s' % (
-          '    private Method %s;\n' % self._method_declare_name,
+      return '%s\n%s\n%s\n%s\n' % (
           self.GenerateBridgeOverrideMethod(),
           self.GenerateBridgeWrapperMethod(),
-          self.GenerateBridgeSuperMethod())
+          self.GenerateBridgeSuperMethod(),
+          '    private ReflectMethod %s = \
+new ReflectMethod(null, null, "%s");\n' %
+              (self._method_declare_name, self._method_name))
```
wang16: The`ReflectMethod`s will be init in `ReflectionInit()` anyway, so here could be `private RefletionMethod %s = new ReflectionMethod();\n`?

tools/reflection_generator/wrapper_generator.py
```diff
@@ -173,64 +154,82 @@ def GenerateMethods(self):
...
-""")
-    value = {'REF_METHODS': ref_methods_string}
-    ref_init_string = ref_init_template.substitute(value)
-    return ref_init_string
-
-  def FormatStaticInitializer(self, method):
-    static_initializer_template = Template("""\
-        ReflectionHelper.registerConstructor(\"${CONSTRUCTOR_NAME}\", \
-\"${FULL_CLASS_NAME}\"${PARAM_LIST}, Object.class);
+    public void reflectionInit() {
```
wang16: `private`?

tools/reflection_generator/bridge_generator.py
```diff
@@ -21,146 +21,124 @@ def RunTask(self):
...
+                'PARAMS': method._bridge_params_declare_for_wrapper}
       ref_methods_string += ref_method_template.substitute(value)
 
     ref_init_template = Template("""\
-    private void reflectionInit() throws NoSuchMethodException,
-            ClassNotFoundException {
-        Class<?> clazz = wrapper.getClass();
-${REF_METHODS}
-    }
+    public void reflectionInit() {
```
wang16: `private`?

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectConstructor.java
```diff
@@ -0,0 +1,70 @@
...
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.concurrent.RejectedExecutionException;
+
+public class ReflectConstructor {
```
wang16: No need to be public

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectMethod.java
```diff
@@ -0,0 +1,116 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.concurrent.RejectedExecutionException;
+
+public class ReflectMethod {
```
wang16: no need to be public

tools/reflection_generator/java_method.py
lincsoon: The method call may be reserved if the reflection layer is not available yet. In this case, it's helpful that we can see the name of reserved method in log. So the method object is initialized with the method name but not just a null method

wang16: ok, thanks for the explanation.

lincsoon:
@wang16 @axinging This is a overall update for shared mode. Please help to review

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,625 @@
...
+        mIsShowing = false;
+    }
 
     @Override
     public Resources getResources() {
         return getApplicationContext().getResources();
     }
+
+    @Override
+    public void onObjectInitFailed(Object object) {
```
wang16: final

wang16: or default decorator in interface.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,625 @@
...
+
+    private void getXWalkLibrary() {
+        try {
+            Intent intent = new Intent(Intent.ACTION_VIEW);
+            startActivity(intent.setData(Uri.parse(XWALK_CORE_MARKET_URL)));
+
+            showDialog(getMarketProgressDialog());
+        } catch (ActivityNotFoundException e) {
+            Log.d("XWalkActivity", "Market open failed");
+            downloadXWalkLibrary();
```
wang16: Add a TODO to make it clear that need to at least edit the download url before final release.

runtime/android/core/src/org/xwalk/core/XWalkEmbeddedModeHelper.java
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core;
+
+public class XWalkEmbeddedModeHelper {
```
wang16: not public

tools/reflection_generator/reflection_generator.py
```diff
@@ -12,6 +12,7 @@
 from bridge_generator import BridgeGenerator
```
wang16: change the mod back?

halton:
@lincsoon please rebase your code and make trybot happy, then we can continue to review.

lincsoon:
I rebased shared mode implementation to 12.41.298.0. There will be no more major changes.
@halton @d0ngjun @axinging @fujunwei Please help to review

lincsoon:
@d0ngjun Please see XWalkLibraryDelegate and XWalkLibraryDelegateEmbedded. You can add a new class like XWalkLibraryDelegateLite or anything and change newInstance method in XWalkLibraryDelegate to let it return the instance of new class. At the end of initLibrary, do not forget to call mLibraryListener.onXWalkLibraryPrepared()

app/tools/android/make_apk.py
```diff
@@ -326,6 +326,12 @@ def Execution(options, app_info):
                 '--path', os.path.join(app_dir, 'xwalk_core_library'),
                 '--target', target_string])
     update_project_cmd.extend(['-l', 'xwalk_core_library'])
+  elif options.mode == 'shared':
+    print(' * Updating project with xwalk_shared_library')
+    RunCommand([android_path, 'update', 'lib-project',
+                '--path', os.path.join(app_dir, 'xwalk_shared_library'),
+                '--target', target_string])
+    update_project_cmd.extend(['-l', 'xwalk_shared_library'])
```
halton: Several places use the string `xwalk_shared_library`, could we define in above?

app/tools/android/make_apk.py
```diff
@@ -465,11 +471,10 @@ def MakeApk(options, app_info, manifest):
...
   if options.mode == 'shared':
-    # For shared mode, it's not necessary to use the whole xwalk core library,
-    # use xwalk_core_library_java_app_part.jar from it is enough.
-    java_app_part_jar = os.path.join(xwalk_dir, 'xwalk_core_library', 'libs',
-                                     'xwalk_core_library_java_app_part.jar')
-    shutil.copy(java_app_part_jar, os.path.join(app_dir, 'libs'))
+    # Copy xwalk_shared_library into app folder
+    target_library_path = os.path.join(app_dir, 'xwalk_shared_library')
+    shutil.copytree(os.path.join(xwalk_dir, 'xwalk_shared_library'),
+        target_library_path)
```
halton: I think pep8 will report wrong indent for this, please align all parameters

build/android/generate_resource_map.py
```diff
@@ -27,21 +27,35 @@ def CreateResourceMap(r_java, res_map):
...
+
+
 def main():
   parser = optparse.OptionParser()
   info = ('The folder contains generated R.java')
   parser.add_option('--gen-dir', help=info)
   info = ('The folder to place resource maps')
   parser.add_option('--resource-map-dir', help=info)
+  info = ('The file to touch on success.')
+  parser.add_option('--stamp', help=info)
```
halton: I'd like to rewrite above options as
```
parser.add_option('--gen-dir',
                  'The folder contains generated R.java')
parser.add_option('--resource-map-dir',
                  'The folder to place resource maps')
parser.add_option('--stamp',
                  'The file to touch on success.')
```

build/android/generate_resource_map.py
```diff
@@ -27,21 +27,35 @@ def CreateResourceMap(r_java, res_map):
...
+    os.makedirs(os.path.dirname(path))
+  with open(path, 'a'):
+    os.utime(path, None)
+
+
 def main():
   parser = optparse.OptionParser()
   info = ('The folder contains generated R.java')
   parser.add_option('--gen-dir', help=info)
   info = ('The folder to place resource maps')
```
halton: 'resource maps' -> 'resource map' if there no multiple map file.

build/android/generate_xwalk_core_library_aar.py
```diff
@@ -26,21 +26,18 @@ def main():
...
+        'R.txt'), 'R.txt'),
   )
   # This is a list of files that will not be packaged: mostly a blacklist of
   # files within |dirs|.
   exclude_files = (
     os.path.join(options.target, 'xwalk_core_library', 'libs',
-                 'xwalk_core_library_java_app_part.jar'),
-    os.path.join(options.target, 'xwalk_core_library', 'libs',
-                 'xwalk_core_library_java_library_part.jar'),
+        'xwalk_core_library_java.jar'),
```
halton: ditto, ally parameters.

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2013 Intel Corporation. All rights reserved.
```
halton: 2015

build/android/generate_resource_map.py
```diff
@@ -27,21 +27,35 @@ def CreateResourceMap(r_java, res_map):
...
   if not os.path.isdir(options.gen_dir):
     return 1
   for root, _, files in os.walk(options.gen_dir):
     if 'R.java' in files:
+      if os.path.basename(root) == 'core':
+        continue
       r_java = os.path.join(root, 'R.java')
       CreateResourceMap(r_java, options.resource_map_dir)
 
+  if options.stamp:
```
halton: `if options.stamp is not None`

build/android/generate_resource_map.py
```diff
@@ -27,21 +27,35 @@ def CreateResourceMap(r_java, res_map):
     output.write(''.join(output_content))
 
 
+def Touch(path):
```
halton: `Stamp`

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def CleanLibraryProject(out_project_dir):
+  if os.path.exists(out_project_dir):
+    for item in os.listdir(out_project_dir):
+      sub_path = os.path.join(out_project_dir, item)
+      if os.path.isdir(sub_path):
+        shutil.rmtree(sub_path)
+      elif os.path.isfile(sub_path):
+        os.remove(sub_path)
```
halton:
```
if not os.path.exists(out_project_dir):
  return

for item in os.listdir(out_project_dir):
...
```

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+def CopyProjectFiles(project_source, out_project_dir):
+  """cp xwalk/build/android/xwalkcore_library_template/<file>
+        out/Release/xwalk_core_library/<file>
+  """
+
+  print 'Copying library project files...'
```
halton: `print()`

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+  ]
+  for f in files_to_copy:
+    source_file = os.path.join(template_dir, f)
+    target_file = os.path.join(out_project_dir, f)
+
+    shutil.copy2(source_file, target_file)
+
+
+def CopyBinaries(out_dir, out_project_dir, src_package):
+  print 'Copying binaries...'
```
halton: ditto

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+    os.mkdir(libs_dir)
+
+  # Copy jar files to libs.
+  if src_package:
+    libs_to_copy = [
+    ]
+  else:
+    libs_to_copy = [
+        'xwalk_core_library_java_app_part.jar',
+    ]
```
halton:
```
libs_to_copy = []
if not src_package:
  libs_to_copy.extend('xwalk_core_library_java_app_part.jar')
```

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+
+  for lib in libs_to_copy:
+    source_file = os.path.join(out_dir, 'lib.java', lib)
+    target_file = os.path.join(libs_dir, lib)
+    shutil.copyfile(source_file, target_file)
```
halton: I would rewrite as 
```
shutil.copyfile(os.path.join(out_dir, 'lib.java', lib),
                os.path.join(libs_dir, lib))
```

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+      relative_path = os.path.relpath(src_file, input_dir)
+      target_file = os.path.join(output_dir, relative_path)
+      target_dir_name = os.path.dirname(target_file)
+      if not os.path.exists(target_dir_name):
+        os.makedirs(target_dir_name)
+      # If the file exists, copy it and rename it with another name to
+      # avoid overwriting the existing one.
+      if os.path.exists(target_file):
+        target_base_name = os.path.basename(target_file)
+        target_base_name = prefix + '_' + target_base_name
```
halton: `target_base_name = prefix + '_' + os.path.basename(target_file)`

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+        target_file = os.path.join(target_dir_name, target_base_name)
+      shutil.copyfile(src_file, target_file)
+
+
+def MoveImagesToNonMdpiFolders(res_root):
+  """Move images from drawable-*-mdpi-* folders to drawable-* folders.
+
+  Why? http://crbug.com/289843
+
```
halton: nit, remove the blank lines

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+    if not os.path.isdir(src_dir):
+      continue
+    dst_components = [c for c in src_components if c != 'mdpi']
+    assert dst_components != src_components
+    dst_dir_name = '-'.join(dst_components)
+    dst_dir = os.path.join(res_root, dst_dir_name)
+    if not os.path.isdir(dst_dir):
+      os.makedirs(dst_dir)
+    for src_file_name in os.listdir(src_dir):
+      if not src_file_name.endswith('.png'):
```
halton: Only png files supported?

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+  ]
+
+  pathname = os.path.basename(filepath)
+  #replace crunched 9-patch image resources.
+  for search in search_dir:
+    absdir = os.path.join(project_source, search)
+    for dirname, _, files in os.walk(absdir):
+      if filename in files:
+        relativedir = os.path.basename(dirname)
+        if (pathname == 'drawable' and relativedir == 'drawable-mdpi') or \
```
halton: nit, is `\` really needed?

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+            relativedir == pathname:
+          source_file = os.path.abspath(os.path.join(dirname, filename))
+          target_file = os.path.join(filepath, filename)
+          shutil.copyfile(source_file, target_file)
+          return
+
+
+def CopyResources(project_source, out_dir, out_project_dir):
+  print 'Copying resources...'
```
halton: nit

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+    os.makedirs(subdir)
+    with zipfile.ZipFile(zip_file) as z:
+      z.extractall(path=subdir)
+    CopyDirAndPrefixDuplicates(subdir, res_dir, zip_name)
+    MoveImagesToNonMdpiFolders(res_dir)
+
+  if os.path.isdir(temp_dir):
+    shutil.rmtree(temp_dir)
+
+  #search 9-patch, then replace it with uncrunch image.
```
halton: nit, space after `#` and capital first letter.

And what does .9.png mean, any document to show?

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+        ReplaceCrunchedImage(project_source, filename, dirname)
+
+
+def main(argv):
+  option_parser = optparse.OptionParser()
+  AddGeneratorOptions(option_parser)
+  options, _ = option_parser.parse_args(argv)
+
+  if not os.path.exists(options.source):
+    print 'Source project does not exist, please provide correct directory.'
```
halton: nit

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+
+  # Copy Eclipse project files of library project.
+  CopyProjectFiles(options.source, out_project_dir)
+  # Copy binaries and resuorces.
+  CopyResources(options.source, out_dir, out_project_dir)
+  CopyBinaries(out_dir, out_project_dir, options.src_package)
+  # Remove unused files.
+  mode = os.path.basename(os.path.normpath(out_dir))
+  RemoveUnusedFilesInReleaseMode(mode,
+      os.path.join(out_project_dir, 'libs'))
```
halton:
```
RemoveUnusedFilesInReleaseMode(
    os.path.basename(os.path.normpath(out_dir)),
    os.path.join(out_project_dir, 'libs'))
```

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+  # Create empty src directory
+  src_dir = os.path.join(out_project_dir, 'src')
+  if not os.path.isdir(src_dir):
+    os.mkdir(src_dir)
```
halton: nit, line break after

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,251 @@
...
+  readme = os.path.join(src_dir, 'README.md')
+  open(readme, 'w').write(
+      "# Source folder for xwalk_shared_library\n"
+      "## Why it's empty\n"
+      "xwalk_shared_library doesn't contain java sources.\n"
+      "## Why put me here\n"
+      "To make archives keep the folder, "
+      "the src directory is needed to build an apk by ant.")
+  print 'Your Android library project has been created at %s' % (
```
halton: print(

build/android/generate_xwalk_shared_library_aar.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

build/android/xwalk_shared_library_template/ant.properties
```diff
@@ -0,0 +1,17 @@
...
+# You can use this to override default values such as
+#  'source.dir' for the location of your java source folder and
+#  'out.dir' for the location of your output folder.
+
+# You can also use it define how the release builds are signed by declaring
+# the following properties:
+#  'key.store' for the location of your keystore and
+#  'key.alias' for the name of the key to use.
+# The password will be asked during the build when you use the 'release' target.
+
```
halton: Remove this empty line at end file.

Does this file really workable, all are comments?

build/android/xwalk_shared_library_template/build.xml
```diff
@@ -0,0 +1,29 @@
...
+	<fail
+		message="sdk.dir is missing. Make sure to generate local.properties using 'android update project' or to inject it through the ANDROID_HOME environment variable."
+		unless="sdk.dir"
+	/>
+
+	<import file="custom_rules.xml" optional="true" />
+
+	<!-- version-tag: custom -->
+	<import file="${sdk.dir}/tools/ant/build.xml" />
+</project>
```
halton: Can we use 4-space indent as AndroidManifest.xml?

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
  */
-public abstract class XWalkActivity extends Activity {
+public abstract class XWalkActivity extends Activity implements XWalkLibraryListener {
+    private static final String XWALK_CORE_APK = "XWalkRuntimeLib.apk";
+
+    private static final String XWALK_CORE_MARKET_URL =
+            "market://details?id=" + XWalkCoreWrapper.XWALK_APK_PACKAGE;
+
+    private static final String XWALK_CORE_DOWNLOAD_URL =
+            "http://10.0.2.2/" + XWALK_CORE_APK; // For test only
```
halton: If is test only, can we use more accurate var name like `XWALK_CORE_TESTING_DOWNLOAD_URL`?

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+    }
 
     @Override
     public Resources getResources() {
         return getApplicationContext().getResources();
     }
+
+    @Override
+    public final void onObjectInitFailed(Object object) {
+        Log.d("XWalkActivity", "Reserve object: " + object.getClass());
```
halton: define a TAG private final member to use. Please do all places.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+    public final void onXWalkLibraryMatched() {
+        Log.d("XWalkActivity", "XWalk library matched");
+        XWalkCoreWrapper.init();
+        mCoreWrapper = XWalkCoreWrapper.getInstance();
+        mLibraryDelegate = XWalkLibraryDelegate.newInstance(this, this);
+
+        mIsInitializingLibrary = true;
+        if (!mLibraryDelegate.initLibrary()) {
+            Assert.fail("XWalk library initialization failed");
+        }
```
halton: Should release build keep running when initLibary() failed? If not the case, please do not use Assert

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+                method = mReservedMethods.poll()) {
+            Log.d("XWalkActivity", "Call reserved methods: " + method);
+            Object[] args = method.getArguments();
+            if (args != null) {
+                for (int i = 0; i < args.length; ++i) {
+                    if (args[i] instanceof ReflectMethod) {
+                        args[i] = ((ReflectMethod) args[i]).invokeWithArguments();
+                    }
+                }
+            }
```
halton:
```
if (args == null)
    continue;

for (int i = 0; i < args.length; ++i) {
...
```

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+
+    private void getXWalkLibrary() {
+        try {
+            Intent intent = new Intent(Intent.ACTION_VIEW);
+            startActivity(intent.setData(Uri.parse(XWALK_CORE_MARKET_URL)));
+            showDialog(getMarketProgressDialog());
+        } catch (ActivityNotFoundException e) {
+            Log.d("XWalkActivity", "Market open failed");
+            showDialog(getMarketOpenFailedDialog());
+            //downloadXWalkLibrary();
```
halton: remove this commentted out code

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+
+    private static class DownloadTask extends AsyncTask<Void, Integer, Integer> {
+        private static final int QUERY_INTERVAL_MS = 100;
+        private static final int MAX_PAUSED_COUNT = 6000; // 10 minutes
+        private static final int MAX_RUNNING_COUNT = 18000; // 30 minutes
+
+        XWalkActivity mXWalkActivity;
+        DownloadManager mDownloadManager;
+        long mDownloadId;
+        ProgressDialog mProgressDialog;
```
halton: public/private/protected modifiers for above members.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+                    int status = cursor.getInt(statusIdx);
+                    if (status == DownloadManager.STATUS_FAILED ||
+                            status == DownloadManager.STATUS_SUCCESSFUL) {
+                        return status;
+                    } else if (status == DownloadManager.STATUS_PAUSED) {
+                        if (++pausedCount == MAX_PAUSED_COUNT) return status;
+                    } else if (status == DownloadManager.STATUS_RUNNING) {
+                        if (++runningCount == MAX_RUNNING_COUNT) return status;
+                    }
+                }
```
halton:
```
if (cursor == null || !cursor.moveToFirst()) continue;

int totalIdx = cursor.getColumnIndex(DownloadManager.COLUMN_TOTAL_SIZE_BYTES);
...
```

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+                } else if (result == DownloadManager.STATUS_RUNNING) {
+                    errMsg = mXWalkActivity.getString(R.string.download_failed_time_out);
+                }
+
+                Log.d("XWalkActivity", "Download failed, " + errMsg);
+                AlertDialog dialog = mXWalkActivity.getDownloadFailedDialog();
+                dialog.setMessage(errMsg);
+                mXWalkActivity.showDialog(dialog);
+            }
+        }
```
halton:
```
if (result == DownloadManager.STATUS_SUCCESSFUL) {
    ...
    return;
}

String errMsg = mXWalkActivity.getString(R.string.download_failed_message);
...
```

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+                        if (reason == DownloadManager.ERROR_DEVICE_NOT_FOUND) {
+                            errMsg = mXWalkActivity.getString(
+                                    R.string.download_failed_device_not_found) ;
+                        } else if (reason == DownloadManager.ERROR_INSUFFICIENT_SPACE) {
+                            errMsg = mXWalkActivity.getString(
+                                    R.string.download_failed_insufficient_space);
+                        }
+                    }
+                } else if (result == DownloadManager.STATUS_PAUSED) {
+                } else if (result == DownloadManager.STATUS_RUNNING) {
```
halton:
```
} else if (result == DownloadManager.STATUS_PAUSED ||
           result == DownloadManager.STATUS_RUNNING) {
```

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+
+    private AlertDialog getStartupNotFoundDialog() {
+        AlertDialog dialog = buildAlertDialog();
+
+        OnClickListener positiveListener = new OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int id) {
+                getXWalkLibrary();
+            }
+        };
```
halton: nit, line break after

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+
+        dialog.setIcon(android.R.drawable.ic_dialog_alert);
+        dialog.setTitle(getString(R.string.startup_not_found_title));
+        dialog.setMessage(getString(R.string.startup_not_found_message));
+        dialog.setButton(DialogInterface.BUTTON_POSITIVE,
+            getString(R.string.get_crosswalk), positiveListener);
+        dialog.setButton(DialogInterface.BUTTON_NEGATIVE,
+            getString(R.string.xwalk_cancel), negativeListener);
```
halton: 8-space indent

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+            public void onClick(DialogInterface dialog, int id) {
+                onXWalkLibraryCancelled();
+            }
+        };
+
+        dialog.setIcon(android.R.drawable.ic_dialog_alert);
+        dialog.setTitle(getString(R.string.startup_not_found_title));
+        dialog.setMessage(getString(R.string.startup_not_found_message));
+        dialog.setButton(DialogInterface.BUTTON_POSITIVE,
+            getString(R.string.get_crosswalk), positiveListener);
```
halton: nit, 8-space indent

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+            }
+        };
+
+        dialog.setIcon(android.R.drawable.ic_dialog_alert);
+        dialog.setTitle(getString(R.string.startup_older_version_title));
+        dialog.setMessage(getString(R.string.startup_older_version_message));
+        dialog.setButton(DialogInterface.BUTTON_POSITIVE,
+            getString(R.string.get_crosswalk), positiveListener);
+        dialog.setButton(DialogInterface.BUTTON_NEGATIVE,
+            getString(R.string.xwalk_cancel), negativeListener);
```
halton: indent

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+            public void onClick(DialogInterface dialog, int id) {
+                onXWalkLibraryCancelled();
+            }
+        };
+
+        dialog.setIcon(android.R.drawable.ic_dialog_alert);
+        dialog.setTitle(getString(R.string.startup_newer_version_title));
+        dialog.setMessage(getString(R.string.startup_newer_version_message));
+        dialog.setButton(DialogInterface.BUTTON_POSITIVE,
+            getString(R.string.xwalk_cancel), positiveListener);
```
halton: indent

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,560 @@
...
+            public void onClick(DialogInterface dialog, int id) {
+                onXWalkLibraryCancelled();
+            }
+        };
+
+        dialog.setIcon(android.R.drawable.ic_dialog_alert);
+        dialog.setTitle(getString(R.string.market_open_failed_title));
+        dialog.setMessage(getString(R.string.market_open_failed_message));
+        dialog.setButton(DialogInterface.BUTTON_POSITIVE,
+            getString(R.string.xwalk_cancel), positiveListener);
```
halton: indent

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+        Assert.assertNull(sInstance);
+        sProvisionalInstance = new XWalkCoreWrapper(-1);
+        return sProvisionalInstance.findEmbeddedCore();
+    }
+
+    public static void check() {
+        check(-1);
+    }
+
+    public static void check(int minSdkVersion) {
```
halton: Difference between `check()` and `check(int)` functions.  I suppose `check(int)` is the one for embedder to call.  Why are they all public functions? Am i miss something?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+            sListener.onXWalkLibraryMatched();
+        } else {
+            Error error = null;
+            if (status == LibraryStatus.NOT_FOUND) {
+                error = new UnsatisfiedLinkError("XWalk Core Not Found");
+            } else {
+                error = new VerifyError("API Incompatible");
+            }
+            sListener.onXWalkLibraryStartupError(status, error);
+        }
```
halton:
```
if (cond) {
    ....
    return;
}

// else code without indent

File: runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java [hunk 56]
```diff
@@ -0,0 +1,251 @@
...
+        LibraryStatus status = sProvisionalInstance.mCoreStatus;
+        if (status == LibraryStatus.MATCHED) {
+            sListener.onXWalkLibraryMatched();
+        } else {
+            Error error = null;
+            if (status == LibraryStatus.NOT_FOUND) {
+                error = new UnsatisfiedLinkError("XWalk Core Not Found");
+            } else {
+                error = new VerifyError("API Incompatible");
+            }
```
halton:
```
Error error = (status == LibraryStatus.NOT_FOUND) ?
               new UnsatisfiedLinkError("XWalk Core Not Found") : new VerifyError("API Incompatible");
```

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+
+    private XWalkCoreWrapper(int minSdkVersion) {
+        mSdkVersion = XWalkSdkVersion.SDK_VERSION;
+        if (minSdkVersion > 0 && minSdkVersion <= mSdkVersion) {
+            mMinSdkVersion = minSdkVersion;
+        } else {
+            mMinSdkVersion = mSdkVersion;
+        }
```
halton: `mMinSdkVersion = (minSdkVersion > 0 && minSdkVersion <= mSdkVersion) ? minSdkVersion : mSdkVersion;`

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+
+        return mCoreStatus != LibraryStatus.NOT_FOUND;
+    }
+
+    private boolean checkCoreArchitecture() {
+        try {
+            Class<?> clazz = getBridgeClass("XWalkCoreVersion");
+            String targetArch = (String) new ReflectField(null, clazz, "TARGET_ARCH").get();
+            boolean targetIsIA = (targetArch == "ia32");
```
halton: How about 64bit support?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    public boolean isSharedMode() {
+        return mBridgeContext != null;
+    }
+
+    public Object getBridgeObject(Object object) {
+        try {
+            return new ReflectMethod(null, object, "getBridge").invoke();
+        } catch (RuntimeException e) {
```
halton: logging?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+        return null;
+    }
+
+    public Object getWrapperObject(Object object) {
+        try {
+            return new ReflectMethod(null, object, "getWrapper").invoke();
+        } catch (RuntimeException e) {
```
halton: ditto

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -0,0 +1,251 @@
...
+        }
+        return null;
+    }
+
+    public Class<?> getBridgeClass(String name) {
+        try {
+            return mBridgeLoader.loadClass(BRIDGE_PACKAGE + "." + name);
+        } catch (ClassNotFoundException e) {
```
halton: ditto

runtime/android/core/src/org/xwalk/core/XWalkLibraryDelegate.java
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core/src/org/xwalk/core/XWalkLibraryDelegate.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    protected XWalkLibraryDelegate(Context context, XWalkLibraryListener listener) {
+        mContext = context;
+        mLibraryListener = listener;
+    }
+
+    public boolean initLibrary() {
+        if (mLibraryListener != null) mLibraryListener.onXWalkLibraryPrepared();
+        return true;
```
halton: no false condition?

runtime/android/core/src/org/xwalk/core/XWalkLibraryDelegateEmbedded.java
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core/src/org/xwalk/core/XWalkLibraryListener.java
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core/src/org/xwalk/core/XWalkReflectionHelper.java
```diff
@@ -0,0 +1,28 @@
+// Use of this source code is governed by a BSD-style license that can be
```
halton: copyright is missing

runtime/android/core/strings/xwalk_app_strings.grd
```diff
@@ -0,0 +1,85 @@
...
+      </message>
+    </messages>
+  </release>
+  <translations>
+  </translations>
+  <outputs>
+    <output filename="values/xwalk_app_strings.xml" lang="en" type="android" />
+  </outputs>
+</grit>
+
```
halton: remove this line

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectConstructor.java
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectField.java
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectMethod.java
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCoreBridge.java
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

tools/reflection_generator/reflection_generator.py
```diff
@@ -27,76 +28,74 @@
   'XWalkJavascriptResultInternal',
 ]
 
+REFLECTION_HERLPER = [
+  'ReflectMethod.java',
+  'ReflectField.java',
+  'ReflectConstructor.java',
+  'ReflectExceptionHandler.java',
```
halton: 4-space indent to follow pep8

tools/reflection_generator/reflection_generator.py
```diff
@@ -27,76 +28,74 @@
...
-  if not os.path.isdir(os.path.dirname(output)):
-    os.makedirs(os.path.dirname(output))
-  fo = open(output, 'w')
-  for line in f.read().split('\n'):
-    if line.startswith('package '):
-      fo.write('package org.xwalk.core;\n')
+def GenerateJavaBindingClass(input_dir):
+  class_loader = JavaClassLoader(input_dir, CLASSES_TO_BE_PROCESS)
+  for input_class in CLASSES_TO_BE_PROCESS:
+    print 'Generate bridge and wrapper code for %s' % input_class
```
halton: print(

tools/reflection_generator/reflection_generator.py
```diff
@@ -108,37 +107,51 @@ def Touch(path):
...
-                                 'is placed.'))
-  option_parser.add_option('--wrap_output',
-                           help=('Output directory where the wrap code'
-                                'is placed.'))
-  option_parser.add_option('--helper_class',
-                           help=('the path of ReflectionHelper java source, '
-                                'will copy it to output folder'))
+      help=('Output directory where the bridge code is placed.'))
+  option_parser.add_option('--wrapper_output',
+      help=('Output directory where the wrap code is placed.'))
```
halton: align parameters to follow pep8

tools/reflection_generator/reflection_generator.py
```diff
@@ -108,37 +107,51 @@ def Touch(path):
...
   option_parser.add_option('--stamp', help='the file to touch on success.')
+  option_parser.add_option('--target_arch', help='Target Architecture')
+  option_parser.add_option('--sdk_version', help='API Version')
+  option_parser.add_option('--min_sdk_version', help='Min API Version')
+
   options, _ = option_parser.parse_args(argv)
-  if not options.input_dir:
-    print('Error: Must specify input.')
+  if (not options.input_dir or not options.bridge_output or
+      not options.wrapper_output):
```
halton:
```
 if (not options.input_dir or
     not options.bridge_output or
     not options.wrapper_output):
```

tools/reflection_generator/reflection_generator.py
```diff
@@ -108,37 +107,51 @@ def Touch(path):
...
   if os.path.isdir(options.bridge_output):
     shutil.rmtree(options.bridge_output)
-  if os.path.isdir(options.wrap_output):
-    shutil.rmtree(options.wrap_output)
+  if os.path.isdir(options.wrapper_output):
+    shutil.rmtree(options.wrapper_output)
+
+  global bridge_path
+  bridge_path = os.path.join(options.bridge_output,
+      os.path.sep.join(BRIDGE_PACKAGE.split('.')))
```
halton: align para

tools/reflection_generator/reflection_generator.py
```diff
@@ -108,37 +107,51 @@ def Touch(path):
...
+  os.makedirs(bridge_path)
+
+  global wrapper_path
+  wrapper_path = os.path.join(options.wrapper_output,
+      os.path.sep.join(WRAPPER_PACKAGE.split('.')))
```
halton: ditto

xwalk_android.gypi
```diff
@@ -43,14 +44,35 @@
         ],
     },
     {
+      'target_name': 'xwalk_app_strings',
+      'type': 'none',
+      'variables': {
+          'grd_file': '../xwalk/runtime/android/core/strings/xwalk_app_strings.grd',
```
halton: indent

halton:
@lincsoon I finished my first-round review, please ping me when you finish update.

build/android/generate_xwalk_core_library_aar.py
fujunwei: The aar only contain xwalk_core_library_java.jar, developer can't use the aar for embedded mode, so we need generate another aar for shared mode, and push it to company maven repository with classify.

lincsoon: xwalk_core_library_java.jar is the combination of app_part and library_part.

build/android/generate_xwalk_shared_library.py
lincsoon: From the calling logic, the prefix must be a valid file name.

build/android/generate_xwalk_shared_library.py
lincsoon: The code around here is just copied from generate_xwalk_core_library. It's better to keep compliance with the core library if no specific needs

halton: Fine for me, but keep in mind. We need figure the situation, we potentially will use webp for Crosswalk Lite.

build/android/generate_xwalk_shared_library.py
lincsoon: Ditto, the code around here is just copied from generate_xwalk_core_library, I need more investigation to know the purpose

build/android/xwalk_shared_library_template/ant.properties
lincsoon: Just add here for developer's later use

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: Currently, it's test only. But it's also a feature we will enable in the future. We plan to support 2 ways to get library: Google Play and direct download.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: initLibrary() failed means the native SO library is not available. This is not allowed in any case.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: We need to call "method.invokeWithArguments()" later even if the args is null. So using continue here is inappropriate.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: Ditto, we will enable direct download in the future.

halton: Please make them as TODO instead of code comment out.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: May specify special message for STATUS_PAUSED later. Currently, just use default message.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: minSdkVersion from app side it not mandatory. So I provide 2 interfaces. Alghough check() is public, the class XWalkCoreWrapper could used by classes inside the package only.

halton: This is incorrect assumption, public means it can be used from anyplace, http://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html. We should avoid this api use outside the core_internal package.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: The exception here is ok. Null will returned if the object is not from the reflection layer

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: The exception here is ok. Null will returned if the object is not from the reflection layer

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: The exception here is ok. Null will returned if the object is not from the reflection layer

runtime/android/core/src/org/xwalk/core/XWalkLibraryDelegate.java
lincsoon: Here is just defining the default operation, do nothing and return true. It may return false in XWalkLibraryDelegateEmbedded.

tools/reflection_generator/reflection_generator.py
lincsoon: recovered mod

build/android/generate_xwalk_shared_library.py
lincsoon: I'll add a TODO to mark this issue

build/android/generate_xwalk_shared_library.py
lincsoon: added line break after mkdir

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: added line break after positiveListener

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: There is no explicit modifier before "class XWalkCoreWrapper", so the whole class is package-private.

runtime/android/core/src/org/xwalk/core/XWalkReflectionHelper.java
lincsoon: added copyright

lincsoon:
@halton Thanks for your careful review! I corrected my commit.

build/android/generate_xwalk_core_library_aar.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright (c) 2014 Intel Corporation. All rights reserved.
+# Copyright (c) 2015 Intel Corporation. All rights reserved.
```
halton: nit, this change is not required. We only increase year for new files.

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,247 @@
...
+  ]
+  for f in files_to_copy:
+    source_file = os.path.join(template_dir, f)
+    target_file = os.path.join(out_project_dir, f)
+
+    shutil.copy2(source_file, target_file)
+
+
+def CopyBinaries(out_dir, out_project_dir, src_package):
+  print ('Copying binaries...')
```
halton: nit, remove space after print

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,247 @@
...
+        if ((pathname == 'drawable' and relativedir == 'drawable-mdpi') or
+            relativedir == pathname):
+          source_file = os.path.abspath(os.path.join(dirname, filename))
+          target_file = os.path.join(filepath, filename)
+          shutil.copyfile(source_file, target_file)
+          return
+
+
+def CopyResources(project_source, out_dir, out_project_dir):
+  print ('Copying resources...')
```
halton: ditto

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,247 @@
...
+        ReplaceCrunchedImage(project_source, filename, dirname)
+
+
+def main(argv):
+  option_parser = optparse.OptionParser()
+  AddGeneratorOptions(option_parser)
+  options, _ = option_parser.parse_args(argv)
+
+  if not os.path.exists(options.source):
+    print ('Source project does not exist, please provide correct directory.')
```
halton: nit

build/android/generate_xwalk_shared_library.py
```diff
@@ -0,0 +1,247 @@
...
+
+  readme = os.path.join(src_dir, 'README.md')
+  open(readme, 'w').write(
+      "# Source folder for xwalk_shared_library\n"
+      "## Why it's empty\n"
+      "xwalk_shared_library doesn't contain java sources.\n"
+      "## Why put me here\n"
+      "To make archives keep the folder, "
+      "the src directory is needed to build an apk by ant.")
+  print ('Your Android library project has been created at %s' %
```
halton: nit

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
halton: Yes, I missed `method.invokeWithArguments()` in the end of loop

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,563 @@
...
+                } catch (InterruptedException e) {
+                    break;
+                }
+
+                Cursor cursor = mDownloadManager.query(query);
+                if (cursor == null || !cursor.moveToFirst()) continue;
+
+                int totalIdx = cursor.getColumnIndex(DownloadManager.COLUMN_TOTAL_SIZE_BYTES);
+                int downloadIdx = cursor.getColumnIndex(
+                    DownloadManager.COLUMN_BYTES_DOWNLOADED_SO_FAR);
```
halton: 8-space indent

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,563 @@
...
+            String errMsg = mXWalkActivity.getString(R.string.download_failed_message);
+            if (result == DownloadManager.STATUS_FAILED) {
+                Query query = new Query().setFilterById(mDownloadId);
+                Cursor cursor = mDownloadManager.query(query);
+                if (cursor != null && cursor.moveToFirst()) {
+                    int reasonIdx = cursor.getColumnIndex(DownloadManager.COLUMN_REASON);
+                    int reason = cursor.getInt(reasonIdx);
+                    if (reason == DownloadManager.ERROR_DEVICE_NOT_FOUND) {
+                        errMsg = mXWalkActivity.getString(
+                            R.string.download_failed_device_not_found) ;
```
halton: ditto

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -5,16 +5,563 @@
...
+                    } else if (reason == DownloadManager.ERROR_INSUFFICIENT_SPACE) {
+                        errMsg = mXWalkActivity.getString(
+                            R.string.download_failed_insufficient_space);
```
halton: ditto

runtime/android/core/src/org/xwalk/core/XWalkLibraryDelegate.java
halton: Then why define a boolean return value? just void is enough.

runtime/android/core_internal/src/org/xwalk/core/internal/ReflectExceptionHandler.java
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
halton: 2015

tools/reflection_generator/reflection_generator.py
```diff
@@ -108,37 +107,56 @@ def Touch(path):
...
-                                'will copy it to output folder'))
+                           help=('Output directory where the bridge code is '
+                                 'placed.'))
+  option_parser.add_option('--wrapper_output',
+                           help=('Output directory where the wrap code is '
+                                 'placed.'))
   option_parser.add_option('--stamp', help='the file to touch on success.')
+  option_parser.add_option('--target_arch', help='Target Architecture')
+  option_parser.add_option('--sdk_version', help='API Version')
+  option_parser.add_option('--min_sdk_version', help='Min API Version')
```
halton: It is not common practice to use `_` as option, instead we should use `-`. I know this is not this PR fault, but let's change that.

xwalk_core_library_android.gypi
```diff
@@ -89,11 +89,44 @@
...
+      'actions': [
+        {
+          'action_name': 'pack_xwalk_shared_library',
+          'message': 'Packaging XwalkCore Shared Library Project.',
+          'inputs': [
+            '<(DEPTH)/xwalk/tools/tar.py',
+          ],
+          'outputs': [
+            '<(PRODUCT_DIR)/xwalk_shared_library.tar.gz',
+            '<(PRODUCT_DIR)/pack_xwalk_shared_library_intermediate/always_run',
```
halton: nit, order plz

runtime/android/core/src/org/xwalk/core/XWalkLibraryDelegate.java
lincsoon: The subclasses may failed to initialized. But in superclass, because there is nothing to do, so it will always return true

halton: that make sense, thanks for explanation.

rakuco:
Every time this PR is tested it causes subsequent make_apk_test runs to fail like this:

```
-dex:
      [dex] input: /tmp/Example-lPkEsc/bin/classes
      [dex] input: /tmp/Example-lPkEsc/xwalk_core_library/bin/classes.jar
      [dex] input: /tmp/Example-lPkEsc/libs/xwalk_app_runtime_java.jar
      [dex] input: /tmp/Example-lPkEsc/xwalk_core_library/libs/xwalk_core_library_java.jar
      [dex] Pre-Dexing /tmp/Example-lPkEsc/xwalk_core_library/bin/classes.jar -> classes-b19fc1fd9d57d69ff55ebdd83c34fbaf.jar
      [dex] Pre-Dexing /tmp/Example-lPkEsc/libs/xwalk_app_runtime_java.jar -> xwalk_app_runtime_java-833cba2fd4ef2d239463faa9159b0323.jar
      [dex] Pre-Dexing /tmp/Example-lPkEsc/xwalk_core_library/libs/xwalk_core_library_java.jar -> xwalk_core_library_java-3197c539b7b6158225d25ff6679c37c0.jar
      [dex] Converting compiled files and external libraries into /tmp/Example-lPkEsc/bin/classes.dex...
       [dx] Merged dex A (26 defs/16.9KiB) with dex B (15 defs/19.1KiB). Result is 41 defs/41.3KiB. Took 0.1s
       [dx] 
       [dx] UNEXPECTED TOP-LEVEL EXCEPTION:
       [dx] com.android.dex.DexException: Multiple dex files define Lorg/xwalk/core/R$attr;
       [dx]     at com.android.dx.merge.DexMerger.readSortableTypes(DexMerger.java:596)
       [dx]     at com.android.dx.merge.DexMerger.getSortedTypes(DexMerger.java:554)
       [dx]     at com.android.dx.merge.DexMerger.mergeClassDefs(DexMerger.java:535)
       [dx]     at com.android.dx.merge.DexMerger.mergeDexes(DexMerger.java:171)
       [dx]     at com.android.dx.merge.DexMerger.merge(DexMerger.java:189)
       [dx]     at com.android.dx.command.dexer.Main.mergeLibraryDexBuffers(Main.java:454)
       [dx]     at com.android.dx.command.dexer.Main.runMonoDex(Main.java:302)
       [dx]     at com.android.dx.command.dexer.Main.run(Main.java:245)
       [dx]     at com.android.dx.command.dexer.Main.main(Main.java:214)
       [dx]     at com.android.dx.command.Main.main(Main.java:106)
```

Is there a way to make either the PR or the existing code in Crosswalk properly clean up the files in the build directory to avoid this?

rakuco:
@wang16 ^

lincsoon:
@rakuco It seems make_apk_test fails randomly on try bot. I ran this test on my local machine, it passed on every time.

lincsoon:
@halton code updated according to your review.

rakuco:
The failures are not random. They are caused by building this patch then reverting it and doing another incremental build. I do not know which files need to be removed, hence my call for help. You should be able to reproduce the problem locally very easily by building with your patch, then checking out a pristine master branch, building again (not from scratch) and using `make_apk.py` from `out/Relase/xwalk_app_template/`.

lincsoon:
Thanks! I have reproduced this building error. @halton please don't merge this patch until I fixed this issue.

lincsoon:
@rakuco The root of this problem is out/Release/gen/xwalk_core_java, this directory is built by java.gypi which is the build tool from chromium. By applying my patch, a new resource directory named java_R is generated, this directory does not exist in the output directory built by master branch. But after reverting source code to master branch and rebuilding xwalk_core_java, the java_R directory is not deleted. When parsing the whole output directory and generating resource map, this extra java_R directory is included and finally caused this problem.
Because the cause of the problem is that java.gypi from upstream does not clean up the output directory, and this problem only occurs when reverting to old source code, I think we could ignore this problem on this pull request. From this patch, the extra java_R is ignored when generating resource map, so this problem will not happen again on further pull request.

rakuco:
The problem's that until your patch is merged every single time you submit a new version all other pull requests will fail on Android.
I pinged Shiliu above because since he wrote this part of the build system he might be able to think of something we can do before landing this change that makes the problem go away (like looking for `R.java` in specific locations or something else).

On an unrelated note: see the build log for your latest patch; even though it looks green there are several errors about `ReflectionMethod` not being found.

lincsoon:
Because shiliu has left the job in Intel, we have to solve this problem by ourself. As you say that looking  for R.java in specific locations, I have already do the same thing in this patch. The problem is that the existing build system has limitation, and to solve this problem, we have to merge this patch.
As for the error about ReflectionMethod, this is about generating java docs, not compilation error.

halton:
LGTM, thanks @lincsoon 's effort to make it good shape now.

@rakuco about the trybot failure back and force, it will disappear after this PR get merged. So let's ignore this error and push this PR get merged as easily as possible. Note, the shard mode support is very important for uploading Crosswalk binary to Google Play Store.

halton:
ping @darktears @kenchris @rakuco

lincsoon:
I just modified generate_resource_map.py to make ture old files of resource map are cleaned up before rebuilding.

lincsoon:
@darktears @kenchris @rakuco because other works like security check, publishing on play store, QA test are blocked by this PR, could you help to review and merge this PR as soon as possible.

rakuco:
Sorry, we haven't had time to look at it yet. Some drive-by comments while I don't take a closer look: `print ()` should be `print()`, existing `print` calls should be converted to `print()` separately so that this patch doesn't get even bigger and I was wondering if there's a way to make the existing scripts for xwalk_core_library work for both core and shared instead of duplicating them?

lincsoon:
Because xwalk_shared_library has different resources and library files, and different target in gyp, I think it's better to use another script. But I wrote this script by referencing the script for xwalk_core_library, so they have similar form and code. 

About the print issue, most of them are mixed with real commit, so it's hard to extract them. And for the whole pull request, the code around here are just a small part. 

Because this pull request has been sent a month ago, it has passed @wang16 and @halton 's reivew, and some other important works are blocked by this patch, I think it's better to end this issue earlier.

lincsoon:
@rakuco Is there anything different with the try bot. I committed the same code(3c96934) again, but the building failed.

lincsoon:
rebased to 13.41.317.0

lincsoon:
@rakuco Could you please help to check the try-bot error of Crosswalk Android x86-64. It failed only in the update_scripts step.

rakuco:
Overall, as you probably know I'm not the best person to review the specific Android bits that you are introducing, and I defer that judgment to the other guys who've already taken a look at it. Please just make sure everything still works with M42. Thank you for working on this.

Now I'd like to get back to the comments I made before. Working on those two things will save us from headaches in the future and shouldn't require more than half an hour or one hour at most.

> Because xwalk_shared_library has different resources and library files, and different target in gyp, I think it's better to use another script. But I wrote this script by referencing the script for xwalk_core_library, so they have similar form and code.

Think of the maintenance burden being introduced by duplicating the scripts in `build/android`. Every time some subtle bug that had to be worked around in the xwalk_core scripts now needs to be fixed twice.

I've just run `diff` between the old scripts and the new ones and it's mostly a difference in paths as you said, so generalizing the existing scripts really should not be difficult at all. And then you just pass different arguments in gyp.

Trust me, the code as it is is already a pain to maintain, and duplicating the work will only bite us back in the future.

> About the print issue, most of them are mixed with real commit, so it's hard to extract them.

Not really, it is easy to spot the changes in, for example, `make_apk.py` or `generate_xwalk_core_library.py` that are unrelated to this change (and make the diff larget and more difficult to review and even revert in the future if necessary. These can be changed back in just a few minutes.

baleboy:
@rakuco

lincsoon:
@rakuco Thanks for your advice! I understood your concern and made existing scripts to work for both core and shared library. I also fixed the print issue. As you know, this patch has been under review for a long time. If there are not other problems, it's very helpful that we could see shared mode in canary this week.

rakuco:
\o/
Thanks for the patience. Let's press the green button and hope for the best!

rakuco:
This commit ended up breaking a lot of Android tests, as mentioned in #2915, so I'll have to revert it. Please make sure those tests are passing so we can re-land it.

PR #2781, opened 2015-01-07
[Android] Strip the not implemented error message.

wuhengzhi:
This patch is to strip all of the error messages about not implemented
functions.
Add 'NOTIMPLEMENTED_POLICY=0' to common gypi file.

BUG=XWALK-3188

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -20,6 +20,11 @@
       ['enable_murphy==1', {
         'defines': ['ENABLE_MURPHY=1'],
       }],
+      ['OS=="android"', {
+        'defines': [
+           'NOTIMPLEMENTED_POLICY=0'
+        ],
+      }],
     ],
     'includes': [
       'xwalk_filename_rules.gypi',


rakuco:
Why do we want this? Comment 4 in http://crbug.com/431412 seems to apply here:

> If we are calling a function that is NOTIMPLEMENTED(), even once, let alone many times, then either:
> 1) the caller should not be calling it in this case, and the caller needs fixing, or
> 2) the NOTIMPLEMENTED() should be removed, if it's okay for the function to just do nothing/return an empty/dummy result, or
> 3) the function should actually be implemented (not possible in this case since obviously there is no IP endpoint for the peer of a unix domain socket)

wang16:
I suggest to only do this for release build.

rakuco:
And for all platforms, if that was to happen. My concern is that it will also hide other useful NOTIMPLEMENTED calls. In this specific case, I'd wait a bit more for other comments in the upstream Chromium bug and maybe even provide a patch if there's no activity after a while.

halton:
@wang16 @wuhengzhi if I remember correctly we discussed this issue in another PR. And we should implement this in crosswalk embedder, anything changed?

wang16:
> @wang16 @wuhengzhi if I remember correctly we discussed this issue in another PR. And we should implement this in crosswalk embedder, anything changed?

Upstream has been knowing this issue and ignore it by just disable NOTIMPLEMENTED macro in released builds. We could do the same thing.

wuhengzhi:
The discussion has been updated, ```NOTIMPLEMENTED``` was removed. This patch can be closed.

wang16:
so you will cherry-pick the related PR to chromium-crosswalk right?

wuhengzhi:
Yes, thanks.

halton:
CL https://codereview.chromium.org/841993002 is landed, the change is diffrent with this PR. Can we simply cherry-pick from upstream?

wuhengzhi:
It's OK to cherry-pick from upstream. I will push it to previous PR.

halton:
So this PR can be closed?

wuhengzhi:
Yes, thanks.

PR #2785, opened 2015-01-07
[Tizen] Remove xwalk-launcher and xwalkctl.

rakuco:
Consequently remove the entire `application/tools/linux` directory.
    
Now that we have dropped support for using Crosswalk in shared process
mode on Tizen (commit e4f3799ac), nothing in that directory gets built
or used anywhere anymore.
    
BUG=XWALK-3203


pozdnyakov:
lgtm

PR #2786, opened 2015-01-07
DEPS.xwalk: Roll ozone-wayland (a64feb1 -> 5d7baa9).

rakuco:
* 5d7baa9 Rename WindowTree*Wayland as WindowManagerWayland.
* 9c19d69 Move WindowTreeHostDelegateWindow to platform/

These two commits are backports from ozone-wayland master and removes a
dependency across shared libraries on a class that is not exported and
was causing problems when building Crosswalk with
component=shared_library.


rakuco:
@darktears

darktears:
LGTM

PR #2787, opened 2015-01-08
[Tizen] Fix order of default widget start files

sstolinski:
This patch fixes improper order of default widget entry pages.

### DIFF
--- a/application/common/package/wgt_package.cc
+++ b/application/common/package/wgt_package.cc
@@ -82,8 +82,8 @@ WGTPackage::WGTPackage(const base::FilePath& path)
 const std::vector<std::string>& WGTPackage::GetDefaultWidgetEntryPages() {
   static std::vector<std::string> entry_pages;
   if (entry_pages.empty()) {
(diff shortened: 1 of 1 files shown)


pozdnyakov:
lgtm

sstolinski:
Commit message corrected.

rakuco:
Can you expand the commit message a bit more? This looks like the same fix from aae1f929200f8f166e4f88ae8cd3a84a010c0a1d applied elsewhere in the code; if that's the case, mentioning that the W3C Widget spec specifies a certain order just like the other commit message does is a good idea.

sstolinski:
@rakuco It is almost the same fix, but the code from https://github.com/crosswalk-project/crosswalk/commit/aae1f929200f8f166e4f88ae8cd3a84a010c0a1d is now deleted by https://github.com/crosswalk-project/crosswalk/commit/916d78dabd6ff1befd4c23dc1c3d37acbc95a52c.

PR #2788, opened 2015-01-08
Remove the fullscreen command line option

pozdnyakov:
Fullscreen is still supported in the manifest and by the fullscreen Web API.


kenchris:
lgtm

PR #2790, opened 2015-01-08
Drop Application::LaunchParams struct

pozdnyakov:
Drop Application::LaunchParams struct (as well as all its client code) as it contains only "shared process model"-specific data.

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -90,7 +90,6 @@ Application::Application(
       security_mode_enabled_(false),
       browser_context_(browser_context),
       observer_(NULL),
-      remote_debugging_enabled_(false),
       weak_factory_(this) {
   DCHECK(browser_context_);
   DCHECK(data_.get());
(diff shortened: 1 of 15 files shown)


pozdnyakov:
@kenchris ptal

rakuco:
Is the Linux test failure related?

rakuco:
Can you elaborate on what the semantics for enabling/disabling remote debugging are? The comment in `LaunchParams` saying "used only when running as a service" applied only to `launched_pid` (see commit 2b20692615e4e8885af3d03e0cfb5e39dab4a18f), so it's not clear to me how is supposed to enable or disable remote debugging now.

pozdnyakov:
@rakuco as for remote debugging, the problem was that "--remote-debugging-port"  is passed to BP (which used to be shared between the apps) so we had to introduce "Runtime" flags to define which of the running apps is being debugged at the moment. Since we dropped the Shared process model, the "--remote-debugging-port" cmd line option is enough.

pozdnyakov:
@rakuco ok, here is the missed intermediate step: https://github.com/crosswalk-project/crosswalk/pull/2798

PR #2791, opened 2015-01-08
[Tizen] Split the package into crosswalk-bin and crosswalk-libs.

rakuco:
This is a big pull request, but hopefully both commits here do a good job at
explaining what is going on.

The TL;DR version is that instead of building one single Crosswalk RPM, we
split it into crosswalk-libs (containing all Chromium shared libraries, such as
`libbase.so` and `libcontent.so`) and crosswalk-bin (containing the actual
`xwalk` binary). This is done so that crosswalk-libs does not depend on any
Tizen-specific library and everything in `src/xwalk` only requires
crosswalk-bin to be rebuilt. The consequence is that crosswalk-libs almost
never needs to be built by developers and crosswalk-bin takes only a few
minutes to build.

In order to do this, we had to do some hacks:
* We now build Crosswalk for Tizen with `-Dcomponent=shared_library`. Since the
  libraries are private, they are installed into a private directory so that
  other packages cannot use it.
* To know which Chromium libraries we need to build for crosswalk-libs, a new
  script was added that parses Crosswalk's build dependency graph and outputs
  all the targets which are not in `src/xwalk`.
* Conversely, to avoid rebuilding the Chromium libraries in crosswalk-bin, we
  apply one big patch that converts several targets in Chromium from `'type':
  '<(component)'` to `'type': 'none'` so that they are cause any files to be
  built or any dependencies to be processed. Maintaining this patch is costly,
  but it did not require many changes when transitioning from M39 to M40.

Building Crosswalk should now require one to specify the spec file to gbs, like
this: `gbs build --spec crosswalk-bin`.

Please check the commit messages of the two commits in this pull request for
more information on what has been done.

BUG=XWALK-2571


rakuco:
@darktears @pozdnyakov @baleboy @halton @baillaw @kooltux
An earlier version of this patch was already tested by the Eurogiciel people, and I've just tested this version on a Tizen Common daily release.

baleboy:
I lack the knowledge to review the actual change, but in general this is fine by me.

kenchris:
lgtm

darktears:
lgtm

rakuco:
The version I've just pushed is the same one with the conflicts resolved. I'm going to merge it.

PR #2794, opened 2015-01-12
[Tizen] Fix "Uncaught ReferenceError: tizen is not defined" error when starting an application

takethathe:
The issue is caused by injecting java script to enable firing widget preferences
storage event, and this will skip the initialization progress of extension
process. The patch will make sure inject this script just after the extension
has been initialized.

BUG=https://bugs.tizen.org/jira/browse/TC-2306


takethathe:
@pozdnyakov @tmpsantos 

rakuco:
Looks sane to me.

pozdnyakov:
looks like it should be a Tizen-specific code

pozdnyakov:
lgtm

PR #2796, opened 2015-01-12
[Tizen] Fixed the parsing of widget manifest

rynkowsg:
Problem:    Whitespaces in some nodes and attributes values should be removed.
Solution:   Collapse and trim whitespaces at selected strings.

Links to specification:
    http://www.w3.org/TR/widgets/#rule-for-getting-text-content-with-normalized-white-space-0
    http://www.w3.org/TR/widgets/#rule-for-getting-a-single-attribute-value-0

BUG=XWALK-2775

### DIFF
--- a/application/common/application_file_util.cc
+++ b/application/common/application_file_util.cc
@@ -19,6 +19,7 @@
 #include "base/path_service.h"
 #include "base/strings/string16.h"
 #include "base/strings/stringprintf.h"
+#include "base/strings/string_util.h"
 #include "base/strings/utf_string_conversions.h"
 #include "base/threading/thread_restrictions.h"
 #include "net/base/escape.h"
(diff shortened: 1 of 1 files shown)


takethathe:
@rynkowsg 
Does this patch support the patten like this:
```
<name> Sample <span>Name</span> Test</name>
```

rakuco:
Which reminds me that we should probably have some tests for this change.

rynkowsg:
@takethathe
Yes, the result is as it should be.

takethathe:
LGTM

rynkowsg:
Tests for this change should be [here](https://github.com/crosswalk-project/crosswalk-test-suite/tree/master/webapi/tct-wgtapi01-w3c-tests/widgetinterface/w3c]). I picked some tests but they are broken, in my opinion.

rynkowsg:
@pozdnyakov 
Have you any doubt about that fix? If not, I will be glad of approving it.

pozdnyakov:
lgtm

PR #2798, opened 2015-01-13
[Tizen/Linux] Fix crash at XWalkDevToolsDelegate::ProcessAndSaveThumbnail

pozdnyakov:
The crash happened if the target page did not have thumbnail icon.


pozdnyakov:
This crash did not show up in the browser tests because the "Runtime::remote_debugging_enabled" flag wasn't set there (erroneously).

kenchris:
aha, ok lgtm

rakuco:
Does it need to be backported to crosswalk-11?

pozdnyakov:
@rakuco apparently yes

PR #2801, opened 2015-01-13
[Tizen] Enable SetVisibilityChangeAnimations by default

joone:
The IVI homescreen UI has CSS animations, but it still runs and
consume more than 35% CPU power when the UI has been hidden.

This patch allows to pause/resume CSS animations when visibilityChange
events are fired.

BUG=TC-1899

### DIFF
--- a/runtime/browser/ui/native_app_window_tizen.cc
+++ b/runtime/browser/ui/native_app_window_tizen.cc
@@ -121,6 +121,9 @@ void NativeAppWindowTizen::Initialize() {
     OnScreenOrientationChanged(
         SensorProvider::GetInstance()->GetScreenOrientation());
   }
+
+  // Allow CSS animations to stop/play according to VisibilityChange event.
+  GetWidget()->SetVisibilityChangedAnimationsEnabled(true);
 }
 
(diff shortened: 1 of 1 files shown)


pozdnyakov:
lgtm

rakuco:
+@ds-hwang
Is this something we want on all platforms?

ds-hwang:
I'm not familiar to current aura window code, but I'm not sure which behavior this PR changes.
As briefly reading code,  SetVisibilityChangeAnimations is implemented in only Windows.
Do you see CPU usage difference after this patch?
```
void DesktopWindowTreeHostX11::SetVisibilityChangedAnimationsEnabled(
    bool value) {
  // Much like the previous NativeWidgetGtk, we don't have anything to do here.
}

void DesktopWindowTreeHostWayland::SetVisibilityChangedAnimationsEnabled(
    bool value) {
  // Much like the previous NativeWidgetGtk, we don't have anything to do here.
}
```

joone:
@ds-hwang  yes, the animations stops when a window has been hidden and I see CPU usage drop.

ds-hwang:
ok, the behavior sounds good.
rubber stamps because I don't fully understand this area.

PR #2803, opened 2015-01-14
[Tizen] Merge the GPU process to the Browser process

pozdnyakov:
Step of the Dual Process model implementation.
See https://lists.crosswalk-project.org/pipermail/crosswalk-dev/2014-December/002308.html for the details.

BUG=XWALK-3203

### DIFF
--- a/runtime/browser/xwalk_browser_main_parts_tizen.cc
+++ b/runtime/browser/xwalk_browser_main_parts_tizen.cc
@@ -34,6 +34,11 @@ void XWalkBrowserMainPartsTizen::PreMainMessageLoopStart() {
   CommandLine* command_line = CommandLine::ForCurrentProcess();
   command_line->AppendSwitch(switches::kIgnoreGpuBlacklist);
 
+  // switches::kInProcessGPU is not being used here because it
+  // doesn't have the CONTENT_EXPORT macro despite the fact it is exposed by
+  // content_switches.h.
+  command_line->AppendSwitch("in-process-gpu");
+
   // Enable Accelerated 2D Canvas.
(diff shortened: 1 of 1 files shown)


pozdnyakov:
To be merged after XWALK-3352 is fixed.

rakuco:
The Tizen Common build failure looks real. Also, can you write a bit more on the commit message and explain why this is helpful and necessary? Otherwise anyone looking at the history will first need to find out what "dual process model implementation" means and how this is related to it.

runtime/browser/xwalk_browser_main_parts_tizen.cc
```diff
@@ -34,6 +34,11 @@ void XWalkBrowserMainPartsTizen::PreMainMessageLoopStart() {
   CommandLine* command_line = CommandLine::ForCurrentProcess();
   command_line->AppendSwitch(switches::kIgnoreGpuBlacklist);
 
+  // switches::kInProcessGPU is not being used here because it
+  // doesn't have the CONTENT_EXPORT macro despite the fact it is exposed by
+  // content_switches.h.
+  command_line->AppendSwitch("in-process-gpu");
```
hmin: Bad style here. Please use ```switches::kInProcessGpu``` instead.

halton: Nice catch and just say hello to @hmin :)

rakuco: There is an explanation in the comment about why it is not being used.

rakuco:
Rendered obsolete by #3299.

PR #2804, opened 2015-01-14
[Tizen] Remove Tizen locale listener

pozdnyakov:
The Tizen locale listener is not used anywhere, so should be removed (has not been used since commit edf502d, which removed its only user, XWalkRunnerTizen::GetLocale()).


rakuco:
It'd be good to mention it has not been used since commit edf502d06921d85b25fcf5b680ba4f631a7dea88, which removed its only user, `XWalkRunnerTizen::GetLocale()`.

PR #2805, opened 2015-01-15
[Android] Fix app crash when two pages load each other for more than 512...

fujunwei:
... times

The crash is caused by local reference table overflow, there is a thread hold the
input stream reference by analysing MAT.
Threads attached through JNI must call DetachCurrentThread before they exit. So
call DetachFromVM after AttachCurrentThread which wraps the DetachCurrentThread
function.

BUG=XWALK-3351

(cherry picked from commit 8fbf5a4cfe89c52e109a3e1c119a3de69a78933c)


fujunwei:
@wang16 PTAL,Thanks.

wang16:
LGTM, thanks

PR #2809, opened 2015-01-15
[Tizen] Skip passing chromium flags to appcore_init

tiwanek:
Pass only positional arguments to appcore (aul parameters).
Skip chromium flags.
Library libbundle expects even number of positional arguments is form:
 __AUL_KEY__ __BASE64_VALUE__ [and so on...]
It crashes for odd number of parameters.
Passing chromium flags may results in parsing it wrong by appcore.

BUG=XWALK-3350


pozdnyakov:
lgtm

PR #2812, opened 2015-01-16
[Crosswalk-Mini][Android] add compressed library dependency

d0ngjun:
@halton please help review, thanks


halton:
code lgtm, comment on git log.
1. add -> Add
2. Explain why this change is need, that is what issue you're fixing.

halton:
LGTM, will merging. Let's monitor the buildbot result.

PR #2813, opened 2015-01-19
[Tizen] packaging: Stop passing shared_process_mode=1.

rakuco:
This is a leftover caused by the Tizen package split happening somewhat
concurrently with the removal of shared process model support on Tizen:
`crosswalk-libs.spec` was still specifying a flag that has no effect.


pozdnyakov:
LGTM

PR #2814, opened 2015-01-20
[Tizen] Remove xwalk backend from libs packaging

jizydorczyk:
When crosswalk-libs and crosswalk-bin is build for the first time installation of rpms works fine. If next build is incremental libxwalk_backend_lib.so is not removed before next packaging, what causes packaging it into both crosswalk-libs and crosswalk-bin packages. This patch removes libxwalk_backend_lib.so before packaging of crosswalk-libs, when package is build with defined "BUILDDIR_NAME".

BUG=XWALK-3416

### DIFF
--- a/packaging/crosswalk-libs.spec
+++ b/packaging/crosswalk-libs.spec
@@ -225,6 +225,13 @@ DEPENDENCIES=`python %{SOURCE1002} dump.json ${NINJA_TARGETS}`
 
 ninja %{?_smp_mflags} -C src/out/Release ${DEPENDENCIES}
 
+# In '%files' section below there are added all built libraries, which satisfy
+# pattern: 'xwalk/lib/lib*.so'. During build crosswalk-bin.spec generates new
+# library - libxwalk_backend_lib.so, which should not be packaged into
+# crosswalk-libs, but then after incremental build it will also satisfy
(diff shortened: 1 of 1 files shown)


jizydorczyk:
@rakuco could you take a look at this?

packaging/crosswalk-libs.spec
```diff
@@ -225,6 +225,10 @@ DEPENDENCIES=`python %{SOURCE1002} dump.json ${NINJA_TARGETS}`
 
 ninja %{?_smp_mflags} -C src/out/Release ${DEPENDENCIES}
 
+%if %{?BUILDDIR_NAME:1}%{!?BUILDDIR_NAME:0}
```
rakuco: Why do you think the check is needed here? Even if one's building in the default directory, removing this file should be harmless.

rakuco: Also, if you're doing this it makes sense to add a comment there explaining why you're removing a seemingly random file from the build directory.

jizydorczyk:
@rakuco I made changes according to your comments, sorry for late response.

rakuco:
lgtm, thanks. In the future we need a better approach, but that's good enough for now.

PR #2817, opened 2015-01-22
[Tizen] Fix read_only can't be NULL

qjia7:
In config.xml, if widget.preference absents read_only attribute, the corresponding item
will fail to be inserted into the db. This change will fix it.

### DIFF
--- a/application/extension/application_widget_storage.cc
+++ b/application/extension/application_widget_storage.cc
@@ -100,11 +100,13 @@ bool AppWidgetStorage::SaveConfigInfoItem(base::DictionaryValue* dict) {
   DCHECK(dict);
   std::string key;
   std::string value;
-  bool read_only = false;
   if (dict->GetString(kPreferencesName, &key) &&
-      dict->GetString(kPreferencesValue, &value) &&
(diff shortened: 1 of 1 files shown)


application/extension/application_widget_storage.cc
```diff
@@ -102,9 +102,11 @@ bool AppWidgetStorage::SaveConfigInfoItem(base::DictionaryValue* dict) {
   std::string value;
   bool read_only = false;
   if (dict->GetString(kPreferencesName, &key) &&
-      dict->GetString(kPreferencesValue, &value) &&
-      dict->GetBoolean(kPreferencesReadonly, &read_only))
+      dict->GetString(kPreferencesValue, &value)) {
+    // read_only column can be NULL.
+    dict->GetBoolean(kPreferencesReadonly, &read_only);
```
halton: nit, you can move declaration of `read_only` into if scope.

qjia7: Done

halton:
LGTM

pozdnyakov:
lgtm

PR #2818, opened 2015-01-22
[Tizen] Transform app id to lowercase in active StoragePartitionPath

qjia7:
In XWalkContentBrowserClient::GetStoragePartitionConfigForSite, we use partition_domain
equals site.host() which is lowercase. However, in CollectUnusedStoragePartitions, the
installed app id could be case-sensitive. So, it will result in deleting installed app
storage since active_paths is case-sensitive and trash path is lowercase for app id domain.

### DIFF
--- a/application/browser/application_service_tizen.cc
+++ b/application/browser/application_service_tizen.cc
@@ -4,6 +4,7 @@
 
 #include "xwalk/application/browser/application_service_tizen.h"
 
+#include <algorithm>
 #include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


halton:
nit in git log, [Tizen] transform - > [Tizen] Transform

halton:
LGTM

application/browser/application_service_tizen.cc
```diff
@@ -36,8 +37,10 @@ void CollectUnusedStoragePartitions(XWalkBrowserContext* context,
       new base::hash_set<base::FilePath>()); // NOLINT
 
   for (unsigned i = 0; i < app_ids.size(); ++i) {
+    std::string app_id = app_ids.at(i);
+    std::transform(app_id.begin(), app_id.end(), app_id.begin(), ::tolower);
```
pozdnyakov: I'd prefer seeing it inside GetStoragePartitionPath

qjia7:
Ping @pozdnyakov

PR #2829, opened 2015-01-26
[Tizen] Add url decoding for 'uri' attribute of signatures

sung-su:
The value of URI attribute of signature files can be url encoded string.
But, SignatureValidator can't recongize url encoded 'uri' of signatures.
Add url decoding routine to SignatureParser.

BUG=XWALK-3479

### DIFF
--- a/application/common/tizen/signature_parser.cc
+++ b/application/common/tizen/signature_parser.cc
@@ -12,11 +12,14 @@
 #include <vector>
 #include "base/files/file_util.h"
 #include "base/logging.h"
+#include "base/strings/string_util.h"
+#include "base/strings/utf_string_conversions.h"
 #include "libxml/parser.h"
 #include "libxml/xmlschemas.h"
 #include "libxml/xpathInternals.h"
 #include "libxml/xmlreader.h"
 #include "third_party/libxml/chromium/libxml_utils.h"
+#include "url/url_util.h"
 
 namespace {
 const char kExpectedXmlns[] = "http://www.w3.org/2000/09/xmldsig#";
@@ -111,6 +114,7 @@ xmlNsPtr GetNamespace(const xmlNodePtr node, const char* expected_href) {
 
 namespace xwalk {
 namespace application {
+
 bool ParseSignedInfoElement(
     const xmlNodePtr node, const xmlNsPtr signature_ns, SignatureData* data) {
   xmlNodePtr signed_info_node =
@@ -159,6 +163,18 @@ bool ParseSignedInfoElement(
     if (uri.empty()) {
       LOG(ERROR) << "Missing URI attribute.";
       return false;
+    }
+    url::RawCanonOutputW<1024> decoded_uri;
+    url::DecodeURLEscapeSequences(uri.c_str(),
+                                  uri.length(),
+                                  &decoded_uri);
+    base::string16 uri16(decoded_uri.data(),
+                         decoded_uri.length());
+    std::string uri8 = base::UTF16ToUTF8(uri16);
+    if (!uri8.empty()) {
+      uri = uri8;
+    } else {
+      LOG(WARNING) << "Failing to decode URI.";
     }
     reference_set.insert(uri);
   }


application/common/tizen/signature_parser.cc
```diff
@@ -111,6 +111,34 @@ xmlNsPtr GetNamespace(const xmlNodePtr node, const char* expected_href) {
 
 namespace xwalk {
 namespace application {
```
halton: nit, empty line after

application/common/tizen/signature_parser.cc
```diff
@@ -111,6 +111,34 @@ xmlNsPtr GetNamespace(const xmlNodePtr node, const char* expected_href) {
...
+int hexToInt(char a) {
```
halton: Please follow chromium naming rule for function, `hexToInt` -> `HexToInt`

application/common/tizen/signature_parser.cc
```diff
@@ -111,6 +111,34 @@ xmlNsPtr GetNamespace(const xmlNodePtr node, const char* expected_href) {
...
+  if (a >= '0' && a <= '9') return a - '0';
+  if (a >= 'A' && a <= 'F') return a - 'A' + 10;
+  if (a >= 'a' && a <= 'f') return a - 'a' + 10;
+  return -1;
```
halton: I'd suggest to copy the implemation of [src/third_party/WebKit/Source/core/inspector/JSONParser.cpp] (https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/core/inspector/JSONParser.cpp&q=hextoInt&sq=package:chromium&l=258&ct=rc&cd=6&dr=CSs). 

Note, please change to chromium coding style, not the WebKit/Blink style.

application/common/tizen/signature_parser.cc
```diff
@@ -111,6 +111,34 @@ xmlNsPtr GetNamespace(const xmlNodePtr node, const char* expected_href) {
...
+}
+
+std::string decodeProcent(const std::string& path) {
```
halton: A example to explain how the decoding look like should be helpful

application/common/tizen/signature_parser.cc
```diff
@@ -151,6 +179,7 @@ bool ParseSignedInfoElement(
   }
 
   std::string uri;
+  std::string decoded_uri;
```
halton: declare this var as close as first time used.

application/common/tizen/signature_parser.cc
halton: Look the code again, please  put `HexToInt()` and `DecodeProcent()` under anonymous namespace.

rakuco:
this kind of parsing is exactly the kind of code I'd expect to find in Chromium. did you look for it before rolling your own version?

halton:
> this kind of parsing is exactly the kind of code I'd expect to find in Chromium. did you look for it before rolling your own version?

I did rough searching chromium code, all the HexToInt is private. Anyway @sung-su can do more searching.

sung-su:
I've fixed except "DecodeProcent" function.
I found the decode function in the Chromium base source code which has similar functionality. However, it is not suitable because complex.

application/common/tizen/signature_parser.cc
```diff
@@ -160,6 +185,12 @@ bool ParseSignedInfoElement(
       LOG(ERROR) << "Missing URI attribute.";
       return false;
     }
+    std::string decoded_uri = DecodeProcent(uri);
+    if (!decoded_uri.empty()) {
+      uri = decoded_uri;
+    } else {
+      LOG(WARNING) << "Failing to decode URI.";
```
rakuco: If decoding fails, shouldn't you abort whatever you're doing and fail?

rakuco:
> However, it is not suitable because complex.

That doesn't sound like a convincing argument. Writing your own URI decoding function should be your very last resort. Doesn't `url::DecodeURLEscapeSequences` solve your problem?
Also, have you checked if other the parsing of other URI fields also have this problem?

sung-su:
I've fixed.
@rakuco 
If decoding fails, abort at the validation check step.

pozdnyakov:
lgtm

PR #2830, opened 2015-01-26
[Tizen] Use appcore RESET event as application starting point

tiwanek:
Starting application should take places in RESET event.
This change is required for futher modifications
for appcontrol implementation.

BUG=XWALK-2779

### DIFF
--- a/application/browser/application_system_tizen.cc
+++ b/application/browser/application_system_tizen.cc
@@ -15,7 +15,6 @@
 #include "content/public/browser/render_process_host.h"
 #include "content/public/common/content_switches.h"
 #include "net/base/filename_util.h"
-#include "xwalk/application/browser/application_tizen.h"
 #include "xwalk/application/browser/application_service_tizen.h"
 #include "xwalk/application/common/application_manifest_constants.h"
 #include "xwalk/application/common/id_util.h"
@@ -49,7 +48,7 @@ namespace xwalk {
 namespace application {
 
 ApplicationSystemTizen::ApplicationSystemTizen(XWalkBrowserContext* context)
-  : ApplicationSystem(context) {
+    : ApplicationSystem(context) {
 }
 
 ApplicationSystemTizen::~ApplicationSystemTizen() {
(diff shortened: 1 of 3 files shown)


rakuco:
Ping @pozdnyakov 

> XWALK-2779

This should either be "BUG=XWALK-2779" or, if the issue is not solved completely, something like "Related to XWALK-2779".

tiwanek:
I will change it to "Related to XWALK-2779".
And looking on code, there are few thing to fix. Please, wait for new version of this.

tiwanek:
@pozdnyakov

application/browser/application_system_tizen.cc
```diff
@@ -59,23 +60,40 @@ namespace {
...
+    case AE_RESET: {
+      base::FilePath exec_path(app_system->launch_params()->argv()[0]);
+      std::string app_id = exec_path.BaseName().MaybeAsASCII();
+      LOG(INFO) << "Attempting to launch the app with id: " << app_id;
+      if (!IsValidApplicationID(app_id))
+        break;
+
+      ApplicationServiceTizen* app_service_tizen =
+          ToApplicationServiceTizen(app_system->application_service());
+      Application* app = app_service_tizen->LaunchFromAppID(app_id);
```
pozdnyakov: can we just re-assign "data" here? (data = app)

pozdnyakov: so that we could avoid adding set_current_app/current_app

application/browser/application_system_tizen.h
```diff
@@ -17,7 +20,18 @@ class ApplicationSystemTizen : public ApplicationSystem {
...
+    current_app_ = current_app;
+  }
+  ApplicationTizen* current_app() { return current_app_; }
+  base::CommandLine* launch_params() { return launch_params_; }
+
  private:
+  // Shared process model was dropped
+  // so I assume this is acceptable
+  ApplicationTizen* current_app_;
+  base::CommandLine* launch_params_;
```
pozdnyakov: I wouldn't add both of these

application/browser/application_system_tizen.cc
```diff
@@ -90,24 +108,14 @@ bool ApplicationSystemTizen::LaunchFromCommandLine(
...
-
-  Application* app = app_service_tizen->LaunchFromAppID(app_id);
-
-  if (!app)
-    return false;
 
   const std::string& name = std::string("xwalk-") + app_id;
   appcore_ops.cb_app = application_event_cb;
-  appcore_ops.data = app;
+  appcore_ops.data = this;
```
pozdnyakov: struct AppCallbackData { 
ApplicationSystem* app_system;
std::string app_id;
} 
 appcore_ops.data = new AppCallbackData(..);

application/browser/application_system_tizen.cc
tiwanek: What do you exactly mean when saying "re-assign"? I need ApplicationSystem in this handler, not Application. It is not possible to reassign date for handler.

application/browser/application_system_tizen.cc
tiwanek: Ok, I can do this that way. I thought about it but idea of passing ownership of some ptr to handler didn't appeal to me. On the second though, this can be deallocated on terminate event.

application/browser/application_system_tizen.cc
```diff
@@ -57,25 +56,58 @@ ApplicationSystemTizen::~ApplicationSystemTizen() {
...
     case AE_UNKNOWN:
     case AE_CREATE:
       break;
     case AE_TERMINATE:
-      app->Terminate();
+      if (handler_data->current_app) {
+        handler_data->current_app->Terminate();
+        delete handler_data;
       break;
+    }
```
pozdnyakov: '}' should be before 'break'

application/browser/application_system_tizen.cc
```diff
@@ -57,25 +56,58 @@ ApplicationSystemTizen::~ApplicationSystemTizen() {
 
 namespace {
 
+struct AppcoreHandlerData {
+  ApplicationSystemTizen* app_system;
+  base::CommandLine* launch_params;
```
pozdnyakov: as I see only app_id is needed, why this should contain 'base::CommandLine*' ?

application/browser/application_system_tizen.cc
```diff
@@ -57,25 +56,57 @@ ApplicationSystemTizen::~ApplicationSystemTizen() {
 
 namespace {
 
+struct AppcoreHandlerData {
+  ApplicationSystemTizen* app_system;
+  std::string app_id;
+  ApplicationTizen* current_app;
+
+  AppcoreHandlerData(ApplicationSystemTizen* system,
+      std::string id)
```
pozdnyakov: const std::string&

application/browser/application_system_tizen.cc
```diff
@@ -93,21 +124,12 @@ bool ApplicationSystemTizen::LaunchFromCommandLine(
...
-
-  if (!app)
-    return false;
 
   const std::string& name = std::string("xwalk-") + app_id;
   appcore_ops.cb_app = application_event_cb;
-  appcore_ops.data = app;
+  appcore_ops.data =
+      new AppcoreHandlerData(this,
+          base::FilePath(cmd_line.argv()[0]).BaseName().MaybeAsASCII());
```
pozdnyakov: we have already "std::string app_id = exec_path.BaseName().MaybeAsASCII();" , so why not
new AppcoreHandlerData(this, app_id ) ?

pozdnyakov:
lgtm

PR #2831, opened 2015-01-26
[Android] Enable WebCL feature for android.

junmin-zhu:


junmin-zhu:
@rakuco @wang16 , would you help to review the patch? thanks.

wang16:
lgtm, thanks.

rakuco:
Can't it be set in gyp instead?

junmin-zhu:
@rakuco , do you mean add it in "build/common.gypi"?

rakuco:
Wherever it makes sense the most, I was referring to "a gyp file" in general instead of piling up even more flags in gyp_xwalk.

junmin-zhu:
@rakuco , update the PR, Add the "enable_webcl" flags in "build/common.gypi"

build/common.gypi
```diff
@@ -20,6 +21,9 @@
       ['enable_murphy==1', {
         'defines': ['ENABLE_MURPHY=1'],
       }],
+      ['OS=="android" and enable_webcl==1', {
```
rakuco: I think it makes more sense to separate these: here, you only check if `enable_webcl==1`, and before this you set `enable_webcl%` to 0 or 1 depending on whether you're on Android or not.

build/common.gypi
```diff
@@ -3,6 +3,7 @@
     'tizen%': 0,
     'tizen_mobile%': 0,
     'enable_murphy%': 0,
+    'enable_webcl%': 1,
```
rakuco: It's sad that the other variables here have no explanation; can you add some comments about this one, as is the case for the variables in Chromium's `build/common.gypi`?

rakuco:
I've left some minor inline comments. In addition to those: you've double- and triple-checked that this works, right?

hmin:
lgtm. 

A cool and advanced feature for Crosswalk on Android!

hmin:
@rakuco 
> I've left some minor inline comments. In addition to those: you've double- and triple-checked that this works, right?

Good comments but this is a disappointed question, doesn't it? I believe the submitter spends lots of effort on making it works, and don't need such an unnecessary double-confirmation. Anyway, it is more or less an experimental feature for Crosswalk, if you find some issues in future, you can submit a patch to fix it.

rakuco:
No, it is not. I'm asking precisely because this change is different from the proposed one, and before that there were problems with the patch in blink-crosswalk caused by misunderstandings about the build system. This is not a condescending question, I am honestly asking if the author verified that the change worked and the macro was defined only under the right circumstances and propagated to the desired files.

hmin:
@rakuco Didn't you see all builtbots are green?

junmin-zhu:
@rakuco , I promise that every updated PR works. Thanks for your reminder. 
Update the PR, could you please have a look again?

Thanks @hmin for your lgtm on so late night.:)

rakuco:
> @rakuco Didn't you see all builtbots are green?

I'm not sure where you're trying to get with the comments here, but they are certainly not helping the review.

Junmin can speak for himself if he feels any of my comments have rubbed him the wrong way (even if it's never been my intention), and other than that we're moving along with the pull request as usual.

If you have anything to comment or suggest about this patch, please go ahead, otherwise I really don't see how your comments are doing any good.

build/common.gypi
```diff
@@ -9,6 +9,16 @@
       'tizen%': '<(tizen)',
       'tizen_mobile%': '<(tizen_mobile)',
       'enable_murphy%': '<(enable_murphy)',
+      # Whether to enable WebCL.
+      'enable_webcl%': '<(enable_webcl)',
+      'variables': {
```
rakuco: The non-Android bots failed because you only set `enable_webcl` to any value on Android. I suggest a structure more similar to Chromium's `common.gypi`:

```python
{
  'variables': {
    # ...
    'conditions': [
      ['OS=="android"', {
        'enable_webcl%': 1,
      }, {
        'enable_webcl%': 0,
      }],
    ],
  },
  'target_defaults': {
    'variables': {
      # ...
      'enable_webcl': '<(enable_webcl)',
    },
    # ...
  },
}
```

rakuco:
lgtm, thanks.

junmin-zhu:
I think we all want to land the feature with good quality. Thanks all on this.
Let's focus on the code. 

Personally, @wang16 and I think the first one is better and simpler.
Anyway, we respect your comments.

wang16:
@junmin-zhu , LGTM, feel free to merge after the bots' result.

baleboy:
Congrats! :)

junmin-zhu:
Nice to see this. Thanks all.

sprd-ffos:
it's finally enabled. congs.

yongsheng:
finally it's enabled. congs.

Issue #229, opened 2013-07-24
[Android] Enable in-process renderer mode for Android port


yongsheng:
On Android, we should use in-process renderer mode, e.g., renderer is a thread not a process, just like AwContents does. We don't need multiprocess for each web application. And more, it's not feasible to start renderer services from web apk to 'library' apk.

PR #2832, opened 2015-01-27
DEPS.xwalk: Roll chromium-crosswalk (bfb8dc3 -> 7260dea).

huningxin:
7260dea Merge pull request #229 from huningxin/fix_xwalk_3344
61ae689 Allow universal access from file if flag is set and url is file scheme.


huningxin:
@rakuco , please take a look. Thanks!

Issue #230, opened 2013-07-24
[Android] Add runtime Java API


yongsheng:
The runtime library will provide a small set of APIs. also add a test shell to test these APIs.

PR #2834, opened 2015-01-27
DEPS.xwalk: Roll chromium-crosswalk (9b9f7b6 -> 178d7d3)

huningxin:
178d7d3 Merge pull request #230 from huningxin/crosswalk-10
92bd952 Allow universal access from file if flag is set and url is file scheme.


huningxin:
To fix https://crosswalk-project.org/jira/browse/XWALK-3344. @rakuco , please take a look. Thanks.

PR #2836, opened 2015-01-27
Fix leaking file descriptor when a new app is launched

tmpsantos:
When establishing a communication channel between the 2 process
we create 2 sockets endpoints and one is sent to the process we
are communicating to. After sending, we need to close this fd.

BUG=TC-2361

### DIFF
--- a/application/browser/linux/running_application_object.cc
+++ b/application/browser/linux/running_application_object.cc
@@ -119,6 +119,9 @@ RunningApplicationObject::~RunningApplicationObject() {
 
 void RunningApplicationObject::TerminateApplication() {
   application_->Terminate();
+
+  if (ep_bp_channel_.socket.fd != -1)
+    close(ep_bp_channel_.socket.fd);
 }
 
 void RunningApplicationObject::OnExported(const std::string& interface_name,


application/browser/linux/running_application_object.cc
```diff
@@ -319,7 +319,8 @@ void RunningApplicationObject::SendChannel(
   dbus::MessageWriter writer(response.get());
   writer.AppendString(ep_bp_channel_.name);
 
-  scoped_ptr<dbus::FileDescriptor> client_fd(new dbus::FileDescriptor(fd));
+  scoped_ptr<dbus::FileDescriptor> client_fd(new dbus::FileDescriptor());
+  client_fd->PutValue(fd);
```
pozdnyakov: hm.. how does it help?

tmpsantos: Passing the fd in the constructor doesn't transfer ownership but PutValue() will do it.
When dbus::FileDescriptor goes out of scope it will close the fd.

pozdnyakov:
lgtm

rakuco:
Shouldn't this target master instead of crosswalk-11?

tmpsantos:
> Shouldn't this target master instead of crosswalk-11?

This code was removed from the master branch together with the shared process model. The leak won't happen there.

rakuco:
Ok

PR #2846, opened 2015-02-05
[Tizen] Rebase Murphy support to M41.

nagineni:
Rebase Murphy support to M41 and fix build failures.

BUG=XWALK-3418

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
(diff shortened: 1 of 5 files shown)


nagineni:
@rakuco

rakuco:
Can you resubmit so we can check if the IVI slave is back to normal? Also, does this patch depend on rolling DEPS.xwalk first or is it independent from the patch in chromium-crosswalk?

nagineni:
Rolled DEPS.xwalk here since the patch in chromium-crosswalk and this one is dependent on each other.

rakuco:
The build was broken and I've fixed it in #2848. Can you please resubmit one last time so I can finally merge the PR?

PR #2848, opened 2015-02-05
Fix the build after 3b52dd1.

rakuco:
Adapt to CL https://codereview.chromium.org/707173004: the code in
3b52dd1 ("Fix the issue about placeholder was not cleared in textarea")
was correct for M40, but we moved to M41 while it was under review and
nobody checked if it still built before merging the pull request.


rakuco:
@pozdnyakov @wuhengzhi @darktears

pozdnyakov:
lgtm

PR #2850, opened 2015-02-05
Adding initialization for remote_debugging_enabled_ (from issue 274840)

terriko:


runtime/browser/runtime.cc
```diff
@@ -60,6 +60,7 @@ Runtime::Runtime(content::WebContents* web_contents)
     : WebContentsObserver(web_contents),
       web_contents_(web_contents),
       fullscreen_options_(NO_FULLSCREEN),
+      remote_debugging_enabled_(false),
```
pozdnyakov: we never use this data member, it should be just removed from the class declaration

terriko: That's even better, if it's possible.

terriko:
Modified code to remove the variable instead as per @pozdnyakov 's comment

rakuco:
Please make sure this is cherry-picked (with `git cherry-pick -x`) to crosswalk-12 and crosswalk-11.

baleboy:
@rakuco crosswalk-12 definitely, but in crosswalk-11 for some reason the variable was initialized (though I'm not sure if it's used)

rakuco:
Oh, right. I forgot the shared process model changes did not go into crosswalk-11.

rakuco:
Done in #2864.

PR #2855, opened 2015-02-10
[Tizen] Fix the crash of RemoteDebuggingServer when repeatedly resetting

wuyuehang:
Guarantee the right time sequence of resetting TCPServerSocket inside
RemoteDebuggingServer in case of repeatedly invoke it with same tcp port
parameter which leads to the crash.

@pozdnyakov @halton 
Please spare time to help review, thank you very much.

### DIFF
--- a/runtime/browser/xwalk_runner.cc
+++ b/runtime/browser/xwalk_runner.cc
@@ -176,6 +176,9 @@ void XWalkRunner::OnRenderProcessHostGone(content::RenderProcessHost* host) {
 void XWalkRunner::EnableRemoteDebugging(int port) {
   const char* local_ip = "0.0.0.0";
   if (port > 0 && port < 65535) {
+    if (remote_debugging_server_.get() &&
+        remote_debugging_server_.get()->port() == port)
+      remote_debugging_server_.reset();
     remote_debugging_server_.reset(
         new RemoteDebuggingServer(browser_context(),
             local_ip, port, std::string()));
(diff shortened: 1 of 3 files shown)


runtime/browser/xwalk_runner.cc
```diff
@@ -176,6 +176,9 @@ void XWalkRunner::OnRenderProcessHostGone(content::RenderProcessHost* host) {
 void XWalkRunner::EnableRemoteDebugging(int port) {
   const char* local_ip = "0.0.0.0";
   if (port > 0 && port < 65535) {
+    // Ensure RemoteDebuggingServer release TCPServerSocket before applying for
+    // a new one even with same port number.
+    remote_debugging_server_.reset();
```
pozdnyakov: would it be better to re-use the existing 'remote_debugging_server' rather than create new instance each time?

pozdnyakov:
lgtm

runtime/browser/xwalk_runner.cc
wuyuehang: Ok. The patch has been updated already. Thank you very much.

rakuco:
@baleboy: Does this need to be backported?

wuyuehang:
@rakuco Before dual process model, there're chances when entering such condition by invoking "xwalkctl -d port" way more often than "xwalk.service" way, especially on emulator sdk.

rakuco:
It's not clear how this answers my question. If it needs to be backported, then please backport it to the appropriate branches.

PR #2860, opened 2015-02-12
Shared library check using PackageManager

terriko:
This is a reworking of https://github.com/crosswalk-project/crosswalk/pull/2718.  In order to avoid the licensing issue, this makes the assumption that the PackageManager is trustworthy and checks that signature.

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/ReflectionHelper.java
@@ -7,11 +7,13 @@ package org.xwalk.core.internal;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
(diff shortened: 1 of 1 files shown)


baleboy:
Is this still required for the shared mode implementation, or was it replaced by another PR? @terriko @lincsoon

terriko:
I assumed that @lincsoon will be rolling it into a more comprehensive patch with actual signature checks, but I can clean this up so it's mergable if needed.

baleboy:
that might be the case, the overall patch about shared mode is https://github.com/crosswalk-project/crosswalk/pull/2780, let's wait for @lincsoon to comment

lincsoon:
Because this PR is based on old implementation, I already prepared another patch upon the approach from Terri in my new implementation of shared mode. As #2780 has been merged, I will soon send the PR about security check.

terriko:
Excellent, this can be closed then.

PR #2865, opened 2015-02-13
DEPS.xwalk: Roll blink-crosswalk (20dcb32 -> 921e4ad).

junmin-zhu:
921e4ad [Android] Fix WebCL fail to run on PowerVR GPU.


junmin-zhu:
@kenchris , @darktears , The roll patch for fix in blink. Thanks.

PR #2869, opened 2015-02-14
[Android] Disable the use of external startup data for V8.

wuhengzhi:
This patch is to disable the use of external startup data for V8
for android platform.
We just disable this feature to unblock the release at present.
Next step we will evaluate the tricky then decide to ship blobs
(for x86, arm, x86-64) or not.

BUG=XWALK-3516


rakuco:
lgtm, but please mention XWALK-3516 in those comments you've added to the code before landing, otherwise it will not be clear in the future that those files are disabled because of that ticket.

wuhengzhi:
@rakuco I have added XWALK-3516, thanks.

rakuco:
Please backport it to crosswalk-12 as well once it is verified that this solves the problem.

PR #2875, opened 2015-02-20
XWView{Internal}TestRunnerActivity: simplify hardware acceleration check.

rakuco:
`hardwareAccelerated` was always true, so there is no need to check for it, or for it to exist at all.


rakuco:
@wang16 @wuhengzhi

wuhengzhi:
LGTM, thanks.

PR #2877, opened 2015-02-23
[Tizen] Load application according to bundle values

tiwanek:
This commit passes encoded bundle to extension server
as 'encoded_bundle' variable. This value will be used
by application tizen extension.

Loading start page will check bundle before loading
'normal' starting URL. If application was launched
from appcontrol then appcontrol src is used to
launch application.

BUG=XWALK-1490

Requires:
 - https://review.tizen.org/gerrit/#/c/35215/
 - https://review.tizen.org/gerrit/#/c/35689/
 - https://review.tizen.org/gerrit/#/c/35700/

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -8,6 +8,7 @@
 
 #include "base/files/file_enumerator.h"
 #include "base/json/json_reader.h"
+#include "base/logging.h"
 #include "base/macros.h"
 #include "base/message_loop/message_loop.h"
 #include "base/stl_util.h"
@@ -102,7 +103,7 @@ Application::~Application() {
 }
 
 template<>
-GURL Application::GetStartURL<Manifest::TYPE_WIDGET>() {
+GURL Application::GetStartURL<Manifest::TYPE_WIDGET>() const {
 #if defined(OS_TIZEN)
   if (data_->IsHostedApp()) {
     std::string source;
@@ -132,7 +133,7 @@ GURL Application::GetStartURL<Manifest::TYPE_WIDGET>() {
 }
 
 template<>
-GURL Application::GetStartURL<Manifest::TYPE_MANIFEST>() {
+GURL Application::GetStartURL<Manifest::TYPE_MANIFEST>() const {
   if (data_->IsHostedApp()) {
     std::string source;
     // Not trying to get a relative path for the "fake" application.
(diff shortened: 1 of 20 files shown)


application/browser/application.h
```diff
@@ -112,6 +112,7 @@ class Application : public Runtime::Observer,
   Application(scoped_refptr<ApplicationData> data,
               XWalkBrowserContext* context);
   virtual bool Launch();
+  virtual GURL AlterStartURL(const GURL& url) const;
```
pozdnyakov: instead let's make instead 'virtual GURL GetStartURL(Manifest::Type type) const'

application/browser/application_service.cc
```diff
@@ -215,6 +215,16 @@ Application* ApplicationService::GetApplicationByID(
   return NULL;
 }
 
+bool ApplicationService::KillApplicationByID(const std::string& app_id) {
```
pozdnyakov: don't think it's needed: without shared process model it's enough just to quit/kill browser process.

tiwanek: I will remove this function by now  as it is used in TODO only.

application/browser/application.cc
```diff
@@ -161,6 +161,16 @@ GURL Application::GetStartURL<Manifest::TYPE_MANIFEST>() {
   return GURL();
 }
 
+GURL Application::GetStartURL(Manifest::Type type) const {
+  switch (type) {
+    case Manifest::Type::TYPE_WIDGET:
+      return GetStartURL<Manifest::Type::TYPE_WIDGET>();
+    case Manifest::Type::TYPE_MANIFEST:
+      return GetStartURL<Manifest::Type::TYPE_MANIFEST>();
+    default:
```
pozdnyakov: assert not reached

application/common/tizen/app_control_info.cc
```diff
@@ -0,0 +1,164 @@
...
+  if (!operation_str) {
+    // if no operation then bundle is invalid
+    service_destroy(service);
+    bundle_free(bundle);
+    return nullptr;
+  }
+  operation = operation_str;
+  free(operation_str);
+
+  char* mime_str = nullptr;
```
pozdnyakov: pls consider using scoped_ptr (with 'free' deleter)

pozdnyakov:
lgtm with a comment

application/common/tizen/app_control_info.cc
tiwanek: I need to pass memory address to C API so I cannot use smart pointers.

application/browser/application.cc
tiwanek: fixed

PR #2879, opened 2015-02-25
[Common]: Add window size configuration in manifest.

yejingfu:
We proposed to add recommended window bounds for w3c manifest (see https://github.com/w3c/manifest/issues/310).
1, This is the implementation in crosswalk common which allows user to set window preferred bounds at launch time.
2, It also allows user to set minimum size and maximum size of the window.

### DIFF
--- a/application/common/application_data.h
+++ b/application/common/application_data.h
@@ -20,6 +20,7 @@
 #include "base/strings/string_util.h"
 #include "base/synchronization/lock.h"
 #include "base/threading/thread_checker.h"
+#include "ui/gfx/geometry/rect.h"
 #include "url/gurl.h"
 #include "xwalk/application/common/manifest.h"
 #include "xwalk/application/common/permission_types.h"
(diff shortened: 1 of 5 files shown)


yejingfu:
@kenchris @baleboy Please take a review.

application/common/application_data.h
```diff
@@ -105,6 +105,8 @@ class ApplicationData : public base::RefCountedThreadSafe<ApplicationData> {
   const std::string& Name() const { return name_; }
   const std::string& NonLocalizedName() const { return non_localized_name_; }
   const std::string& Description() const { return description_; }
+  int WindowWidth() const { return window_width_; }
```
pozdnyakov: let's follow coding guidelines: int window_width() const

application/common/application_data.h
```diff
@@ -200,6 +203,11 @@ class ApplicationData : public base::RefCountedThreadSafe<ApplicationData> {
   // The source the application was loaded from.
   SourceType source_type_;
 
+  // The optional window size.
+  int window_width_;
```
pozdnyakov: it would be better to have some pair type for both width and height, gfx::Size maybe?

yejingfu:
@pozdnyakov and @kenchris I've updated the patch according to your feedback. Please take a review thanks.

pozdnyakov:
lgtm

yejingfu:
@kenchris I am planing to add the min-width / min-height and max-width / max-height in the next step, this is addressed in the commit log. Would you please merge this patch firstly?

yejingfu:
well, I've updated the patch by adding the minimum / maximum size and I also have basic tests to ensure it's working expected. Please take a review and merge if it's Okay for you. @kenchris @pozdnyakov

yejingfu:
@kenchris I have updated the changes according to your comments. Please take a look, thanks.

PR #2849, opened 2015-02-05
[Android] Don't specify compositor mode.

ds-hwang:
[Android] Don't specify compositor mode.
composite to mailbox is very old legacy. Don't specify this. CrossWalk relies
on upstream Android policy in terms of compositor mode.

BUG=XWALK-322

PR #2883, opened 2015-02-26
Revert "[Android] Enable Experimental WebPlatform Features"

rakuco:
Enabling the entire set of web platform features currently marked
experimental in Blink is dangerous. The features in this set changes
across Chromium releases, and it can be the case that some of them just
are not ready for production at all.

Case in point: this enables support for the @viewport CSS rule, which
currently has problems and causes regressions for our users (see
XWALK-3574).

If we are interested in enabling certain features (the commit message of
the change being reverted mentions CSS Animations, for example), they
should be enabled individually.

This reverts commit 9255613355cf56547953f2ac816fbc4581659a49.

Related to: XWALK-322
BUG=XWALK-3574

### DIFF
--- a/runtime/browser/xwalk_browser_main_parts_android.cc
+++ b/runtime/browser/xwalk_browser_main_parts_android.cc
@@ -116,10 +116,6 @@ void XWalkBrowserMainPartsAndroid::PreMainMessageLoopStart() {
   command_line->AppendSwitch(switches::kIgnoreGpuBlacklist);
 #endif
 
-  // Enable experiemntal features like polymer and css animations because
-  // CrossWalk is testbed of state of art of web technology.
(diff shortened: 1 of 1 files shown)


rakuco:
@ds-hwang @kenchris @baleboy 
This reverts part of #2849 to fix XWALK-3574. It is not clear to me whether this commit was included in that pull request by mistake or not, as it does not have anything to do with the compositor mode.

axinging:
lgtm

ds-hwang:
thank you for fixing this regression. lgtm
@kenchris is your mwc demo fine without this patch. especially, `res.style.animation = "result-in 0.5s forwards";`

kenchris:
yes, I just use prefixed versions

PR #2885, opened 2015-02-28
[Android] Fix remote debugging doesn't work issue after M41 rebasing

huningxin:
Remote protocol handler should now report "inspector.html", according to
upstream change: https://crrev.com/a1409c477411b0467b4bc729d9a2c972e79347d1

BUG=XWALK-3522


huningxin:
@darktears @rakuco , please take a look. Thanks!

PR #2888, opened 2015-03-02
Add loading Pepper Flash plugin from command line support.

seanlong:
In order to support sites using Adobe Flash on desktop we need to
support loading external Flash plugins. This patch added support to
loading Pepper Flash plugin specified from command line. The code
is mostly taken from Chromium chrome/common/chrome_content_client.cc.

The command line switches are the same with Chromium:
--ppapi-flash-path: full path of plugin library
--ppapi-flash-version: optional plugin version

### DIFF
--- a/runtime/browser/runtime.cc
+++ b/runtime/browser/runtime.cc
@@ -301,4 +301,13 @@ void Runtime::RequestMediaAccessPermission(
       web_contents, request, callback);
 }
 
+bool Runtime::CheckMediaAccessPermission(
+    content::WebContents* web_contents,
+    const GURL& security_origin,
+    content::MediaStreamType type) {
+  // TODO(xiang): Pepper flash plugin will trigger this check a lot, return
+  // false at the moment.
+  return false;
+}
+
 }  // namespace xwalk
--- a/runtime/common/xwalk_content_client.cc
+++ b/runtime/common/xwalk_content_client.cc
@@ -1,3 +1,4 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
 // Copyright (c) 2013 Intel Corporation. All rights reserved.
 // Copyright (c) 2014 Samsung Electronics Co., Ltd All Rights Reserved
 // Use of this source code is governed by a BSD-style license that can be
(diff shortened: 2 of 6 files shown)


seanlong:
@pozdnyakov @halton PTAL, thanks.

seanlong:
ping @pozdnyakov, thanks!

runtime/browser/runtime.cc
```diff
@@ -301,4 +301,13 @@ void Runtime::RequestMediaAccessPermission(
       web_contents, request, callback);
 }
 
+bool Runtime::CheckMediaAccessPermission(
```
pozdnyakov: pls do not add empty methods

runtime/common/xwalk_content_client.cc
```diff
@@ -45,6 +51,69 @@ const char kPnaclPluginExtension[] = "";
...
+  content::PepperPluginInfo plugin;
+
+  plugin.is_out_of_process = true;
+  plugin.name = content::kFlashPluginName;
+  plugin.path = path;
+  plugin.permissions = kPepperFlashPermissions;
+
+  std::vector<std::string> flash_version_numbers;
+  base::SplitString(version, '.', &flash_version_numbers);
+  if (flash_version_numbers.size() < 1)
```
pozdnyakov: this looks overcomplicated.. can we just check it once for being empty?

runtime/browser/runtime.cc
seanlong: It's not empty as commented, it was added in order to avoid flood of "Not supported" msg.

runtime/common/xwalk_content_client.cc
seanlong: These functions are just copied from Chromium, I will update the commit msg. Thanks.

runtime/common/xwalk_content_client.cc
```diff
@@ -1,3 +1,4 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
```
halton: nit, any reason to add an old copyright that does not belong to intel.

runtime/common/xwalk_switches.cc
```diff
@@ -1,3 +1,4 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
```
halton: ditto

runtime/common/xwalk_content_client.cc
seanlong: I just updated the commit msg, and please check this comment https://github.com/crosswalk-project/crosswalk/pull/2888/files#r25746765.

halton: Not sure it is right to add this copyright because we copied code from upstream. cc @baleboy

rakuco: What do you mean? It sounded like if the code comes from upstream the copyright notice should not be preserved, which I guess is not what you intended to convey.

runtime/browser/runtime.cc
rakuco: Isn't it better to return `false` if you just want to avoid the warnings then? Otherwise there's no check for the security origin and all checks will just pass.

runtime/common/xwalk_content_client.cc
halton: Okay

runtime/browser/runtime.cc
seanlong: Done, return false make more sense. Thanks.

halton:
LGTM

pozdnyakov:
lgtm

PR #2889, opened 2015-03-02
[Tizen] Enable HWKey emulation for all native platform

WonyoungChoi:
Use EventFromNative() function of chromium to emulate hardware
key events from all native platform(x/win/ozone).

Currently, HWKey doesn't work in X11 environment.

BUG=XWALK-3642


pozdnyakov:
lgtm

PR #2891, opened 2015-03-03
Fix ApplicationData::LoadID

tiwanek:
Calling ApplicationData::LoadID before finished_parsing_manifest_
is set to true cause DCHECK to fail.

GetManifestData is expected to be used by client of this class
but not internally.


tiwanek:
I have missed the thread_checker_.CalledOnValidThread() value. I assume this means that "getter" can be used in same thread or after parsing in other threads. So it seems ok.

PR #2893, opened 2015-03-05
Add manifest "start_url" field support for web URL.

seanlong:
Before this change "start_url" filed only support local relative URL,
after this it will also support http/https web URLs

BUG=XWALK-3656

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -141,11 +141,18 @@ GURL Application::GetStartURL<Manifest::TYPE_MANIFEST>() {
     return GURL();
   }
 
-  GURL url = GetAbsoluteURLFromKey(keys::kStartURLKey);
-  if (url.is_valid())
-    return url;
+  std::string start_url_source;
+  if (data_->GetManifest()->GetString(keys::kStartURLKey, &start_url_source) &&
+      !start_url_source.empty()) {
+    GURL url(start_url_source);
+    if (url.is_valid() && url.SchemeIsHTTPOrHTTPS())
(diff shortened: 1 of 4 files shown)


seanlong:
@pozdnyakov @kenchris PTAL, thanks.

pozdnyakov:
From the http://w3c.github.io/manifest/#start_url-member:

"If url is not same origin as document URL:
    Issue a developer warning that the start_url needs to be same-origin as Document of the top-level browsing context.
    Return a new URL whose input is document URL."
So looks like http/https web URLs cannot be there

seanlong:
Yes, I raised the same question to @kenchris in mail, he replied it would be OK.
@kenchris would you confirm this? Thanks.

takethathe:
@pozdnyakov @kenchris @seanlong 
How about to use http://w3c.github.io/manifest/#scope-member

kenchris:
the spec is written with hosted apps in mind. We do packaged apps (and so will Microsoft for Win10). I think that we need to find out how we can support this @anssik. We need to think about how to avoid man-in-the-middle attacks ect, restriction to https etc.

I can talk to @pozdnyakov @poussa @anssik next week.

rakuco:
I think you mean @anssiko ;-)

seanlong:
@kenchris Do you have decision on this now?
@takethathe I think the "scope" is to specify the manifest working scope which is different from start URL.

kenchris:
Let's follow MS's lead here

PR #2895, opened 2015-03-05
generate_xwalk_core_library: Ignore certain files from resources.

rakuco:
This is in preparation for M42, where `content_java.zip` includes an
OWNERS file in the `res/` directory that causes apk generation with
make_apk.py to fail.

We now keep a blacklist of files that we do not want to copy in
`CopyDirAndPrefixDuplicates()`.


rakuco:
@lincsoon @halton

halton:
LGTM

lincsoon:
LGTM

PR #2896, opened 2015-03-05
Build fix on Windows

darktears:


rakuco:
What's the error? It's weird that Windows is only complaining about this `switch`.

PR #2897, opened 2015-03-06
tests: Always call document.close() after using document.write().

rakuco:
This commit is in preparation for M42.

Blink has recently changed the definition of load completion, which exposed some flakiness in our HTML tests: `document.write()` keeps the document stream open and can cause Blink to never consider it has finished loading the document, causing time outs in our tests.

[Chromium bug 426520](https://code.google.com/p/chromium/issues/detail?id=426520) contains more information and a list of commits related to the Blink changes.

While `document.write()` is not strictly needed by the majority of the tests that use it, I chose to keep the calls and just explicitly close the document in most cases, those calls provide some information on what the test is doing (though we could consider replacing those calls with something similar to what Blink does in its layout tests).

### DIFF
--- a/extensions/test/data/iframe_using_document_write.html
+++ b/extensions/test/data/iframe_using_document_write.html
@@ -7,6 +7,7 @@ var c = document.createElement("iframe");
 document.body.appendChild(c);
 c.contentDocument.write('lala');
 c.src = "counter.html";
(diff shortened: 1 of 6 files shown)


rakuco:
@darktears @kenchris

darktears:
LGTM

PR #2898, opened 2015-03-10
Enable open external url in web browser.

takethathe:
The XPK package format should support 'scope' field in manifest.json, which is defined in spec: 
https://w3c.github.io/manifest/#scope-member, this patch will enable this feature for XPK package, and open an external link in web browser.

BUG=XWALK-3690

### DIFF
--- a/application/browser/application_security_policy.cc
+++ b/application/browser/application_security_policy.cc
@@ -76,9 +76,9 @@ bool ApplicationSecurityPolicy::IsAccessAllowed(const GURL& url) const {
   if (!enabled_)
     return true;
 
-  // Accessing own resources is always allowed.
-  if (url.SchemeIs(application::kApplicationScheme) &&
-      url.host() == app_->id())
+  // Accessing own resources in W3C Widget is always allowed.
+  if (app_->data()->manifest_type() == Manifest::TYPE_WIDGET &&
+      url.SchemeIs(application::kApplicationScheme) && url.host() == app_->id())
     return true;
 
   for (std::vector<WhitelistEntry>::const_iterator it =
@@ -87,7 +87,8 @@ bool ApplicationSecurityPolicy::IsAccessAllowed(const GURL& url) const {
     bool subdomains = it->subdomains;
     bool is_host_matched = subdomains ?
         url.DomainIs(policy.host().c_str()) : url.host() == policy.host();
-    if (url.scheme() == policy.scheme() && is_host_matched)
+    if (url.scheme() == policy.scheme() && is_host_matched &&
+        StartsWithASCII(url.path(), policy.path(), false))
       return true;
   }
   return false;
@@ -185,11 +186,10 @@ void ApplicationSecurityPolicyCSP::Enforce() {
         app_->data()->GetManifestData(widget_keys::kAllowNavigationKey));
     if (info) {
       const std::vector<std::string>& allowed_list = info->GetAllowedDomains();
-      for (std::vector<std::string>::const_iterator it = allowed_list.begin();
-           it != allowed_list.end(); ++it) {
+      for (const auto& it : allowed_list) {
         // If the policy is "*", it represents that any external link is allowed
         // to navigate to.
-        if ((*it) == kAsterisk) {
+        if (it == kAsterisk) {
           enabled_ = false;
           return;
         }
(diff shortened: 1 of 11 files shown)


takethathe:
@kenchris @pozdnyakov 

runtime/browser/runtime_platform_util_linux.cc
```diff
@@ -43,7 +43,13 @@ void XDGUtil(const std::string& util, const std::string& arg) {
 }
 
 void XDGOpen(const std::string& path) {
-  XDGUtil("xdg-open", path);
+  GURL url(path);
+  if (url.is_valid()) {
+    std::string cmd = "xdg-open " + path;
+    std::system(cmd.c_str());
```
pozdnyakov: when this happens?

application/browser/application_security_policy.cc
```diff
@@ -206,22 +208,56 @@ void ApplicationSecurityPolicyCSP::Enforce() {
...
+        if (parse_result != URLPattern::PARSE_SUCCESS) {
+          LOG(INFO) << "URLPattern " << match_pattern << " is wrong: "
+                    << URLPattern::GetParseResultString(parse_result);
+          continue;
+        }
+
+        for (size_t j = 0; j < arraysize(schemes); ++j) {
+          if (allowedUrl.MatchesScheme(schemes[j]))
+            AddWhitelistEntry(GURL(std::string(schemes[j]) +
+                                   std::string("://") + allowedUrl.host()),
```
pozdnyakov: I think there is a const var in chromium for '://'

application/browser/application_security_policy.cc
```diff
@@ -206,22 +208,56 @@ void ApplicationSecurityPolicyCSP::Enforce() {
...
 #endif
   } else {
-    if (!csp_info || csp_info->GetDirectives().empty()) {
-      LOG(ERROR) << "Failed to obtain CSP directives from the manifest";
-      return;
+    if (csp_info && !csp_info->GetDirectives().empty()) {
+      enabled_ = true;
+      const std::map<std::string, std::vector<std::string> >& policies =
+          csp_info->GetDirectives();
+      std::map<std::string, std::vector<std::string> >::const_iterator it =
```
pozdnyakov: auto

application/browser/application_security_policy.cc
```diff
@@ -206,22 +208,56 @@ void ApplicationSecurityPolicyCSP::Enforce() {
     }
 #endif
   } else {
-    if (!csp_info || csp_info->GetDirectives().empty()) {
-      LOG(ERROR) << "Failed to obtain CSP directives from the manifest";
-      return;
+    if (csp_info && !csp_info->GetDirectives().empty()) {
+      enabled_ = true;
+      const std::map<std::string, std::vector<std::string> >& policies =
```
pozdnyakov: auto

application/browser/application_security_policy.cc
```diff
@@ -206,22 +208,56 @@ void ApplicationSecurityPolicyCSP::Enforce() {
...
+          csp_info->GetDirectives();
+      std::map<std::string, std::vector<std::string> >::const_iterator it =
+          policies.begin();
+      for (; it != policies.end(); ++it) {
+        const std::vector<std::string>& allowed_list = it->second;
+        for (std::vector<std::string>::const_iterator it = allowed_list.begin();
```
pozdnyakov: auto

runtime/browser/runtime_platform_util_linux.cc
takethathe: The url like http://, https:// will use this, when open the external url in web browser, 'base::LaunchProcess' can not open the url in web browser when already has a running browser, so use 'std:system' instead.

I'll change to check whether the scheme is http or https.

takethathe:
@pozdnyakov 
I've changed the patch according to your comments, PTAL, thanks!

kenchris:
External? Actually can we implement the scope part of the manifest spec and open everything outside the scope in an external browseR?

r0b5t4:
@kenchris how does the manifest in that regard interact with the html target attribute?

takethathe:
@kenchris 
Think the 'scope' is not enough, it can only support one domain currently, 'xwalk_hosts' support a list of domains which are granted the access permission for web application.

takethathe:
@kenchris 
I've changed the patch, use 'scope' field instead of 'xwalk_hosts' for this feature, PTAL! Thanks!

application/browser/application_security_policy.cc
```diff
@@ -198,32 +198,42 @@ void ApplicationSecurityPolicyCSP::Enforce() {
...
+          GURL url(it);
+          if (url.is_valid())
+            AddWhitelistEntry(url, false);
+        }
       }
     }
+
+    // scope
+    std::string scope;
+    GURL allowedUrl(ApplicationData::GetBaseURLFromApplicationId(app_->id()));
```
kenchris: I guess allowedurl is a bit confusing, because it basically represents urls that are considered part of the app.

internalUrl?

application/browser/application_security_policy.cc
```diff
@@ -198,32 +198,42 @@ void ApplicationSecurityPolicyCSP::Enforce() {
...
+    if (app_->data()->GetManifest()->GetString(keys::kScopeKey, &scope) &&
+        !scope.empty()) {
+      enabled_ = true;
+      url::Replacements<char> replacements;
+      replacements.SetPath(scope.c_str(), url::Component(0, scope.length()));
+      allowedUrl = allowedUrl.ReplaceComponents(replacements);
+    }
+    if (allowedUrl.is_valid())
+      AddWhitelistEntry(allowedUrl, false);
```
kenchris: Add a comment that non whilelisted entries are opened externally by the system (e.g. by the browser for HTTP and HTTPS) ?

takethathe:
@kenchris @pozdnyakov 
I've changed the patch, PTAL! Looks like the trybot failure for Android platform is not related to this patch.

application/browser/application_security_policy.cc
```diff
@@ -76,9 +76,9 @@ bool ApplicationSecurityPolicy::IsAccessAllowed(const GURL& url) const {
   if (!enabled_)
     return true;
 
-  // Accessing own resources is always allowed.
-  if (url.SchemeIs(application::kApplicationScheme) &&
-      url.host() == app_->id())
+  // Accessing own resources in W3C Widget is always allowed.
+  if (app_->data()->manifest_type() == Manifest::TYPE_WIDGET &&
```
pozdnyakov: why this new condition is necessary?

pozdnyakov:
lgtm

application/browser/application_security_policy.cc
takethathe: After enable 'scope' field, the old check is not enough to XPK package in some conditions, it has to check the url 'path' if there's a 'scope' set in manifest.json.

rakuco:
`std::system()`, really? I mean, **really**?!

This does not even _build_ with GCC 4.9 on Debian:
```
../../xwalk/runtime/browser/runtime_platform_util_linux.cc: In function ‘void {anonymous}::XDGOpen(const string&)’:
../../xwalk/runtime/browser/runtime_platform_util_linux.cc:49:29: error: ignoring return value of ‘int system(const char*)’, declared with attribute warn_unused_result [-Werror=unused-result]
```

> 'base::LaunchProcess' can not open the url in web browser when already has a running browser, so use 'std:system' instead.

I don't buy this argument, and I've just opened two Konqueror windows here with that part of the patch reverted. This is basically saying you can't run `base::LaunchProcess()` with the same arguments more than once.

This part of the change **must** be reverted.

darktears:
@takethathe : can you have a look please?

takethathe:
@darktears 
I have tested for chrome browser, the ```base::LaunchProcess()``` does not work at all when there already has a running browser instance. And I'll try to fix the compiling issue.

darktears:
@takethathe we can't let the system() call, this is not negociable, there must be another way to do something different.

pozdnyakov:
My fall as a reviewer :(  https://github.com/crosswalk-project/crosswalk/pull/2955 - partial revert, a better solution has to be found.

PR #2899, opened 2015-03-11
[Android] Fix immersive full screen mode doesn't recover.

wuhengzhi:
Immersive Fullscreen mode does not recover after changing the volume on
the device.
Refer https://github.com/fujunwei/crosswalk-cordova-android/commit/c9e30e77cf71da2d189faecd861c29c04a1548f6
to fix this issue.

BUG=XWALK-3651, XWALK-2367

### DIFF
--- a/app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
+++ b/app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
@@ -5,9 +5,10 @@
 package org.xwalk.app.template;
 
 import android.graphics.Color;
-import android.os.Build.VERSION;
-import android.os.Build.VERSION_CODES;
+import android.os.Build;
 import android.os.Bundle;
+import android.util.Log;
+import android.view.Window;
 import android.view.WindowManager;
 import android.view.View;
 import android.widget.TextView;
@@ -15,9 +16,23 @@ import android.widget.TextView;
 import org.xwalk.app.XWalkRuntimeActivityBase;
 
 public class AppTemplateActivity extends XWalkRuntimeActivityBase {
+    private final int SYSTEM_UI_OPTIONS = View.SYSTEM_UI_FLAG_LAYOUT_STABLE |
+            View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
+            View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
+            View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
+            View.SYSTEM_UI_FLAG_FULLSCREEN |
+            View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY;
+    private final String TAG = this.getClass().getName();
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+    }
+
+    @Override
+    public void onResume() {
+        super.onResume();
+        enterFullscreen();
     }
 
     @Override
@@ -35,24 +50,39 @@ public class AppTemplateActivity extends XWalkRuntimeActivityBase {
     }
 
     private void enterFullscreen() {
-        if (VERSION.SDK_INT >= VERSION_CODES.KITKAT &&
-                ((getWindow().getAttributes().flags &
-                        WindowManager.LayoutParams.FLAG_FULLSCREEN) != 0)) {
+        if (isNewerThanKitkat() && ((getWindow().getAttributes().flags &
+                WindowManager.LayoutParams.FLAG_FULLSCREEN) != 0)) {
             View decorView = getWindow().getDecorView();
-            decorView.setSystemUiVisibility(
-                    View.SYSTEM_UI_FLAG_LAYOUT_STABLE |
-                    View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
+            decorView.setSystemUiVisibility(SYSTEM_UI_OPTIONS);
         }
     }
 
     public void setIsFullscreen(boolean isFullscreen) {
-        if (isFullscreen) {
-            enterFullscreen();
+        if (!isFullscreen) return;
+        if (!isNewerThanKitkat()) {
+            Log.w(TAG, "setIsFullscreen() should not be called if " +
+                    "android version is older than Kitkat.");
+            return;
         }
+
+        setSystemUiVisibilityChangeListener();
     }
 
+    private void setSystemUiVisibilityChangeListener() {
+        final View decorView = getWindow().getDecorView();
+        decorView.setOnSystemUiVisibilityChangeListener(
+            new View.OnSystemUiVisibilityChangeListener() {
+                @Override
+                public void onSystemUiVisibilityChange(int visibility) {
+                    if ((visibility & View.SYSTEM_UI_FLAG_FULLSCREEN) == 0) {
+                        decorView.setSystemUiVisibility(SYSTEM_UI_OPTIONS);
+                    }
+                }
+            }
+        );
+    }
+
+    private boolean isNewerThanKitkat() {
+        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT;
+    }
 }


app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -34,25 +38,44 @@ protected void didTryLoadRuntimeView(View runtimeView) {
         }
     }
 
-    private void enterFullscreen() {
-        if (VERSION.SDK_INT >= VERSION_CODES.KITKAT &&
```
axinging: LGTM. One suggestion:
    private boolean isNewerThanKitkat() {
        return Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT;
    }

wuhengzhi:
@fujunwei "enterFullscreen()" was deleted since it's duplicate with "toggleFullscreen()". "setIsFullscreen()" was not deleted.

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -15,9 +15,23 @@
...
+            View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
+            View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
+            View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
+            View.SYSTEM_UI_FLAG_FULLSCREEN |
+            View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY;
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+        setSystemUiVisibilityChangeListener();
```
axinging: Hi, we do not need to listen to system ui visibility when setIsFullscreen(true); not called.
So please put setSystemUiVisibilityChangeListener into setSystemUiVisibilityChangeListener:
    public void setIsFullscreen(boolean isFullscreen) {
        if (isFullscreen) {
            enterFullscreen();
            setSystemUiVisibilityChangeListener();
        }
    }

baleboy:
@lincsoon @darktears @rakuco can you review?

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -35,24 +48,35 @@ protected void didTryLoadRuntimeView(View runtimeView) {
...
         if (isFullscreen) {
             enterFullscreen();
+            setSystemUiVisibilityChangeListener();
         }
     }
 
+    private void setSystemUiVisibilityChangeListener() {
+        if (isNewerThanKitkat()) {
+            final View decorView = getWindow().getDecorView();
+            decorView.setOnSystemUiVisibilityChangeListener(new View.OnSystemUiVisibilityChangeListener() {
```
lincsoon: Line length exceeded limit

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -35,24 +48,35 @@ protected void didTryLoadRuntimeView(View runtimeView) {
...
 
     public void setIsFullscreen(boolean isFullscreen) {
         if (isFullscreen) {
             enterFullscreen();
+            setSystemUiVisibilityChangeListener();
         }
     }
 
+    private void setSystemUiVisibilityChangeListener() {
+        if (isNewerThanKitkat()) {
```
lincsoon: setSystemUiVisibilityChangeListener should be a pure set method. It's better to remove the version check outside.

```
if (isNewerThanKitkat())
    setSystemUiVisibilityChangeListener()
```

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -35,24 +48,35 @@ protected void didTryLoadRuntimeView(View runtimeView) {
...
-                    View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
+            decorView.setSystemUiVisibility(SYSTEM_UI_OPTIONS);
         }
     }
 
     public void setIsFullscreen(boolean isFullscreen) {
```
lincsoon: I know it's not about this patch. But why setIsFullscreen have to take a boolean parameter even it will do nothing if the isFullscreen is false?

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -15,12 +15,25 @@
...
+
     @Override
     public void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
     }
 
     @Override
+    public void onResume() {
+        super.onResume();
+        enterFullscreen();
```
lincsoon: Is this necessary? As my test, even the home key is hidden with this change.

wuhengzhi: @lincsoon You mean home key is hidden without this change? I have tested this function on emulator(android 4.4.2) with skin WVGA800 which has physical keys on the right.
Without calling enterFullscreen() in onResume:
1. Launch app, the virtual home key was always display.
2. If we press the dial key, then back with return key, the virtual home key does not disappear.
3. If we press the home key, then resume app, the virtual home key does not disappear.

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
wuhengzhi: ```setIsFullscreen``` will be added to onCreate when fullscreeen flag was specified. The parameter was specified in customize.py.
```
  if app_info.fullscreen_flag:
    SetVariable(dest_activity,
                'super.onCreate(savedInstanceState)',
                'IsFullscreen', 'true')
```
```SetVariable``` is general for other flags. Thanks.

File: app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java (hunk 6 above)
Comment (lincsoon) [inline app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java]: I mean the virtual keys(back, home, switch) at the bottom is hidden if call enterFullscreen here in my test

File: app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java (hunk 5 above)
Comment (lincsoon) [inline app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java]: It's also the customize.py's problem and we should make it more flexible. But I think this is a minor problem here and no hurry to fix in this patch.

File: app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java (hunk 6 above)
Comment (wuhengzhi) [inline app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java]: In activity of your test app, add ```onResume``` and call ```enterFullscreen```? Do you thinks it's no necessary to add ```onResume``` in AppTemplateActivity.java? Thanks.

Comment (lincsoon) [inline app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java]: OK, looks like I have some misunderstanding about full screen mode

File: app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java [hunk 7]
```diff
@@ -35,24 +48,37 @@ protected void didTryLoadRuntimeView(View runtimeView) {
...
-                    View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_HIDE_NAVIGATION |
-                    View.SYSTEM_UI_FLAG_FULLSCREEN |
-                    View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY);
+            decorView.setSystemUiVisibility(SYSTEM_UI_OPTIONS);
         }
     }
 
     public void setIsFullscreen(boolean isFullscreen) {
         if (isFullscreen) {
```
lincsoon: If the setIsFullScreen is supported only after Kitkat, it'll be more clear to check the version at the beginning.

```
if (!isNewerThanKitkat() || !isFullScreen) return;

enterFullscreen();
setSystemUiVisibilityChangeListener();
```
then remove the version check from enterFullScreen.

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
```diff
@@ -35,24 +48,37 @@ protected void didTryLoadRuntimeView(View runtimeView) {
...
+            }
         }
     }
 
+    private void setSystemUiVisibilityChangeListener() {
+        final View decorView = getWindow().getDecorView();
+        decorView.setOnSystemUiVisibilityChangeListener(
+            new View.OnSystemUiVisibilityChangeListener() {
+                @Override
+                public void onSystemUiVisibilityChange(int visibility) {
```
lincsoon: This listener does the same things with enterFullscreen in the `onResume()`. What kind of situation it will cover and `onResume()` doesn't?

lincsoon:
If the --fullscreen option is enabled, there will be additional `setIsFullScreen(true)` in `onCreate()`. Since enterFullScreen is always called in `onResume()`, maybe we could skip it in `onCreate()`?

app/android/app_template/src/org/xwalk/app/template/AppTemplateActivity.java
wuhengzhi: Press volume button, ```onSystemUiVisibilityChange``` will be called, ```onResume()``` won't.
If the dial key was pressed, ```onSystemUiVisibilityChange``` will be called, presss back key to return, ```onSystemUiVisibilityChange``` will not be called again since the visibility was not changed, so call ```enterFullScreen``` in ```onResume```.

lincsoon: Thanks for the explanation!

wuhengzhi:
@lincsoon Thanks a lot for your review. I have updated the code.

lincsoon:
LGTM

lincsoon:
@rakuco

rakuco:
Please backport if necessary.

PR #2901, opened 2015-03-13

fujunwei:

BUG=XWALK-2132


halton:
Does this change also affect  Android? https://lists.crosswalk-project.org/pipermail/crosswalk-help/2015-March/000913.html is asking same ability.

fujunwei:
Maybe. I will reply it after the PR is merged

fujunwei:
@halton @lincsoon @axinging  PTAL. Thanks

halton:
LGTM, but you need work together with @rakuco to fix the trybot failures.

axinging:
LGTM

runtime/browser/xwalk_content_browser_client.cc
```diff
@@ -429,4 +430,8 @@ content::DevToolsManagerDelegate*
   return new XWalkDevToolsDelegate(browser_context_);
 }
 
+std::string XWalkContentBrowserClient::GetApplicationLocale() {
+  return base::android::GetDefaultLocale();
```
rakuco: Isn't this Java-specific?

halton:
@fujunwei why delete your branch, thus this PR is closed.

fujunwei:
I confuse why it can't compiled successfully.

fujunwei:
i want to send a new PR instead of this.

PR #2902, opened 2015-03-16
Support to replace user agent string for XPK package

takethathe:
The Crosswalk web runtime should support to replace the user agent string when a
web application running on top of it, so that the remote server could return the
right response to the web application as expecting. The patch will enable this
feature by adding the 'xwalk_user_agent' field in manifest.json file on all
platforms except the Android.

BUG=XWALK-3689

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -19,14 +19,20 @@
 #include "content/public/browser/render_process_host.h"
 #include "content/public/browser/site_instance.h"
 #include "net/base/net_util.h"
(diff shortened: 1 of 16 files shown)


takethathe:
@pozdnyakov @kenchris 
PTAL, thanks!

kenchris:
I don't really think this is a good idea. It is a hack that if we enable, it will be very hard to get rid of

* use start_url instead, by adding an URL argument like http://www.mysite.com?launch=crosswalk

PR #2903, opened 2015-03-16
Move to Chromium 42.0.2311.39.

rakuco:
As usual, this includes a lot of changes to make the code work with a
new Chromium milestone (files got moved, classes were changed etc).

SIMD.js and XDK support are already enabled and working.

Important things to note:
 * ChannelMojo support is being disabled on Android while we investigate
   XWALK-3650.

 * Upstream has introduced a dependency on Google Play Services from the
   Android SDK for the Android build that is a pain to automate, as
   users are expected to accept its EULA first.
   Since Crosswalk does not depend on the Play Services for now, some
   changes were made to chromium-crosswalk to skip this dependency while
   we think of a better way of handling this with upstream.


rakuco:
@darktears

darktears:
LGTM

PR #2905, opened 2015-03-17
Fix wrong nacl plugin gyp location

halton:
src/trusted/plugin/ is moved to components/nacl/renderer/plugin/ by
CL https://codereview.chromium.org/876483002

PR #2906, opened 2015-03-17
Roll DEPS.xwalk to latest chromium/blink/v8/ozone master.

halton:


halton:
@rakuco @darktears . I am not sure why those dependency are not updated with master. Please let me know if it is dup work.

rakuco:
It is -- this is part of #2903, which moves Crosswalk to Chromium M42. All changes to make the code build are supposed to go in together (instead of breaking it into separate PRs like this one and #2905), otherwise we have a broken tree until all commits are merged.

PR #2908, opened 2015-03-18
[Android] Fix the display for the website's title after navigation.

wuhengzhi:
This patch is to display the website's title when open multiple URLs and
press the back or forward button.
The event of update title will be fired in render frame host and title will be
compared to the one in navigation entry, if they are same, the update
title event will not be sent to Java layer.

BUG=XWALK-2333


rakuco:
Don't we need tests?

wuhengzhi:
@rakuco The fixing is for displaying the title on tab in XWalkCoreShell, I have verified the title manually.
Refine the test case for navigation and add test case for onLoadFinished to ensure the title is identical.

wuhengzhi:
@lincsoon @axinging @fujunwei PTAL. Thanks

axinging:
LGTM. Thanks for quick patch.

lincsoon:
LGTM

lincsoon:
@rakuco

rakuco:
Please backport if needed.

PR #2909, opened 2015-03-18
[Tizen] Do not add NaCl's toolchain to the source tarball.

rakuco:
Currently NaCl support is not enabled on Tizen, so we can save a few
gigabytes in the source tarball and the extracted archive if we do not
package NaCl's `toolchain/` directory.


darktears:
YEAH

triple-it:
Ok.. trying android NDK

PR #2911, opened 2015-03-19
[Tizen] Fix parser handlers

tiwanek:
In current execution flow, there is call of Parse() function
for each handler. Handler should be prepared for situation that
there is no value it expects.

Following handlers are broken:
 - AppControl,
 - Category,
 - Ime,
 - Metadata.

BUG=XWALK-3774

### DIFF
--- a/application/common/manifest_handlers/tizen_metadata_handler.cc
+++ b/application/common/manifest_handlers/tizen_metadata_handler.cc
@@ -66,9 +66,9 @@ bool TizenMetaDataHandler::Parse(scoped_refptr<ApplicationData> application,
   const Manifest* manifest = application->GetManifest();
   DCHECK(manifest);
 
-  base::Value* metadata_value = NULL;
+  base::Value* metadata_value = nullptr;
   if (!manifest->Get(keys::kTizenMetaDataKey, &metadata_value)) {
-    *error = base::ASCIIToUTF16("Failed to get value of tizen metaData");
+    return true;
   }
 
   MetaDataPair metadata_item;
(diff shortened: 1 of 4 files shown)


application/common/manifest_handlers/tizen_metadata_handler.cc
```diff
@@ -66,9 +66,9 @@ bool TizenMetaDataHandler::Parse(scoped_refptr<ApplicationData> application,
   const Manifest* manifest = application->GetManifest();
   DCHECK(manifest);
 
-  base::Value* metadata_value = NULL;
+  base::Value* metadata_value = nullptr;
   if (!manifest->Get(keys::kTizenMetaDataKey, &metadata_value)) {
-    *error = base::ASCIIToUTF16("Failed to get value of tizen metaData");
+    return true;
```
rakuco: Shouldn't you return `false` instead? And is it really an improvement not to set `error`? If that's the case the parameter could be removed altogether.

tiwanek: If I correctly understood, in the situation that there is no metadata key, this branch of execution is chosen and it should return positive result, not negative -> metadata is not obligatory.

rakuco: I see. How about the `error` part? Is it better to remove it altogether or add it to the other `Parse()` calls?

tiwanek: Methods Get() and HasPath() calls CanAccessPath() internally so I made choise just to call one.
Setting error is not needed.

tiwanek:
@rakuco 
Would you accept current solution?

rakuco:
I think so. @pozdnyakov?

pozdnyakov:
lgtm

PR #2914, opened 2015-03-23
[Android] No need to disable ChannelMojo in XWalk

leonhsl:
As we've fixed the root case of bug XWALK-3650 on upstream side:
https://codereview.chromium.org/1003123002/
and currently upstream side disabled ChannelMojo,
and, we've also done back port the disablement commit to Chromium-Crosswalk:
https://github.com/crosswalk-project/chromium-crosswalk/pull/237
now no need to disable ChannelMojo in XWalk side anymore.

BUG=XWALK-3650


leonhsl:
PTAL @halton @rakuco

halton:
LGTM

PR #2916, opened 2015-03-24
[Android] Shared mode implementation

lincsoon:
Shared mode allows multiple xwalk apps to share one xwalk runtime
library. Each xwalk app is bundled with a reflection layer instead
of the whole runtime library, whereas there is a library apk to be
installed on device for xwalk apps to use.

BUG=XWALK-2948


lincsoon:
On the basis of #2780 , this patch fixed error of instrumentation test. Please see original pull request to check reviews and discussions.

halton:
@lincsoon could you mark the delta change to fix the testing issue? Thus will save reviewer time.

lincsoon:
@halton I separate original patch to 2 commits. Please help to reivew

lincsoon:
The reflection layer will be initialized before all things on both of embedded mode and shared mode. But in test of XWalkCoreInternalShell, this step is skipped because there's no reflection layer. In this case, the instance of XWalkCoreBridge is null.

halton:
LGTM

rakuco:
Thank you for acting so promptly. Could you please re-squash the commits together? If that's not done, we will have one commit in the tree that is knowingly bootleg and make bisecting and reverting more difficult.

rakuco:
:+1:

lincsoon:
Thanks for everybody's quick response!

PR #2918, opened 2015-03-24

fujunwei:

BUG=XWALK-2132


rakuco:
Is this approach better than calling `l10n_util::GetApplicationLocale(std::string())`?

fujunwei:
The `l10n_util::GetApplicationLocale(std::string())` wrapper different platform implementation, it will call `base::android::GetDefaultLocale()` on android platform.

PR #2919, opened 2015-03-24
Fix crash on startup after https://codereview.chromium.org/933893003

darktears:
Commit above changed the way initialisation is done and expect the
application icon to be set before arriving in Init(). This commit fixes
it.


darktears:
@rakuco PTAL

rakuco:
I have no idea about the code, so I trust whatever what you are doing. My only question is whether this is going to work on systems that don't use Views (Mac?)

darktears:
I'll figure out Mac when I build it.

PR #2920, opened 2015-03-24
Make native filesystem extension work on Windows.

darktears:
I believe it needs to be extended and completed.

### DIFF
--- a/experimental/native_file_system/virtual_root_provider_win.cc
+++ b/experimental/native_file_system/virtual_root_provider_win.cc
@@ -4,9 +4,42 @@
 
 #include "xwalk/experimental/native_file_system/virtual_root_provider.h"
 
-#include "base/logging.h"
+#include "base/files/file_path.h"
+#include "base/files/file_util.h"
+
+bool VirtualRootProvider::testing_enabled_ = false;
 
 VirtualRootProvider::VirtualRootProvider() {
-  // TODO(darktears): Windows support to be added.
-  NOTIMPLEMENTED();
+  if (!virtual_root_map_.empty())
+    return;
+
+  if (testing_enabled_) {
+    base::GetTempDir(&home_path_);
+  } else {
+    home_path_ = base::GetHomeDir();
+  }
+
+  virtual_root_map_["DESKTOP"] =
+      home_path_.Append(FILE_PATH_LITERAL("Desktop"));
+  virtual_root_map_["DOWNLOADS"] =
+      home_path_.Append(FILE_PATH_LITERAL("Downloads"));
+  virtual_root_map_["DOCUMENTS"] =
+      home_path_.Append(FILE_PATH_LITERAL("Documents"));
+  virtual_root_map_["MUSIC"] =
+      home_path_.Append(FILE_PATH_LITERAL("Music"));
+  virtual_root_map_["PICTURES"] =
+      home_path_.Append(FILE_PATH_LITERAL("Pictures"));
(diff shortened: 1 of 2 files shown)


darktears:
@shawngao5 PTAL

experimental/native_file_system/virtual_root_provider_win.cc
```diff
@@ -1,12 +1,39 @@
...
+    return;
+
+  if (testing_enabled_) {
+    base::GetTempDir(&home_path_);
+  }
+  else {
+    home_path_ = base::GetHomeDir();
+  }
+
+  virtual_root_map_["DESKTOP"] = home_path_.Append(FILE_PATH_LITERAL("Desktop"));
```
pozdnyakov: "DESKTOP" (and others) are used somewhere else, right? it's worth making const variables for them

darktears: Yep in the linux provider, however this one should be rewritten to use xdg to get these names. Not sure it's worth it at the moment. I think overall the native file system access needs a lot of love.

experimental/native_file_system/virtual_root_provider_win.cc
```diff
@@ -1,12 +1,45 @@
-// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
shawngao5: 2015

experimental/native_file_system/virtual_root_provider_win.cc
```diff
@@ -1,12 +1,45 @@
...
+  virtual_root_map_["DOWNLOADS"] =
+      home_path_.Append(FILE_PATH_LITERAL("Downloads"));
+  virtual_root_map_["DOCUMENTS"] =
+      home_path_.Append(FILE_PATH_LITERAL("Documents"));
+  virtual_root_map_["MUSIC"] =
+      home_path_.Append(FILE_PATH_LITERAL("Music"));
+  virtual_root_map_["PICTURES"] =
+      home_path_.Append(FILE_PATH_LITERAL("Pictures"));
+  virtual_root_map_["VIDEOS"] =
+      home_path_.Append(FILE_PATH_LITERAL("Videos"));
```
shawngao5: In windows such folders are called `Known Folder`, https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457(v=vs.85).aspx.
These folders can be redirected to other paths than just under home directory by API, `SHSetKnownFolderPath`,  https://msdn.microsoft.com/en-us/library/windows/desktop/bb762249(v=vs.85).aspx.
So the better way for Windows is `SHGetKnownFolderPath`, https://msdn.microsoft.com/en-us/library/windows/desktop/bb762188(v=vs.85).aspx.

experimental/native_file_system/virtual_root_provider_win.cc
```diff
@@ -1,12 +1,45 @@
...
-
-#include "xwalk/experimental/native_file_system/virtual_root_provider.h"
-
-#include "base/logging.h"
-
-VirtualRootProvider::VirtualRootProvider() {
-  // TODO(darktears): Windows support to be added.
-  NOTIMPLEMENTED();
-}
+// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
shawngao5: 2015

darktears:
Tracked there instead : https://github.com/crosswalk-project/crosswalk/pull/2936

PR #2921, opened 2015-03-25
[Android] Configure download url of crosswalk library apk

lincsoon:
When crosswalk app running on shared mode, there's a built-in updater to help to download and update crosswalk library apk. By default, the updater will get library apk from Google Play Store.

By this feature, the developer of crosswalk app could self-deploy library apk according to their own requirements, and specify the download url of apk in the app. The updater will fetch the library apk using android download manager.

Setup Method 1: define a meta-data item with the name of "xwalk_apk_url" under application tag in AndroidManifest.xml. For example,

```
<application android:name="org.xwalk.core.XWalkApplication">
    <meta-data android:name="xwalk_apk_url"
               android:value="http://host/XWalkRuntimeLib.apk" />
    ...
</application>
```

Setup Method 2: use --xwalk-apk-url option for make_apk.py. For example,

```
make_apk.py --xwalk-apk-url=http://host/XWalkRuntimeLib.apk ...
```

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -97,6 +97,8 @@ def ParseManifest(options):
   else:
     print('Error: there is no app launch path defined in manifest.json.')
     sys.exit(9)
+  if not options.xwalk_apk_url and parser.GetXWalkApkUrl():
+    options.xwalk_apk_url = parser.GetXWalkApkUrl()
   options.icon_dict = {}
   if parser.GetAppRoot():
     options.app_root = parser.GetAppRoot()
@@ -254,6 +256,8 @@ def Customize(options, app_info, manifest):
     app_info.orientation = options.orientation
   if options.icon:
     app_info.icon = '%s' % os.path.expanduser(options.icon)
+  if options.xwalk_apk_url:
+    app_info.xwalk_apk_url = options.xwalk_apk_url
 
   #Add local extensions to extension list.
   extension_binary_path_list = GetExtensionBinaryPathList()
@@ -280,7 +284,7 @@ def Execution(options, app_info):
   # start to generate suitable versionCode
   app_info.app_versionCode = MakeVersionCode(options, app_info.app_version)
   # Write generated versionCode into AndroidManifest.xml.
-  # Later if we have other customization, 
+  # Later if we have other customization,
   # we can put them together into CustomizeManifest func.
   CustomizeManifest(app_info)
   name = app_info.android_name
@@ -581,6 +585,11 @@ def main(argv):
           '\'app_root\'. This flag should work with \'--app-root\' together. '
           'For example, --app-local-path=/relative/path/of/entry/file')
   group.add_option('--app-local-path', help=info)
+  info = ('The download URL of the Crosswalk runtime library APK. '
+          'The built-in updater uses the Android download manager to fetch '
+          'the url. '
+          'For example, --xwalk-apk-url=http://myhost/XWalkRuntimeLib.apk')
+  group.add_option('--xwalk-apk-url', help=info)
   parser.add_option_group(group)
   # Mandatory options group
   group = optparse.OptionGroup(parser, 'Mandatory arguments',
--- a/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
@@ -20,6 +20,9 @@ import android.content.DialogInterface.OnDismissListener;
 import android.content.DialogInterface.OnShowListener;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.res.Resources;
 import android.database.Cursor;
 import android.net.Uri;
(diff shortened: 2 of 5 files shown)


lincsoon:
@halton @fujunwei @axinging

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -231,22 +232,37 @@ protected void dismissDialog() {
...
     private void getXWalkLibrary() {
-        try {
-            Intent intent = new Intent(Intent.ACTION_VIEW);
-            startActivity(intent.setData(Uri.parse(XWALK_CORE_MARKET_URL)));
-            showDialog(getMarketProgressDialog());
-        } catch (ActivityNotFoundException e) {
-            Log.d(TAG, "Market open failed");
-            showDialog(getMarketOpenFailedDialog());
-            // TODO(sunlin): enable direct download in the future
-            // downloadXWalkLibrary();
```
halton: I'd suggest add comment here to explain how mXWalkApkDownloadUrl get, and when to use the defined url, when to go to Google Play Store.

lincsoon: I illustrated setting methods in commit log, and will document it on website. As you see in commit log, maybe instructions around here are too long to be commented in code?

halton: Commit log is good for this PR review. But think about when other people reading this piece of code, it may modified as well, it is hard to identify which commit log should be read to get the logic.

lincsoon: Done

halton:
LGTM

baleboy:
Does the downloaded APK need to be signed with the Crosswalk store key, or can it be anything? How do we ensure that no malicious crosswalk is loaded by an application installed from the store?

lincsoon:
In the security check implementation, there's an option "verify_xwalk_apk" for building. If this option is enabled, the app will check the signature of package before loading crosswalk library. 

In this case, the apk need to be signed with the crosswalk store key as well. We verify the download apk with the same way of verifying crosswalk from play store, and don't differentiate where the apk came from.

lincsoon:
@rakuco

axinging:
lgtm

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -231,22 +232,40 @@ protected void dismissDialog() {
...
+        // under the application tag in AndroidManifest.xml. It can also be specified via
+        // --xwalk-apk-url option of make_apk script indirectly.
+        if (mXWalkApkDownloadUrl == null) {
+            try {
+                PackageManager packageManager = getPackageManager();
+                ApplicationInfo appInfo = packageManager.getApplicationInfo(
+                        getPackageName(), PackageManager.GET_META_DATA);
+                mXWalkApkDownloadUrl = appInfo.metaData.getString("xwalk_apk_url");
+                if (mXWalkApkDownloadUrl == null) mXWalkApkDownloadUrl = "";
+                Log.d(TAG, "Download Url: " + mXWalkApkDownloadUrl);
```
rakuco: I think it makes sense to provide more context in the log message, like "Crosswalk APK download URL: "

app/tools/android/make_apk.py
```diff
@@ -581,6 +583,10 @@ def main(argv):
           '\'app_root\'. This flag should work with \'--app-root\' together. '
           'For example, --app-local-path=/relative/path/of/entry/file')
   group.add_option('--app-local-path', help=info)
+  info = ('The download url of crosswalk runtime library APK'
+          'Default updater will use android download manager to fetch the url'
+          'For example, --xwalk-apk-url=http://myhost/XWalkRuntimeLib.apk')
+  group.add_option('--xwalk-apk-url', help=info)
```
rakuco: Do we want an equivalent in the manifest as well?

app/tools/android/make_apk.py
```diff
@@ -581,6 +583,10 @@ def main(argv):
           '\'app_root\'. This flag should work with \'--app-root\' together. '
           'For example, --app-local-path=/relative/path/of/entry/file')
   group.add_option('--app-local-path', help=info)
+  info = ('The download url of crosswalk runtime library APK'
+          'Default updater will use android download manager to fetch the url'
```
rakuco: "the android download manager"

app/tools/android/make_apk.py
```diff
@@ -581,6 +583,10 @@ def main(argv):
           '\'app_root\'. This flag should work with \'--app-root\' together. '
           'For example, --app-local-path=/relative/path/of/entry/file')
   group.add_option('--app-local-path', help=info)
+  info = ('The download url of crosswalk runtime library APK'
```
rakuco: Nits: "Crosswalk", not "crosswalk". "Android", not "android". "URL", not "url".

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -231,22 +232,40 @@ protected void dismissDialog() {
...
+            } catch (NameNotFoundException | NullPointerException e) {
```
rakuco: Naive question: which of the calls above could throw `NullPointerException`? Isn't that the kind of exception you _want_ to be thrown since it indicates something actually wrong happened?

baleboy:
@lincsoon ok, thanks for the explanation. @terriko you may want to have a look too.

app/tools/android/make_apk.py
lincsoon: Added comment in the manifest by customize.py:293

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: If there is no meta-data item under application tag, the appInfo.metaData will be null. I rewrote here to make it more clearly.

lincsoon:
Because this patch is based on shared mode implementation, and the PR of shared mode need to be reverted due to a packing problem, please hold the merging of this PR.

rakuco:
Please note you need to rebase your branch, as 4ac738e shouldn't be part of this PR.

lincsoon:
This patch is not ready to merge, still need refine and rebase

lincsoon:
@rakuco I added manifest support. Now this patch is ready.

rakuco:
lgtm, thanks

PR #2925, opened 2015-03-26
[Android] Fix make_apk fail to pack app when API level is lower than 21

lincsoon:
Some attributes in layout files introduced by upstream is added from
API level 21. These resources are also included in xwalk core library.
When building app by make_apk, command "ant release" will fail if API
level is lower than 21.

BUG=XWALK-3219

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -453,8 +453,8 @@ def CheckSystemRequirements():
     print('failed\nThe "android" binary could not be found. Check your Android '
           'SDK installation and your PATH environment variable.')
     sys.exit(1)
-  if GetAndroidApiLevel(android_path) < 14:
-    print('failed\nPlease install Android API level (>=14) first.')
+  if GetAndroidApiLevel(android_path) < 21:
+    print('failed\nPlease install Android API level (>=21) first.')
     sys.exit(3)
 
   # Check ant install


lincsoon:
@rakuco @baleboy @halton

halton:
LGTM, upstream has rolling the required SDK version, it does make sense that we do the same thing. 

But one thing that we do not want to mis-leading users is that we still support from 4.0 (API level 14). Only the building crosswalk requires API level 21.

baleboy:
thanks, LGTM

lincsoon:
Because this patch is based on shared mode implementation, and the PR of shared mode need to be reverted due to a packing problem, please hold the merging of this PR.

rakuco:
Are we good to go now that #2932 has been merged?

lincsoon:
Yes, since no changed in shared mode commit, this PR is good to be merge.

ksarantakos:
This is stopped us from using the AAR pulished in https://download.01.org/crosswalk/releases/crosswalk/android/maven2/org/xwalk/xwalk_core_library/11.40.277.7/

A new release once this is resolved would be super!

thanks!

baleboy:
@ksarantakos this is a fix to our standalone python packaging tool, to give an error message when the installed SDK has API level < 21. I doubt it will help anything with the AAR. What is the exact problem you have?

PR #2929, opened 2015-03-27
[Android] Support disable disk cache

fujunwei:
Use the command line of "--disk-cache-size=1" to disable the disk cache.


rakuco:
Why do you keep closing your pull requests instead of updating them?

PR #2930, opened 2015-03-27
[Android] Limit disk cache size

fujunwei:
Limit the disk cache size with a very small number with an command line:
--disk-cache-size = N
where N is the cache is size limit, by bytes.
Zero size means use the default that is 80m.

BUG=XWALK-3821

### DIFF
--- a/runtime/browser/runtime_url_request_context_getter.cc
+++ b/runtime/browser/runtime_url_request_context_getter.cc
@@ -6,8 +6,10 @@
 #include "xwalk/runtime/browser/runtime_url_request_context_getter.h"
 
 #include <algorithm>
+#include <string>
 #include <vector>
 
+#include "base/command_line.h"
 #include "base/logging.h"
 #include "base/strings/string_number_conversions.h"
 #include "base/strings/string_split.h"
@@ -41,6 +43,7 @@
 #include "xwalk/application/common/constants.h"
 #include "xwalk/runtime/browser/runtime_network_delegate.h"
 #include "xwalk/runtime/common/xwalk_content_client.h"
+#include "xwalk/runtime/common/xwalk_switches.h"
 
 #if defined(OS_ANDROID)
 #include "xwalk/runtime/browser/android/cookie_manager.h"
@@ -53,6 +56,24 @@
 using content::BrowserThread;
 
 namespace xwalk {
+
+int GetDiskCacheSize() {
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  if (!command_line->HasSwitch(switches::kDiskCacheSize))
+      return 0;
+
+  std::string str_value = command_line->GetSwitchValueASCII(
+      switches::kDiskCacheSize);
+
+  int size = 0;
+  if (!base::StringToInt(str_value, &size)) {
+      LOG(ERROR) << "The value " << str_value
+                 << " can not be converted to integer, ignoring!";
+  }
+
+  return size;
+}
 
 RuntimeURLRequestContextGetter::RuntimeURLRequestContextGetter(
     bool ignore_certificate_errors,
@@ -139,7 +160,7 @@ net::URLRequestContext* RuntimeURLRequestContextGetter::GetURLRequestContext() {
             net::DISK_CACHE,
             net::CACHE_BACKEND_DEFAULT,
             cache_path,
-            0,
+            GetDiskCacheSize(),
             BrowserThread::GetMessageLoopProxyForThread(
                 BrowserThread::CACHE));
 
--- a/runtime/common/xwalk_switches.cc
+++ b/runtime/common/xwalk_switches.cc
@@ -13,6 +13,9 @@ const char kAppIcon[] = "app-icon";
 // Disables the usage of Portable Native Client.
 const char kDisablePnacl[] = "disable-pnacl";
 
+// Forces the maximum disk space to be used by the disk cache, in bytes.
+const char kDiskCacheSize[] = "disk-cache-size";
+
 // Enable all the experimental features in XWalk.
 const char kExperimentalFeatures[] = "enable-xwalk-experimental-features";
 
--- a/runtime/common/xwalk_switches.h
+++ b/runtime/common/xwalk_switches.h
@@ -12,6 +12,7 @@ namespace switches {
 
 extern const char kAppIcon[];
 extern const char kDisablePnacl[];
+extern const char kDiskCacheSize[];
 extern const char kExperimentalFeatures[];
 extern const char kListFeaturesFlags[];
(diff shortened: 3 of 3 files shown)


js0701:
Why don't use  "--disk-cache-size=0" instead of   "--disk-cache-size=1"  to disable cache?

fujunwei:
Limit the disk cache size with a very small number with an command line:
--disk-cache-size = N
where N is the cache is size limit, by bytes.
Zero size means use the default that is 80m.

fujunwei:
@halton @lincsoon PTAL, thanks.

fujunwei:
@rakuco PTAL,thanks.

baleboy:
Is this a command line parameter we have introduced, or does it exist in Chrome as well?

baleboy:
It exists in chrome according to http://peter.sh/experiments/chromium-command-line-switches/

baleboy:
@halton @lincsoon could you prioritize this review?

halton:
LGTM

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -82,6 +85,23 @@ RuntimeURLRequestContextGetter::RuntimeURLRequestContextGetter(
 RuntimeURLRequestContextGetter::~RuntimeURLRequestContextGetter() {
 }
 
+inline int GetDiskCacheSize() {
+  int size = 0;
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  if (command_line->HasSwitch(switches::kDiskCacheSize)) {
```
rakuco: It's cleaner if you bail out earlier:

```c++
if (!command_line->HasSwitch(...))
  return 0;
// yadda, yadda
return size;
```

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -139,7 +159,7 @@ net::URLRequestContext* RuntimeURLRequestContextGetter::GetURLRequestContext() {
             net::DISK_CACHE,
             net::CACHE_BACKEND_DEFAULT,
             cache_path,
-            0,
+            GetDiskCacheSize(),
```
rakuco: This means you are repeatedly parsing the command line and trying to convert the same string, over and over again. You should do this once during startup and then use a cached value instead.

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -82,6 +85,23 @@ RuntimeURLRequestContextGetter::RuntimeURLRequestContextGetter(
...
+  int size = 0;
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  if (command_line->HasSwitch(switches::kDiskCacheSize)) {
+    std::string str_value = command_line->GetSwitchValueASCII(
+        switches::kDiskCacheSize);
+
+    if (!base::StringToInt(str_value, &size)) {
+        LOG(ERROR) << "The value " << str_value
+                   << " can not be converted to integer, ignoring!";
```
rakuco: No exclamation marks, it sounds like you are shouting at the user.

runtime/browser/runtime_url_request_context_getter.cc
fujunwei: The URLRequestContextGetter should only be called once per ContentBrowserClient object. Is it necessary to cache it?

runtime/browser/runtime_url_request_context_getter.cc
fujunwei: I will delete the logs.

runtime/browser/runtime_url_request_context_getter.cc
rakuco: I don't like the idea of needlessly re-parsing command line arguments like that; isn't it easy to just do it once and store the value?

baleboy:
can we merge this? @lincsoon @halton

lincsoon:
@fujunwei Could you please respond @rakuco 's review?

runtime/browser/runtime_url_request_context_getter.cc
fujunwei: Sorry for replying later. It's need take time to implement Chromium browser's PrefStore.
The xwalk::GetUserAgent() in this function is also need re-parsing.
How about optimize those after this PR for these command line.

rakuco: I thought it'd be easier to store this value somewhere and pass it to this class. Let's not worry about this part for now then.

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -82,6 +85,23 @@ RuntimeURLRequestContextGetter::RuntimeURLRequestContextGetter(
 RuntimeURLRequestContextGetter::~RuntimeURLRequestContextGetter() {
 }
 
+inline int GetDiskCacheSize() {
```
rakuco: I think you can drop the `inline` part here, and please move this function upper in the file (to an anonymous namespace either inside or outside the `xwalk` one, and before the `RuntimeURLRequestContextGetter` implementations).

fujunwei: Thanks, Done.

baleboy:
@rakuco were your comments addressed?

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -54,6 +57,20 @@ using content::BrowserThread;
...
+  base::CommandLine* command_line = base::CommandLine::ForCurrentProcess();
+
+  if (!command_line->HasSwitch(switches::kDiskCacheSize))
+      return 0;
+
+  std::string str_value = command_line->GetSwitchValueASCII(
+      switches::kDiskCacheSize);
+
+  int size = 0;
+  base::StringToInt(str_value, &size);
```
rakuco: I think it still makes sense to tell the user if the conversion fails.

fujunwei: Ok, done.

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -54,6 +57,24 @@ using content::BrowserThread;
...
+  if (!command_line->HasSwitch(switches::kDiskCacheSize))
+      return 0;
+
+  std::string str_value = command_line->GetSwitchValueASCII(
+      switches::kDiskCacheSize);
+
+  int size = 0;
+  if (!base::StringToInt(str_value, &size)) {
+      LOG(ERROR) << "The value " << str_value
+                 << " can not be converted to integer, ignoring!";
```
rakuco: I said before that exclamation marks should be avoided, but let's address that separately as its already taken so long.

PR #2931, opened 2015-03-27
[Android] Fix the crash issue when launching cordova app.

wuhengzhi:
This patch is to fix the crash issue when launching cordova app.
"initEmbeddedMode()" should be called before get instance otherwise the
instance is always null.

BUG=XWALK-3812


wuhengzhi:
@fujunwei @axinging @lincsoon Please help to review, thanks.

rakuco:
This seems to be superseded by #2932. @lincsoon?

PR #2932, opened 2015-03-27
[Android] Fixed error of checking target architecture

lincsoon:
When checking whether the architecture of Crosswalk matches the device,
load native library firstly and check the flag in libxwalkcore.so
instead of the flag marked in the building phase.


lincsoon:
@halton @fujunwei

lincsoon:
@rakuco This patch is to fix the critical issue I mentioned in the mail that shared mode implementation has impact on packing app for ARM. Please help to review and merge it.

halton:
This is out-of my scope to review. But I trust you because you contribute a lot in shared mode and embedding API. LGTM.

PR #2933, opened 2015-03-27

fujunwei:

BUG=XWALK-2132

### DIFF
--- a/runtime/browser/xwalk_content_browser_client.cc
+++ b/runtime/browser/xwalk_content_browser_client.cc
@@ -51,6 +51,7 @@
 #endif
 
 #if defined(OS_ANDROID)
+#include "base/android/locale_utils.h"
 #include "base/android/path_utils.h"
 #include "base/base_paths_android.h"
 #include "xwalk/runtime/browser/android/xwalk_cookie_access_policy.h"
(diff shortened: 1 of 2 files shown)


rakuco:
Please stop re-sending the same pull request several times, just update the existing one instead. Every time you send the same pull request we lose all context and discussions from the previous versions.

rakuco:
In #2918 you said

> The `l10n_util::GetApplicationLocale(std::string())` wrapper different platform implementation, it will call `base::android::GetDefaultLocale()` on android platform.

When I asked why not use the former instead of the latter. That still doesn't really answer my question -- as far as I can see, it would still have the desired effect on Android while also improving the situation for other platforms.

fujunwei:
The 'l10n_util::GetApplicationLocale(std::string())' will check whether the locale Data exists, but in the phase it can't available by call 'bool IsLocaleAvailable(const std::string& locale)', so `l10n_util::GetApplicationLocale(std::string())` function will return null in this stack call.

fujunwei:
@rakuco @darktears  PTAL.

rakuco:
I see. Does it on the other hand mean we could be loading the locale data earlier? I'd just like to, if possible, make the behavior consistent across platforms (as far as I can see, XWALK-2132 would still be broken outside Android).

fujunwei:
The `10n_util::GetApplicationLocale(std::string())` load resources data from file suffixed with pak (en-us.pak, fr.pak) for chrome browser, it stored localized string such as the title of media error.
`<string name="media_player_error_title">"Cannot play video"</string>`

but Crosswalk extracting those data in xwalk_core_library/res/values-en-rGB/android_content_strings.xml and android_ui_strings.xml. so it's fail to call this function.

halton:
Is this PR dup with #2918? If so, please close one of them.

rakuco:
Understood, thanks for the explanation. So is the change as you wrote it also going to work with the Crosswalk files you mentioned? Or is this already taken care of?

fujunwei:
Yes, it's work finely.

rakuco:
Merged, thanks. Please backport it to the relevant branches.

PR #2934, opened 2015-03-27
Enable WebNotification for Linux defautly.

PeterWangIntel:
To enable WebNotification for Linux defautly. 
Add "pkg-config --cflags gdk-pixbuf-2.0", since some header files need to include files from that patch.

### DIFF
--- a/build/system.gyp
+++ b/build/system.gyp
@@ -265,6 +265,32 @@
           ],
         }
       ],  # targets
+    }, {  # tizen == 0
+      'targets': [
+        {
+          'target_name': 'libnotify',
+          'type': 'none',
+          'conditions': [
+            ['building_crosswalk_bin==1', {
+              'direct_dependent_settings': {
+                'defines': ['USE_LIBNOTIFY=1'],
+                'cflags': [
+                  '<!@(pkg-config --cflags libnotify)',
+                ],
+              },
+              'link_settings': {
+                'ldflags': [
+                  '<!@(pkg-config --libs-only-L --libs-only-other libnotify)',
+                ],
+                'libraries': [
+                  '<!@(pkg-config --libs-only-l libnotify)',
+                ],
+              },
+            },
+           ],
+          ],
+        },
+      ],
     }],
   ],  # conditions
 }
--- a/xwalk.gyp
+++ b/xwalk.gyp
@@ -3,7 +3,6 @@
     'xwalk_product_name': 'XWalk',
     'xwalk_version': '<!(python ../build/util/version.py -f VERSION -t "@MAJOR@.@MINOR@.@BUILD@.@PATCH@")',
     'chrome_version': '<!(python ../build/util/version.py -f ../chrome/VERSION -t "@MAJOR@.@MINOR@.@BUILD@.@PATCH@")',
-    'use_libnotify%': 0,
     'conditions': [
       ['OS=="win" or OS=="mac"', {
         'disable_nacl': 1,
@@ -363,6 +362,15 @@
           # TODO(jschuh): crbug.com/167187 fix size_t to int truncations.
           'msvs_disabled_warnings': [ 4267, ],
         }],  # OS=="win"
+        ['OS=="linux" and tizen!=1', {
+          'dependencies': [
+            'build/system.gyp:libnotify',
+          ],
+          'sources': [
+            'runtime/browser/linux/xwalk_notification_manager.cc',
+            'runtime/browser/linux/xwalk_notification_manager.h',
+          ]
+        }],  # OS=="linux" and tizen!=1
         ['OS=="linux"', {
           'dependencies': [
             '../build/linux/system.gyp:fontconfig',
@@ -400,18 +408,6 @@
             '../ui/aura/aura.gyp:aura',
             '../ui/base/ime/ui_base_ime.gyp:ui_base_ime',
           ],
-        }],
-        ['OS=="linux" and use_libnotify==1', {
-          'defines': ['USE_LIBNOTIFY=1'],
-          'link_settings': {
-            'libraries': [
-              '<!@(pkg-config --libs libnotify)',
-            ],
-          },
-          'sources': [
-            'runtime/browser/linux/xwalk_notification_manager.cc',
-            'runtime/browser/linux/xwalk_notification_manager.h',
-          ]
         }],
         ['disable_nacl==0', {
             'conditions': [


PeterWangIntel:
@rakuco 
weird, It works well in a clean local machine. can you give some tips? thank you very much.

halton:
As your code indicate, you need ask Raphael to install `libgdk-pixbuf-2.0-dev` on the linux and Tizen trybot slaves.

build/system.gyp
```diff
@@ -35,6 +35,25 @@
...
     },
+    {
+      'target_name': 'libnotify',
+      'type': 'none',
+      'conditions': [
+        ['building_crosswalk_bin==1', {
+          'defines': ['USE_LIBNOTIFY=1'],
+          'cflags': [
+            '<!@(pkg-config --cflags libnotify)',
+            '<!@(pkg-config --cflags gdk-pixbuf-2.0)',
```
halton: Why gdk-pixbuf-2.0 is required for the libnotify target, I do not see those two modules has dependency in nature. If your work requires both of them, then create two separate target make more sense to me.

rakuco: Yeah, I think Peter was manually trying to work around the gdk-pixbuf errors the Linux builders were having. Don't, this is likely a packaging error in Ubuntu Precise's `libnotify-dev`. I've manually installed the gdk development packages in the bots so you shouldn't need this.

rakuco:
Note you also need to add a dependency on libnotify in `crosswalk-bin.spec` for it to work in Tizen.

PeterWangIntel:
@rakuco @halton 
is this the expected by trybot? thank you very much.

halton:
LGTM

build/system.gyp
```diff
@@ -35,6 +35,24 @@
        ],
       ],
     },
+    {
+      'target_name': 'libnotify',
+      'type': 'none',
+      'conditions': [
+        ['building_crosswalk_bin==1', {
+          'defines': ['USE_LIBNOTIFY=1'],
```
rakuco: You need to follow what the "gio" target above does more closely: `defines` and `cflags` need to be in an `direct_dependent_settings` block, otherwise they will only be used when building this target itself.

build/system.gyp
```diff
@@ -35,6 +35,24 @@
...
+          'cflags': [
+            '<!@(pkg-config --cflags libnotify)',
+          ],
+          'link_settings': {
+	    'libraries': [
```
rakuco: Same thing about following the "gio" target: splitting `libraries` into `ldflags` and `libraries` is a good idea (plus your indentation here is wrong, you used 1 space instead of 2).

packaging/crosswalk-bin.spec
```diff
@@ -69,6 +69,7 @@ BuildRequires:  pkgconfig(haptic)
 BuildRequires:  pkgconfig(icu-i18n)
 BuildRequires:  pkgconfig(libdrm)
 BuildRequires:  pkgconfig(libexif)
+BuildRequires:  pkgconfig(libnotify)
```
rakuco: Argh! My apologies: I thought Tizen had libnotify packaged, but it turns out their `notification` package is not libnotify, but rather some Tizen-specific notification API :(
I guess this won't work on Tizen at the moment, sigh.

build/system.gyp
PeterWangIntel: thank you.

build/system.gyp
PeterWangIntel: sorry. yes.

packaging/crosswalk-bin.spec
PeterWangIntel: ok

PeterWangIntel:
@rakuco @halton 
Please have a look. Thank you very much.

halton:
You already have my lgtm.

build/system.gyp
```diff
@@ -35,6 +35,26 @@
...
+      'conditions': [
+        ['building_crosswalk_bin==1', {
+          'direct_dependent_settings': {
+            'defines': ['USE_LIBNOTIFY=1'],
+            'cflags': [
+              '<!@(pkg-config --cflags libnotify)',
+            ],
+          },
+          'link_settings': {
+	    'libraries': [
```
rakuco: The indentation is still missing one space, and you still need an `ldflags` section.

rakuco:
Tizen also matches `OS=="linux"`, which is why the build is failing there.

build/system.gyp
PeterWangIntel: sorry. I'll fix it soon.

PeterWangIntel:
> Tizen also matches OS=="linux", which is why the build is failing there.

Yes, I've noticed. 
And, only for building RPM, it turns on the flag "tizen" ( crosswalk-bin.spec : 246) which defines "OS_TIZEN" eventually.
But, seems, our trybot for Tizen will try to build a xwalk without flag "tizen" first. 
So, unless the code works for both Linux and Tizen, seems, it cannot pass the trybot.

rakuco:
I think you are mixing different things here. You are right that the Tizen slaves call `gclient runhooks` from the host (and thus regular Linux) side even though they should not call some of the hooks. However, that is orthogonal to this patch, as `tizen=1` is set _in addition_ to the regular Linux flags. That is, Tizen is also Linux, so if you only check for `OS=="linux"` you will match both regular Linux as well as Tizen. If you only want to target the former, you need something like `OS=="linux" and tizen!=1`.

rakuco:
lgtm, but we need to land #2949 first (or just make it part of this patch).

PeterWangIntel:
@rakuco @halton 
Please help to merge, thank you.

halton:
I'm not owner, I'll wait for @rakuco

rakuco:
Can you push one more time so we can double-check that all bots are passing?

PeterWangIntel:
@rakuco 
I'm so sorry for late response, just noticed your last comment. 
Seems, it still has problem with Tizen trybot.

rakuco:
sigh, why does gyp have to suck so much :/
it's basically processing the libnotify target in system.gyp even if it's never referenced in any gyp file on Tizen. Try moving it to the `else` part of the `tizen==1` check there:

```python
{
  # ...
  'targets': [
    {
      'target_name': 'gio',
      # ...
    },
  ],
  'conditions': [
    ['tizen==1', {
      # ...
    }, {  # tizen == 0
      {
        'target_name': 'libnotify',
        # ...
      },
    }],
  ],
}
```

halton:
Hi @PeterWangIntel , maybe not very related with this PR.

Upstream seems support WhatWG notification API already, does our notification on Linux on top of that. You could get several bugs on upstream, one is https://code.google.com/p/chromium/issues/detail?id=392187

rakuco:
I think we've begun talking about this in JIRA: https://crosswalk-project.org/jira/browse/XWALK-3654

PeterWangIntel:
@rakuco 
Seems, it works for Tizen trybot now.
The failure of unittest "xwalk_sysapps_browsertest" seems not caused by my patch. Without it in my local build, the unittest still works wrong.

rakuco:
yeah, don't worry about it, it is unrelated to your change and we need to investigate the flakiness separately

rakuco:
merged, thanks for the patience with all the back and forth!

PeterWangIntel:
:), @rakuco @halton, thank you for your time.

PR #2935, opened 2015-03-27
crosswalk-7: Backport commits that track Chromium dependencies with git.

rakuco:
This pull request backports several commits related to XWALK-1894 (which was about tracking all Chromium dependencies with git repositories instead of SVN ones). They are present in Crosswalk >= 8, but we may need to do one additional Crosswalk 7 build, and since the 36.0.1985.144 release is not present in SVN we need those commits for the build to work.


rakuco:
@darktears @baleboy

baleboy:
thanks. @darktears the exceptional crosswalk 7 build is needed for XDK users who can't move to a later crosswalk because of performance problems still under investigation.

PR #2938, opened 2015-03-28
Second round of android fixes

mrunalk:
This patch properly fixes the RequestPermission replies with
PermissionStatus instead of bool as per this patch,
https://codereview.chromium.org/955383003

### DIFF
--- a/runtime/browser/runtime_geolocation_permission_context.h
+++ b/runtime/browser/runtime_geolocation_permission_context.h
@@ -26,7 +26,7 @@ class RuntimeGeolocationPermissionContext
   virtual void RequestGeolocationPermission(
       content::WebContents* web_contents,
       const GURL& requesting_frame,
-      base::Callback<void(content::PermissionStatus)> result_callback);
+      base::Callback<void(bool)> result_callback);
   virtual void CancelGeolocationPermissionRequest(
       content::WebContents* web_contents,
       const GURL& requesting_frame);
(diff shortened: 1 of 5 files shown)


runtime/browser/xwalk_content_browser_client.cc
```diff
@@ -82,6 +82,13 @@ namespace {
 // The application-wide singleton of ContentBrowserClient impl.
 XWalkContentBrowserClient* g_browser_client = nullptr;
 
+void CallbackPermisisonStatusWrapper(
```
rakuco: Typo: "Permisison"

runtime/browser/runtime_geolocation_permission_context.h
```diff
@@ -26,7 +26,7 @@ class RuntimeGeolocationPermissionContext
   virtual void RequestGeolocationPermission(
       content::WebContents* web_contents,
       const GURL& requesting_frame,
-      base::Callback<void(content::PermissionStatus)> result_callback);
+      base::Callback<void(bool)> result_callback);
```
darktears: This is not what upstream is moving to. I actually changed bool -> PermissionStatus, it was bool before. My quick look is that we just need to modify the ultimate piece going to Java from bool to enum.

runtime/browser/xwalk_content_browser_client.cc
mrunalk: Looks like that typo is also present in the upstream code :) But I think we can change it in our code and then I can submit a patch to fix it upstream as well.

runtime/browser/runtime_geolocation_permission_context.h
mrunalk: I think it is. If you look at InvokeGeolocationCallback, https://code.google.com/p/chromium/codesearch#chromium/src/android_webview/native/aw_contents.cc&q=aw_conte&sq=package:chromium&l=1
then they haven't changed java equivalent from bool to enum. That's why I had to change it here. This is in fact similar to the upstream change.

darktears:
OK then,

PR #2940, opened 2015-03-30
[Android]Update the string mapping and query logic(Crosswalk 9)

LaneRuan:
According to http://www.w3.org/2012/sysapps/contacts-manager-api .the string mapping should be "givenNames to givenNames".

BUG=XWALK-2451.


rakuco:
Why is this targeting Crosswalk 9?

LaneRuan:
I will test it with master and will pull agian soon.

PR #2862, opened 2015-02-13
[Android] Set the command line to webkit Preference

fujunwei:
The command of "--allow-running-insecure-content" in assets directory do not effect the Webkit behavior because not set it to Webkit preference.

PR #2941, opened 2015-03-30
[Android] Set the command line to webkit Preference

fujunwei:
The command of "--allow-running-insecure-content" in assets directory
do not effect the Webkit behavior because not set it to Webkit
preference.

(cherry picked from commit 19971ff4b322372b622f3432a9ea89c1f4cf2cdd)

### DIFF
--- a/runtime/browser/android/xwalk_settings.cc
+++ b/runtime/browser/android/xwalk_settings.cc
@@ -17,6 +17,7 @@
 #include "content/public/common/web_preferences.h"
 #include "jni/XWalkSettings_jni.h"
 #include "xwalk/runtime/common/xwalk_content_client.h"
+#include "xwalk/runtime/common/xwalk_switches.h"
 #include "xwalk/runtime/browser/android/renderer_host/xwalk_render_view_host_ext.h"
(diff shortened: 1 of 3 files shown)


rakuco:
Neither this nor the original pull request, #2862, mention a related bug report in JIRA, so it is hard to determine if this is actually needed in crosswalk-12.
@baleboy?

fujunwei:
There is a customer ask for this patch, it's block their to release app. I forward the helped email to you.

rakuco:
Please keep the communication in the open whenever possible. In this case you are simply referring to a post sent to a public mailing list by one of our users: https://lists.crosswalk-project.org/pipermail/crosswalk-help/2015-March/000964.html

It would be good if you could point to this thread in the PR and the commit message to explain why this is worth cherry-picking into the stable branch, otherwise there is no context.

@baleboy: Shall we get it into crosswalk-12?

rakuco:
Hmm, Alexis was faster...

fujunwei:
Thanks for reminding, i will follow those rules to cherry pick PR.

PR #2942, opened 2015-03-30
[Linux] Add Linux IME support.

seanlong:
Before this patch we use the default FakeInputMethodContext, IMEs like
Chinese IME is not work. This patch leverages Chromium
X11InputMethodContextImplGtk2 to communicate with underlying IMEs.

Manually tested with Chinese IME on Ubuntu & Deepin.

BUG=XWALK-3823

### DIFF
--- a/runtime/browser/ui/x11_input_method_context_factory.cc
+++ b/runtime/browser/ui/x11_input_method_context_factory.cc
@@ -1,0 +1,108 @@
+// Copyright 2013 The Chromium Authors. All rights reserved.
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/runtime/browser/ui/x11_input_method_context_factory.h"
+
+#include <gdk/gdk.h>
+#include <gdk/gdkx.h>
+#include <gtk/gtk.h>
+#include <X11/X.h>
+
+#include <string>
+#include <vector>
+
+#include "base/command_line.h"
+#include "base/debug/leak_annotations.h"
+#include "ui/gfx/x/x11_types.h"
+#include "xwalk/runtime/browser/ui/x11_input_method_context_impl_gtk2.h"
+
+namespace {
+
+void GtkInitFromCommandLine(const base::CommandLine& command_line) {
+  const std::vector<std::string>& args = command_line.argv();
+  int argc = args.size();
+  scoped_ptr<char *[]> argv(new char *[argc + 1]);
+  for (size_t i = 0; i < args.size(); ++i) {
+    // TODO(piman@google.com): can gtk_init modify argv? Just being safe
+    // here.
+    argv[i] = strdup(args[i].c_str());
+  }
+  argv[argc] = NULL;
+  char **argv_pointer = argv.get();
+
+  {
+    // http://crbug.com/423873
+    ANNOTATE_SCOPED_MEMORY_LEAK;
+    gtk_init(&argc, &argv_pointer);
+  }
+  for (size_t i = 0; i < args.size(); ++i) {
+    free(argv[i]);
+  }
+}
+
+void ProcessGdkEventKey(const GdkEventKey& gdk_event_key) {
+  // This function translates GdkEventKeys into XKeyEvents and puts them to
+  // the X event queue.
+  //
+  // base::MessagePumpX11 is using the X11 event queue and all key events should
+  // be processed there.  However, there are cases(*1) that GdkEventKeys are
+  // created instead of XKeyEvents.  In these cases, we have to translate
+  // GdkEventKeys to XKeyEvents and puts them to the X event queue so our main
+  // event loop can handle those key events.
+  //
+  // (*1) At least ibus-gtk in async mode creates a copy of user's key event and
+  // pushes it back to the GDK event queue.  In this case, there is no
+  // corresponding key event in the X event queue.  So we have to handle this
+  // case.  ibus-gtk is used through gtk-immodule to support IMEs.
+
+  XEvent x_event = {0};
+  x_event.xkey.type =
+      gdk_event_key.type == GDK_KEY_PRESS ? KeyPress : KeyRelease;
+  x_event.xkey.send_event = gdk_event_key.send_event;
+  x_event.xkey.display = gfx::GetXDisplay();
+  x_event.xkey.window = GDK_WINDOW_XID(gdk_event_key.window);
+  x_event.xkey.root = DefaultRootWindow(x_event.xkey.display);
+  x_event.xkey.time = gdk_event_key.time;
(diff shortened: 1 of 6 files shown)


seanlong:
@pozdnyakov, thanks.

seanlong:
ping @pozdnyakov

runtime/browser/ui/x11_input_method_context_factory.cc
```diff
@@ -0,0 +1,108 @@
...
+#include "base/debug/leak_annotations.h"
+#include "ui/gfx/x/x11_types.h"
+#include "xwalk/runtime/browser/ui/x11_input_method_context_impl_gtk2.h"
+
+namespace {
+
+void GtkInitFromCommandLine(const base::CommandLine& command_line) {
+  const std::vector<std::string>& args = command_line.argv();
+  int argc = args.size();
+  scoped_ptr<char *[]> argv(new char *[argc + 1]);
```
pozdnyakov: pls look at ApplicationSystemTizen::LaunchFromCommandLine, think similar code there looks nicer

runtime/browser/ui/x11_input_method_context_factory.cc
```diff
@@ -0,0 +1,108 @@
...
+  gtk_main_do_event(gdk_event);
+}
+
+}  // namespace
+
+namespace xwalk {
+
+X11InputMethodContextFactory::X11InputMethodContextFactory() {
+  GtkInitFromCommandLine(*base::CommandLine::ForCurrentProcess());
+  gdk_event_handler_set(DispatchGdkEvent, NULL, NULL);
```
pozdnyakov: nullptr (pls fix other places too)

pozdnyakov:
so now we're having dependency to GTK, can we do without it?
@rakuco

rakuco:
If possible I'd really like not forking those files from `src/chrome/ui/libgtk2ui`. They are big and it is obvious we are not going to keep them in sync with the upstream versions.
Do you know what upstream's plans are for IME + Aura? Do you intend to engage with the Chromium folks to find a common solution?
@joone: How does this interact with your IME code in ozone-wayland?

seanlong:
Agree on the sync effort. But could we just use this simple way first and adapt to better & comprehensive solution later? As you know this is a big missing when we want to integrate Crosswalk to China Linux distributors.

rakuco:
I understand your point, but on the other hand you haven't answered any of the questions. And this is not the first time that the "this is critical for $THING" argument has been used to justify doing things as fast as possible while never fixing the technical debt. I am all for spreading Crosswalk as widely as possible, but I would also like to have all these matters discussed so we can all agree on the path forward and not shoot ourselves on the foot in the future.

seanlong:
> If possible I'd really like not forking those files from src/chrome/ui/libgtk2ui. They are big and it is obvious we are not going to keep them in sync with the upstream versions.

Is directly include src/chrome/ui/libgtk2ui to xwalk gyp an option?

> Do you know what upstream's plans are for IME + Aura? Do you intend to engage with the Chromium folks to find a common solution?

I searched around, not found any particular discussion about replacing current GTK based IME impl. https://code.google.com/p/chromium/issues/detail?id=133476 is the original upstream issue for GTK IME impl, it emphasized "**gtk-immodule is not necessarily GUI framework, and it manages a tons of IMEs gracefully**". So personally I don't know what's the good to re-implement these wheels.

xwalk.gyp
```diff
@@ -372,6 +372,17 @@
             'experimental/native_file_system/virtual_root_provider_linux.cc',
           ]
         }],  # OS=="linux"
+        ['OS=="linux"' and 'tizen==0', {
```
rakuco: This is supposed to be one string (`"OS=='linux' and tizen==0"`). And I think there already exists a block with this very check somewhere else in this file.

rakuco:
> Is directly include src/chrome/ui/libgtk2ui to xwalk gyp an option?

In the worst case, yes. So far we've managed to avoid depending on anything above Chromium's content layer, especially since code in `chrome/` is usually related to the browser itself. Still, I'd say it's better to do that than copy the files if they don't have any additional build dependencies on other parts of `chrome/`. @darktears?

> I searched around, not found any particular discussion about replacing current GTK based IME impl. https://code.google.com/p/chromium/issues/detail?id=133476 is the original upstream issue for GTK IME impl, it emphasized "gtk-immodule is not necessarily GUI framework, and it manages a tons of IMEs gracefully". So personally I don't know what's the good to re-implement these wheels.

I agree that reinventing the wheel does not make sense here. There are still some points I would like to address though:

* Have you looked at @joone's [work on IME for ozone-wayland](https://github.com/01org/ozone-wayland/tree/master/ui/ime)? Is there anything there that could be of use here?
* If upstream is not considering moving away from gtk-immodule in the foreseeable future I still expect the people doing the integration work in Crosswalk to engage with the Chromium people -- for example, by trying to move this code away from `chrome/`.
* Introducing a new dependency on something like GTK+ may be problematic especially for people targeting smaller platforms (think Crosswalk on Yocto, for example) where IME is not important. I suggest adding a new build system variable to `build/common.gypi` (something like `enable_x11_ime` defaulting to 1) so people can opt out of this if they want.

seanlong:
> Have you looked at @joone's work on IME for ozone-wayland? Is there anything there that could be of use here?

I not familiar with this area, from my point view, the GTK impl is a glue between X and gtk-immodule:
X <-> **chromium_x11_ime_context_gtk** <-> gtk-immodule <-> IMF(idus, scim, fcitx...) <-> IME(pinyin...), the complicate part is gtk-immodule handling different IMFs. How to leverage ozone-wayland code to handle this?

> If upstream is not considering moving away from gtk-immodule in the foreseeable future I still expect the people doing the integration work in Crosswalk to engage with the Chromium people -- for example, by trying to move this code away from chrome/.

I can try to ping upstream guys about this.

> Introducing a new dependency on something like GTK+ may be problematic especially for people targeting smaller platforms (think Crosswalk on Yocto, for example) where IME is not important. I suggest adding a new build system variable to build/common.gypi (something like enable_x11_ime defaulting to 1) so people can opt out of this if they want.

Make sense. So I will firstly try to include these files into xwalk gyp, if you guys agree. Thanks.

joone:
Support for IME has not fully implemented and tested in ozone-wayland. It only tested with weekeyboard that only supports Latin characters.

We may be able to implement an IME extension using Chrome Input APIs for Crosswalk like Chrome browser:

https://chrome.google.com/webstore/detail/google-input-tools/mclkkofklkfljcocdinagocijmpgbhab?utm_source=chrome-ntp-icon  

https://www.chromium.org/developers/design-documents/extensions/proposed-changes/apis-under-development/input-method-editor

In this case, we need to implement the similar Input APIs in Crosswalk. I think this looks like wrappers of Aura Input APIs

seanlong:
Seems IME is ready supported by https://github.com/crosswalk-project/crosswalk/pull/2988, so close it.

PR #2943, opened 2015-03-30
[Extensions] Enable cross-context reference for lazy loading

DonnaWuDongxia:
By default, crosswalk extensions are lazy loaded when they
are actually accessed through Trampoline callbacks. But cross
context reference(JS reference to the leaf extension objects
iframes' contentWindow or popup window by window.open) will
mess up the contexts in the callback which will cause runtime
crash. This commit enable the cross-context reference for
trampoline callback.

BUG=https://crosswalk-project.org/jira/browse/XWALK-3575

### DIFF
--- a/extensions/renderer/xwalk_module_system.cc
+++ b/extensions/renderer/xwalk_module_system.cc
@@ -357,6 +357,7 @@ void XWalkModuleSystem::DeleteExtensionModules() {
 void XWalkModuleSystem::LoadExtensionForTrampoline(
     v8::Isolate* isolate,
     v8::Local<v8::Value> data) {
+  v8::HandleScope handle_scope(isolate);
   v8::Local<v8::Array> params = data.As<v8::Array>();
   void* ptr = params->Get(
       v8::Integer::New(isolate, 0)).As<v8::External>()->Value();
@@ -365,8 +366,10 @@ void XWalkModuleSystem::LoadExtensionForTrampoline(
 
   if (!entry)
     return;
+  v8::Handle<v8::Context> context = params->CreationContext();
 
-  v8::Handle<v8::Context> context = isolate->GetCurrentContext();
+  // Switch to the target context in case cross-context reference.
+  v8::Context::Scope context_scope(context);
 
   DeleteAccessorForEntryPoint(context, entry->name);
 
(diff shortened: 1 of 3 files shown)


DonnaWuDongxia:
@vcgomes @halton Please help to review, Thanks!

halton:
miss up -> mess up

extensions/renderer/xwalk_module_system.cc
```diff
@@ -357,6 +357,7 @@ void XWalkModuleSystem::DeleteExtensionModules() {
 void XWalkModuleSystem::LoadExtensionForTrampoline(
     v8::Isolate* isolate,
     v8::Local<v8::Value> data) {
+  v8::HandleScope handle_scope(isolate);
```
halton: This variable is not used, is it?

pozdnyakov: this probably a scope guard to make sure 'isolate' is not deleted

extensions/renderer/xwalk_module_system.cc
```diff
@@ -365,8 +366,11 @@ void XWalkModuleSystem::LoadExtensionForTrampoline(
 
   if (!entry)
     return;
+  v8::Handle<v8::Context> context = params->CreationContext();
+
+  // Switch to the target context in case cross-context reference.
+  v8::Context::Scope context_scope(context);
```
halton: nit, leave one empty line

extensions/renderer/xwalk_module_system.cc
DonnaWuDongxia: It create a sub HandleScope here instead of using the default scope, please refer the comments , https://code.google.com/p/chromium/codesearch#chromium/src/v8/src/handles.h&q=HandleScope&sq=package:chromium&type=cs&l=170

halton: Thank for explanation.

halton:
LGTM. ping @rakuco

rakuco:
I have no idea about this code :-) I only wonder if there's a way to test it -- the extensions subsystem is pretty well unit-tested, so adding a test for this problem would be excellent if possible.

rakuco:
It's possible that your new test is timing out because of the reasons I've outlined in #2897. I suggest removing the `document.write()` call (or just making it some plain text in `<body></body>`) and calling `iframe.contentDocument.close()` after writing to it.

DonnaWuDongxia:
@rakuco Thanks for your advices.

rakuco:
Don't forget to backport this to crosswalk-12 and crosswalk-13 (after at least one round of QA in master).

PR #2944, opened 2015-03-30
[Android] Security check for shared mode

lincsoon:
Before the application loading the Crosswalk runtime library, it is
necessary to make sure the library's package is not tampered and is
exactly the one we published.

The verification process is:
1. Get the binary certificates of the Crosswalk runtime library
   via package manager
2. Generate the hash value with SHA256 algorithm from each certificate
3. Check against the preset hash code generated with the the public key
   of the signing key

This verification is disabled by default. Use option
"./xwalk/gyp -Dverify_xwalk_apk=1" to enable it.

BUG=XWALK-3804

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
@@ -5,8 +5,14 @@
 package org.xwalk.core;
 
 import android.content.Context;
+import android.content.pm.PackageInfo;
+import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
+import android.content.pm.Signature;
 import android.util.Log;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 
 import junit.framework.Assert;
 
@@ -132,12 +138,12 @@ class XWalkCoreWrapper implements ReflectExceptionHandler {
     private boolean findEmbeddedCore() {
         mBridgeContext = null;
         mBridgeLoader = XWalkCoreWrapper.class.getClassLoader();
-        if (!checkCoreVersion() || !checkCoreArchitecture()) {
+        if (!checkCoreArchitecture()) {
             mBridgeLoader = null;
             return false;
         }
 
-        Log.d(TAG, "Running on embedded mode");
+        Log.d(TAG, "Running in embedded mode");
         return true;
     }
 
@@ -145,11 +151,31 @@ class XWalkCoreWrapper implements ReflectExceptionHandler {
         XWalkApplication application = XWalkApplication.getApplication();
         if (application == null) Assert.fail("Must use or extend XWalkApplication");
 
+        if (!XWalkSdkVersion.VERIFY_XWALK_APK) {
+            Log.d(TAG, "Not verifying the package integrity of Crosswalk runtime library");
+        } else {
+            try {
+                PackageInfo packageInfo = application.getPackageManager().getPackageInfo(
+                        XWALK_APK_PACKAGE, PackageManager.GET_SIGNATURES);
+                if (!verifyPackageInfo(packageInfo,
+                        XWalkSdkVersion.XWALK_APK_HASH_ALGORITHM,
+                        XWalkSdkVersion.XWALK_APK_HASH_CODE)) {
+                    mCoreStatus = LibraryStatus.SIGNATURE_CHECK_ERROR;
+                    return false;
+                }
+            } catch (NameNotFoundException e) {
+                Log.d(TAG, "Crosswalk package not found");
+                mCoreStatus = LibraryStatus.NOT_FOUND;
+                return false;
+            }
+        }
+
         try {
             mBridgeContext = application.createPackageContext(XWALK_APK_PACKAGE,
                     Context.CONTEXT_INCLUDE_CODE | Context.CONTEXT_IGNORE_SECURITY);
         } catch (NameNotFoundException e) {
-            Log.d(TAG, "XWalk apk not found");
+            Log.d(TAG, "Crosswalk package not found");
+            mCoreStatus = LibraryStatus.NOT_FOUND;
             return false;
         }
 
@@ -160,8 +186,8 @@ class XWalkCoreWrapper implements ReflectExceptionHandler {
             return false;
         }
 
-        Log.d(TAG, "Running on shared mode");
         application.addResource(mBridgeContext.getResources());
+        Log.d(TAG, "Running in shared mode");
         return true;
     }
 
@@ -182,10 +208,12 @@ class XWalkCoreWrapper implements ReflectExceptionHandler {
                 mCoreStatus = LibraryStatus.MATCHED;
             }
         } catch (RuntimeException e) {
+            Log.d(TAG, "Failed to check library version");
             mCoreStatus = LibraryStatus.NOT_FOUND;
+            return false;
         }
 
-        return mCoreStatus != LibraryStatus.NOT_FOUND;
+        return true;
     }
 
     private boolean checkCoreArchitecture() {
@@ -194,6 +222,7 @@ class XWalkCoreWrapper implements ReflectExceptionHandler {
             ReflectMethod method =
                     new ReflectMethod(null, clazz, "loadXWalkLibrary", Context.class);
             method.invoke(mBridgeContext);
+            mCoreStatus = LibraryStatus.MATCHED;
         } catch (RuntimeException e) {
             Log.d(TAG, "Failed to load native library");
             mCoreStatus = LibraryStatus.NOT_FOUND;
@@ -201,6 +230,54 @@ class XWalkCoreWrapper implements ReflectExceptionHandler {
         }
 
         return true;
+    }
+
+    private boolean verifyPackageInfo(PackageInfo packageInfo,
+            String hashAlgorithm, String hashCode) {
+        if (packageInfo.signatures == null) {
+            Log.e(TAG, "No signature in package info");
+            return false;
+        }
+
+        MessageDigest md = null;
+        try {
+            md = MessageDigest.getInstance(hashAlgorithm);
+        } catch (NoSuchAlgorithmException | NullPointerException e) {
+            Assert.fail("Hash algorithm " + hashAlgorithm + " is not available");
+        }
+
+        byte[] hashArray = hexStringToByteArray(hashCode);
+        if (hashArray == null) {
+            Assert.fail("Hash code is invalid");
+        }
+
+        for (int i = 0; i < packageInfo.signatures.length; ++i) {
+            Log.d(TAG, "Checking signature " + i);
+            byte[] binaryCert = packageInfo.signatures[i].toByteArray();
+            byte[] digest = md.digest(binaryCert);
+            if (!MessageDigest.isEqual(digest, hashArray)) {
+                Log.e(TAG, "Hash code does not match");
+                continue;
+            }
+
+            Log.d(TAG, "Signature passed verification");
+            return true;
+        }
+
+        return false;
+    }
+
+    private byte[] hexStringToByteArray(String str) {
+        if (str == null || str.isEmpty() || str.length()%2 != 0) return null;
+
+        byte[] result = new byte[str.length() / 2];
+        for (int i = 0; i < str.length(); i += 2) {
+            int digit = Character.digit(str.charAt(i), 16);
+            digit <<= 4;
+            digit += Character.digit(str.charAt(i+1), 16);
+            result[i/2] = (byte) digit;
+        }
+        return result;
     }
 
     public boolean isSharedMode() {
--- a/runtime/android/core/src/org/xwalk/core/XWalkLibraryListener.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkLibraryListener.java
@@ -8,6 +8,7 @@ interface XWalkLibraryListener {
     public enum LibraryStatus {
         MATCHED,
         NOT_FOUND,
+        SIGNATURE_CHECK_ERROR,
         NEWER_VERSION,
         OLDER_VERSION
     }
--- a/runtime/android/core/strings/xwalk_app_strings.grd
+++ b/runtime/android/core/strings/xwalk_app_strings.grd
@@ -29,6 +29,12 @@
       <message desc="Startup Not Found Message [CHAR-LIMIT=100]" name="IDS_STARTUP_NOT_FOUND_MESSAGE">
         Please get Crosswalk Runtime Library
       </message>
+      <message desc="Startup Signature Check Error Title [CHAR-LIMIT=100]" name="IDS_STARTUP_SIGNATURE_CHECK_ERROR_TITLE">
+        Crosswalk Signature Verification Failed
+      </message>
+      <message desc="Startup Signature Check Error Message [CHAR-LIMIT=100]" name="IDS_STARTUP_SIGNATURE_CHECK_ERROR_MESSAGE">
+        The integrity of Crosswalk runtime library can not be verified. Please uninstall the runtime library and restart this application
+      </message>
       <message desc="Startup Older Version Title [CHAR-LIMIT=100]" name="IDS_STARTUP_OLDER_VERSION_TITLE">
         Crosswalk Version Too Old
       </message>
--- a/runtime/android/templates/XWalkSdkVersion.template
+++ b/runtime/android/templates/XWalkSdkVersion.template
@@ -2,4 +2,10 @@ package org.xwalk.core;
 
 class XWalkSdkVersion {
     public static final int SDK_VERSION = ${SDK_VERSION};
+    public static final boolean VERIFY_XWALK_APK = ${VERIFY_XWALK_APK};
+    // This is the hash of the "xwalk_cert" signing key and the algorithm with which the hash is generated
+    public static final String XWALK_APK_HASH_ALGORITHM = "SHA-256";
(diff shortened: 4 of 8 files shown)


lincsoon:
@rakuco @baleboy @terri @halton

tools/reflection_generator/reflection_generator.py
```diff
@@ -127,6 +129,8 @@ def main(argv):
   option_parser.add_option('--target-arch', help='Target Architecture')
   option_parser.add_option('--sdk-version', help='API Version')
   option_parser.add_option('--min-sdk-version', help='Min API Version')
+  option_parser.add_option('--verify-xwalk-apk',
```
rakuco: I think it makes sense to pass `type="int"` here to make sure you get consistent values.

runtime/android/templates/XWalkSdkVersion.template
```diff
@@ -2,4 +2,7 @@ package org.xwalk.core;
 
 class XWalkSdkVersion {
     public static final int SDK_VERSION = ${SDK_VERSION};
+    public static final boolean VERIFY_XWALK_APK = ${VERIFY_XWALK_APK};
+    // This is finterprint of android signing key "Crosswalk Project for Android" in codesign.intel.com
```
rakuco: Typo: "finterprint" instead of "fingerprint". You're missing some "the"s too, and codesign.intel.com is an internal domain so I'd rather not mention it in the source code like that.
Suggestion: _"This is the fingerprint of the "Crosswalk Project for Android" signing key."_

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -145,6 +152,11 @@ private boolean findSharedCore() {
         XWalkApplication application = XWalkApplication.getApplication();
         if (application == null) Assert.fail("Must use or extend XWalkApplication");
 
+        if (XWalkSdkVersion.VERIFY_XWALK_APK &&
+                !verifyPackageSignature(XWALK_APK_PACKAGE, XWalkSdkVersion.XWALK_APK_FINGERPRINT)) {
+            return false;
```
rakuco: Shouldn't you complain very loudly if the signature check fails?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -203,6 +215,48 @@ private boolean checkCoreArchitecture() {
...
+            Log.d(TAG, "No signature in package");
+            return false;
+        }
+
+        for (int i = 0; i < signatures.length; ++i) {
+            Log.d(TAG, "Checking signature " + i);
+            byte[] signature = signatures[i].toByteArray();
+
+            try {
+                MessageDigest md = MessageDigest.getInstance("SHA1");
```
rakuco: Does the Play Store mandate SHA1 or could we use a stronger hashing algorithm?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -203,6 +215,48 @@ private boolean checkCoreArchitecture() {
...
+                StringBuffer buffer = new StringBuffer();
+                for (byte code : md.digest(signature)) {
+                    buffer.append(String.format("%02X", code));
+                }
+                String result = buffer.toString();
+                if (!result.equals(fingerprint)) {
+                    Log.d(TAG, "Fingerprint not match, " + result);
+                    continue;
+                }
+            } catch (NoSuchAlgorithmException e) {
```
rakuco: Is it possible that Java has no support for SHA1 these days? If that's the case shouldn't you at least add a log message to inform the user why the check failed?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -203,6 +215,48 @@ private boolean checkCoreArchitecture() {
...
+
+            try {
+                MessageDigest md = MessageDigest.getInstance("SHA1");
+                StringBuffer buffer = new StringBuffer();
+                for (byte code : md.digest(signature)) {
+                    buffer.append(String.format("%02X", code));
+                }
+                String result = buffer.toString();
+                if (!result.equals(fingerprint)) {
+                    Log.d(TAG, "Fingerprint not match, " + result);
```
rakuco: "does not"

gyp_xwalk
```diff
@@ -354,6 +354,18 @@ if __name__ == '__main__':
     args.append('-Dnotifications=1')
     args.append('-Drelease_unwind_tables=0')
 
+  parameter = '-Dverify_xwalk_apk='
```
rakuco: To me it makes more sense to set (and document) this in `build/common.gypi` instead like a regular variable.

rakuco:
Ping @terriko. I'm not sure if the actual code to check the fingerprint is OK (e.g. if the comparisons there are constant-time or if it does not matter in this context, or whether we should test all signatures instead of breaking when finding a match etc).

rakuco:
> from codesign.intel.com. The foundation of this approach is that the

As I mentioned in an inline comment, I don't see much point in mentioning this website in the commit message as it is an internal domain.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
terriko: Agreed: if the signature check fails, you need to do something big and obvious so that the user knows they have an invalid crosswalk apk installed and that's why things aren't running.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
terriko: No, we can use sha256.  You probably want to check the edss config (the stuff on codesign.intel.com), because Matt tells me that the default hashing algorithm in there is currently SHA1 (it used to be 256 but was reverted due to some compatibility) and you'll need to make sure it's set explicitly to sha256 (and not set to sha1 or left blank).

terriko: And to make it more confusing, it does seem that the rest of things *are* sha256, it's just the hashing algorithm that defaults to sha1

gyp_xwalk
lincsoon: My purpose is to make verify_xwalk_apk to be 1 on release mode and to be 0 on debug mode automatically. I removed the -Dverify_xwalk_apk option and tried to insert following code into xwalk.gyp but still got the error message "Undefined variable vefify_xwalk_apk". Could you give any hints about this?

```
'configurations' : {
    'Debug': {
        'variables': {
             'verify_xwalk_apk': 0,
        },
    },
    'Release': {
        'variables': {
             'verify_xwalk_apk': 1,
        },
    },
}

**Comment by lincsoon** [comment]
I confirmed that the fingerprint of Android signing key in codesign.intel.com is generated by SHA1. If I generate the digest of the signature of Crosswalk package by SHA256, the fingerprint doesn't match.

File: gyp_xwalk (hunk 7 above)
Comment (rakuco) [inline gyp_xwalk]: You'd still need to declare a default value first I think:

```python
'variables': {
  # ...
  'verify_xwalk_apk%: 0,
},
'configurations': {
  # ...
  # Likely use 'verify_xwalk_apk%' instead of 'verify_xwalk_apk'
},
```

Why do you think it makes sense to distinguish release and debug in this case?

lincsoon: If verify_xwalk_apk is enabled, the XWalkRuntimeLib.apk must be signed with the official Android signing key through SignFile tool. But to use SignFile, we need signer permission, certificate file, passphrase and internal network, what won't be applicable to some environment like regular development and testing. So I want this flag to be false by default and to be true only in the release package. Actually, I'm not sure what might be more appropriate around here. I'm very glad if you have any advice!

rakuco: I understand and agree with what you said, but I don't see how this is related to Debug vs Release. Our bots and pretty much anyone builds in release mode, so if we enabled `verify_xwalk_apk` by default based on build type Crosswalk would break in most cases. I think you just need to make this opt-in, and then we enable it only when building official beta/stable releases.

terriko:
Yes, it is expected that the SHA256 version not match.  You should regenerate and use the new signature, and replace all the code using the sha1 signature so that it is updated to use the better one.

gyp_xwalk
lincsoon: I agree that this should not be related to Debug vs Release. As you said that "set this in build/common.gypi instead like a regular variable", do you mean the users have to modify the parameter in common.gypi if they want to enable APK verification?

rakuco: Nope; just like with any other variable defined with "%" at the end, a user can change the value of `verify_xwalk_apk` by passing -D to the gyp invocation (for example, `./xwalk/gyp_xwalk -Dverify_xwalk_apk=1`).

lincsoon: Thanks for the explanation! I defined it in xwalk.gyp

lincsoon:
If we agree with the approach of checking fingerprint and the related code, I think we could merge this patch first. If it's necessary to switch from SHA1 to SHA256, I can submit another patch to fix it once the new signing key is ready.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -145,11 +152,30 @@ private boolean findSharedCore() {
         XWalkApplication application = XWalkApplication.getApplication();
         if (application == null) Assert.fail("Must use or extend XWalkApplication");
 
+        if (!XWalkSdkVersion.VERIFY_XWALK_APK) {
+            Log.d(TAG, "Do not verify the xwalk package");
```
rakuco: The message here can be improved. How about "Not verifying Crosswalk runtime package integrity."?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -160,8 +186,8 @@ private boolean findSharedCore() {
             return false;
         }
 
-        Log.d(TAG, "Running on shared mode");
         application.addResource(mBridgeContext.getResources());
+        Log.d(TAG, "Running on shared mode");
```
rakuco: Since you're moving this line anyway, please use "in shared mode" instead of "on shared mode".

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -203,6 +232,39 @@ private boolean checkCoreArchitecture() {
         return true;
     }
 
+    private boolean verifyPackageInfo(PackageInfo packageInfo,
+            String algorithm, String fingerprint) {
+        if (packageInfo.signatures == null) {
+            Log.d(TAG, "No signature in package info");
```
rakuco: Shouldn't this be a fatal error with a log error higher than debug?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -203,6 +232,39 @@ private boolean checkCoreArchitecture() {
...
+                StringBuffer buffer = new StringBuffer();
+                for (byte code : md.digest(signature)) {
+                    buffer.append(String.format("%02X", code));
+                }
+                if (!buffer.toString().equals(fingerprint)) {
+                    Log.d(TAG, "Fingerprint does not match");
+                    continue;
+                }
+            } catch (NoSuchAlgorithmException e) {
+                Log.d(TAG, "Algorithm " + algorithm + " is not available");
```
rakuco: Same here.

runtime/android/core/src/org/xwalk/core/XWalkLibraryListener.java
```diff
@@ -8,6 +8,7 @@
     public enum LibraryStatus {
         MATCHED,
         NOT_FOUND,
+        VERIFY_ERROR,
```
rakuco: `SIGNATURE_CHECK_ERROR` sounds more descriptive. Using a verb like "verify" in an enum is not usual and does not sound good.

runtime/android/core/strings/xwalk_app_strings.grd
```diff
@@ -29,6 +29,12 @@
       <message desc="Startup Not Found Message [CHAR-LIMIT=100]" name="IDS_STARTUP_NOT_FOUND_MESSAGE">
         Please get Crosswalk Runtime Library
       </message>
+      <message desc="Startup Verify Error Title [CHAR-LIMIT=100]" name="IDS_STARTUP_VERIFY_ERROR_TITLE">
```
rakuco: `IDS_STARTUP_SIGNATURE_CHECK_ERROR_TITLE`?

runtime/android/core/strings/xwalk_app_strings.grd
```diff
@@ -29,6 +29,12 @@
...
+        Crosswalk Signature Verification Failed
+      </message>
+      <message desc="Startup Verify Error Message [CHAR-LIMIT=100]" name="IDS_STARTUP_VERIFY_ERROR_MESSAGE">
+        Please uninstall current Crosswalk Runtime Library
```
rakuco: That doesn't sound like an appropriate error message. Aren't you supposed to inform the user that the integrity of the runtime package could not be verified and thus you're aborting, and then perhaps provide some instructions on how to fix the situation?

xwalk.gyp
```diff
@@ -10,6 +10,9 @@
       }],
       ['OS=="android"', {
         'enable_extensions': 1,
+        'sdk_version': '<!(python ../build/util/version.py -f SDK_VERSION -t "@SDK@")',
+        'min_sdk_version': '<!(python ../build/util/version.py -f SDK_VERSION -t "@MIN_SDK@")',
+        'verify_xwalk_apk%': 0,
```
rakuco: It would be good if you could add a short comment to document this variable. See how Chromium's `build/common.gypi` has much more information than ours.

rakuco:
Overall the code looks fine to me, though I'd like @terriko to sign off the algorithm check part before landing.

As for the SHA1 vs SHA256 thing: I'm fine with keeping SHA1 for now, but we must change the algorithm before shipping the support stuff in beta or stable.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
lincsoon: Because the developer can do anything with the APK, this is just a common situation we have considered, but not a error what we can not handle here. What we need to do is letting the user know the integrity of the runtime library can not be verified, but not raising a fatal exception or something like that.

rakuco: I had understood that if we build with the integrity check enabled, a failure in checking the runtime's signature would have to be fatal and we'd need to be as vocal as we can, since it would mean the user has a tampered runtime APK installed.

lincsoon: As you said "we should be as vocal as we can", we already prepared a dialog in `XWalkActivity` for this kind of situation. Generally, I treat the failures in checking the signature as a whole, it means that just check the result of `verifyPackageInfo()` but not deal with the specific error.

The verification failure just means the runtime library is not officially released, not necessarily it's malicious. We have done our job to inform the user that the integrity cannot be verified and also prompt to uninstall it and get the correct one. What to do next is up to the user. If we raise a assertion error or something else during the verification, the subsequent user interaction would be breaked.

rakuco: My bad for saying "fatal error", I think we're saying the same thing differently. My point here is that failures in this method should be reported with `Log.e` instead of `Log.d` since they are more serious than regular informative debugging messages.

The only thing I didn't get is why a verification failure would just mean the runtime hasn't been released; the code is not likely to change once it is, and both now and then if the verification fails it should mean something very fishy is going on and the runtime should not be trusted, no?

lincsoon: Sorry, my misunderstand :-) I'll change the log level to `Log.e`.

We do not just publish the APK of Crosswalk runtime library on Google Play Store, but also on our official website, because a lot of devices are isolated from Google service. The developer could self-deploy their own runtime library and specify the download URL, so the user could get the specific APK file. For this case, the next thing we plan to do is to provide the feasibility that the developers could customize the package name of the runtime library, as well as the hashing algorithm and fingerprint (as you see, I extracted these two variables from `verifyPackageInfo()` as parameter) in order to avoid conflict with official package.

Let's assume that two runtime library from different developers happen to have the same package name, but apparently they do not likely have the same signature. So if the signature verification failed, it just means the runtime library installed on device is not expected to be used by current application, but it's not malicious.

Anyway, I have to admit that it is a mess, and we should document this clearly on website to avoid this kind of situation once this feature is implemented. The error message should also be refined in the future.

lincsoon:
@terriko is there any problem with the fingerprint checking part? If it's ok, we will merge this patch first.

lincsoon:
@rakuco Does the update_scripts error matter? Other tests are all okay.

rakuco:
> @rakuco Does the update_scripts error matter? Other tests are all okay.

No, it doesn't, it looks unrelated to the PR.

poussa:
@terriko we need your blessing of the code before we can merge. Can you check and judge, please.

lincsoon:
Updated `verifyPackageInfo()` using SHA-256 instead of SHA-1

baleboy:
Confirmed off-line that this is OK, the security team is unable to comment for the time being. LGTM by proxy.

PR #2945, opened 2015-03-30
Cleanup an API hook not implemented in the backend.

darktears:


darktears:
@shawngao5 PTAL

lincsoon:
How is this issue now? Is it still available?

rakuco:
This was re-done independently in #3645

PR #2946, opened 2015-03-31
Final round of Android build fixes

mrunalk:
1. Fix the missing include for mojo
2. Get rid of OnJNIOnload and OnLoadDelegate as per,
https://codereview.chromium.org/935413004

### DIFF
--- a/xwalk_android.gypi
+++ b/xwalk_android.gypi
@@ -14,6 +14,7 @@
         '../components/components.gyp:visitedlink_renderer',
         '../components/components.gyp:web_contents_delegate_android',
         '../skia/skia.gyp:skia',
+        '../third_party/mojo/mojo_public.gyp:mojo_cpp_bindings',
         'xwalk_core_extensions_native_jni',
         'xwalk_core_jar_jni',
         'xwalk_core_native_jni',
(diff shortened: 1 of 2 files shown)


xwalk_android.gypi
```diff
@@ -22,6 +22,7 @@
       ],
       'include_dirs': [
         '..',
+        '../third_party/mojo/src',
```
rakuco: We usually don't need to change `include_dirs` like this, since most of the time the targets we depend on in `dependencies` already add whatever directories are required for them to work. Don't we need to depend on some mojo target instead?

mrunalk: I have updated the patch to include the mojo dependency. Although I'm not sure if mojo_environment_chromium is the right target.

rakuco: It depends on the build error you are getting when you don't add this, that is, take a look at which header(s) it complains about and try to figure out which target it comes from. Though to be honest I'd expect us to be using the mojo headers directly for this to fail like you describe, otherwise it's another target we depend on that is not correctly depending on mojo.

mrunalk:
@rakuco I have updated the patch again to include only '../third_party/mojo/mojo_public.gyp:mojo_cpp_bindings' target instead of bigger 'mojo/mojo_base.gyp:mojo_environment_chromium'

PR #2928, opened 2015-03-26
disable multiple window support on Android crosswalk by default.

DonnaWuDongxia:
BUG=https://crosswalk-project.org/jira/browse/XWALK-3575

[Android] Disable multiple windows support by default
    Crosswalk Android will not support multiple windows, such as window.open
    by default, unless you set XWalkPreferences.SUPPORT_MULTIPLE_WINDOWS.

[Linux] do not register widget extension the app is not widget
    If the application is not a widget, access to exposed "widget" object will cause crash.

PR #2947, opened 2015-03-31
[Android] Do not open default browser when support multiple windows

lincsoon:
When supportMultipleWindows is true, the link will be opened in a new
XWalkView which created in onCreateWindowRequested(), rather than opened
in default browser.

BUG=XWALK-2754


lincsoon:
@DonnaWuDongxia @axinging @fujunwei

axinging:
LGTM.

lincsoon:
You're right! If supportMultibleWindows is false, content::WebContentImpl::CreateNewWindow which is the origin of onCreateWindow should never be called, the the link should be opened in current view. All of these work correctly in WebView but not in XWalkView. It seems like setSupportMultibleWindows has no effect on XWalkView. I'll investigate this issue later and submit another PR. But for this PR, I just want to fix XWALK-2754 that onCreateWindowRequsted is not fired event supportMultibleWindows is true, which is also a release-blocker.

About shouldOpenWithDefaultBrowser, I don't think it's a good idea to disable a feature in existing code, because there must be some reasons to introduce it before. I think @gaochun could give some comments.

lincsoon:
As discussed with @DonnaWuDongxia, the issue that setSupportMultibleWindows has no effect on XWalkView will be fixed by #2928. I'll change the approach of fixing XWALK-2754 and resubmit this patch.

gaochun:
@DonnaWuDongxia and @lincsoon As opening hyper links in new XwalkView will provide more advantages such as sharing cookies, sessions and other resources, I think it's OK to disable opening hyper links with default browser.

lincsoon:
@DonnaWuDongxia

DonnaWuDongxia:
LGTM, Thanks!

PR #2892, opened 2015-03-04
[Android] Optimize the prompt for make_apk.py.

wuhengzhi:
This patch is to optimize the prompt in make_apk.py when the field
length of app version is greater than 3.

BUG=XWALK-3473

PR #2948, opened 2015-03-31
[Android] make_apk: Improve the version handling code related messages.

rakuco:
This is an alternative take at XWALK-3473 (compared to #2892), which is about the error messages and option descriptions we have for the `--app-version`, `--app-versionCode` and `--app-versionCodeBase` options being hard to understand.

While here, I've also simplified the code in both `MakeVersionCode()` and `TryCodeBaseFromVersionName()` and made them more Pythonic and easier to understand.

BUG=XWALK-3473


rakuco:
@lincsoon @wuhengzhi

PR #2949, opened 2015-04-01
Fix compiling error of WebNotification of Crosswalk for linux

PeterWangIntel:
Due to a redundant parameter of the caller has been removed in Version Bump, the callee should be change despondently.

### DIFF
--- a/runtime/browser/linux/xwalk_notification_manager.cc
+++ b/runtime/browser/linux/xwalk_notification_manager.cc
@@ -65,7 +65,6 @@ void XWalkNotificationManager::ShowDesktopNotification(
     const GURL& origin,
     const content::PlatformNotificationData& notification_data,
     scoped_ptr<content::DesktopNotificationDelegate> delegate,
-    int render_process_id,
     base::Closure* cancel_callback) {
   if (!initialized_)
     return;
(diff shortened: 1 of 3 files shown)


PeterWangIntel:
"Bump version to 14.42.326.0". The related code has been missed since it's disabled by default in that time.
During rebase of the patch for  pull request #2934, the error is revealed.
Seems. to enable them is necessary, the trybot will help check missed things. 
Thank you very much.

rakuco:
The patch is correct, but the explanation is not. [The commit you refer to](https://github.com/crosswalk-project/crosswalk/commit/2846aece2557153210b245f0e527a620f3228df3) has nothing to do with this as you can see in the link. The error in #2934 is likely caused by our move to M42 weeks ago. The interface in Chromium changed but the files in Crosswalk were not updated because the code was never built.
Please update the PR description and commit message.

rakuco:
I'm merging the new version as-is because the change is small enough and there is some context here, but for future pull requests please be less concise in your commit message: by looking solely at it one cannot figure out when `content` changed, why and how long these files have been broken, for example.

PeterWangIntel:
ok. thank you.

PR #2951, opened 2015-04-07
[Android] Fix the crash issue when exiting fullscreen for video.

wuhengzhi:
This patch is to fix the crash issue when exiting fullscreen for video.
In the native code, fullscreen requests have been moved to render frame,
then tear down ContentVideoView from content-layer to reduce code
duplication.
Android content embedder has no responsibility for releasing the surface
when exiting fullscreen.

BUG=XWALK-3796

(cherry picked from commit cf4cf2173617e35b352b5aa480aa9f5b4d973865)


wuhengzhi:
@rakuco Please merge this patch to crosswalk-13, thanks.

rakuco:
Does it apply to crosswalk-12 as well?

wuhengzhi:
This issue occurs after canary 13.41.318.0.

PR #2952, opened 2015-04-07
Fix the compiling warning of unused result.

takethathe:
The 'std::system()' call has a return value, if not use it, there would cause a
warning of unused return value, this patch will solve this issue.

### DIFF
--- a/runtime/browser/runtime_platform_util_linux.cc
+++ b/runtime/browser/runtime_platform_util_linux.cc
@@ -46,7 +46,8 @@ void XDGOpen(const std::string& path) {
   GURL url(path);
   if (url.is_valid() && url.SchemeIsHTTPOrHTTPS()) {
     std::string cmd = "xdg-open " + path;
-    std::system(cmd.c_str());
+    if (std::system(cmd.c_str()) != 0)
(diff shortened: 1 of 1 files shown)


takethathe:
@darktears @pozdnyakov 
PTAL! Thanks!

runtime/browser/runtime_platform_util_linux.cc
```diff
@@ -46,7 +46,8 @@ void XDGOpen(const std::string& path) {
   GURL url(path);
   if (url.is_valid() && url.SchemeIsHTTPOrHTTPS()) {
     std::string cmd = "xdg-open " + path;
-    std::system(cmd.c_str());
+    if (std::system(cmd.c_str()) != 0)
```
halton: Maybe it is not this PR problem. Any better option than `std::system()` ? @rakuco

rakuco:
I'd rather go with #2955. Halton is right, the problem is _not_ that the build was failing with GCC 4.9, that was just a symptom of the bigger underlying problem. Calling `system()` is a bad idea 99% of the time. If something is wrong with the way `LaunchProcess()` is being invoked, it is much better to spend some time figuring this out than using `system()`.

PR #2953, opened 2015-04-07
Add PepperFlash hosts in renderer.

seanlong:
In order to implement some functionalties these pepper hosts are need in
renderer. These code is copied from src/chrome/renderer/pepper/.

BUG=XWALK-3670,XWALK-3791

### DIFF
--- a/runtime/renderer/pepper/xwalk_renderer_pepper_host_factory.cc
+++ b/runtime/renderer/pepper/xwalk_renderer_pepper_host_factory.cc
@@ -4,7 +4,13 @@
 
 #include "xwalk/runtime/renderer/pepper/xwalk_renderer_pepper_host_factory.h"
 
+#include "chrome/renderer/pepper/pepper_flash_drm_renderer_host.h"
+#include "chrome/renderer/pepper/pepper_flash_font_file_host.h"
+#include "chrome/renderer/pepper/pepper_flash_fullscreen_host.h"
+#include "chrome/renderer/pepper/pepper_flash_menu_host.h"
+#include "chrome/renderer/pepper/pepper_flash_renderer_host.h"
 #include "content/public/renderer/renderer_ppapi_host.h"
+#include "ppapi/host/ppapi_host.h"
 #include "ppapi/proxy/ppapi_messages.h"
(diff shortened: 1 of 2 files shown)


seanlong:
@pozdnyakov @halton PTAL, some copy&paste again.

halton:
@seanlong Seems the files your copied are untouched. In that way, can we do the `action` by copying them on build-time instead of maintain in the src/xwalk repo?

Second question: PR #2888 is for the flash plugin as well, is this PR last one to get Pepper flash working?

seanlong:
Can we directly add these files to xwalk gyp? I'm not sure as they are coming from src/chrome/.

For the 2rd question, these classes are needed to complete some jobs delegated from plugin process. 
For example, https://crosswalk-project.org/jira/browse/XWALK-3791 is caused by not support drawing glyphs.

halton:
> Can we directly add these files to xwalk gyp? I'm not sure as they are coming from src/chrome/.

Ideally we should not use any files under `src/chrome`, so I think directly include them in gyp is not good idea and suggest to use copy action. But it is only my personal comment. Let's include @rakuco @darktears in loop.

seanlong:
@rakuco @darktears wdyt?

kenchris:
to implement some functionalities? Please explain exactly what it is blocking you from implementing.

seanlong:
More precisely the msg should be "to **support** some functionalities". As Chromium already support them, so I want to borrow Chromium implementation directly. I am not that familiar with PepperFlash impl, so I think borrow existing upstream code is the correct choice.

seanlong:
ping @rakuco @kenchris

rakuco:
I am not familiar at all with this code, so I don't have much to add. We don't normally depend on stuff from `chrome/`, but it's probably better than copying it to our tree. If Flash support is something that makes any sense, I don't have any objections.

@darktears @kenchris, anything to say?

halton:
Think it over again, copying the ***unchanged*** code from `src/chrome/renderer/pepper/` to `src/xwalk` is also kind of ugly way, besides it brings potential hidden bugs which upstream fixes time to time. 

Comparing copying solution and include chrome files solution, I would prefer choose the latter one, which is currently @seanlong implemented. So LGTM from my side.

PR #2954, opened 2015-04-07
XWalkViewShellActivity upgrade

d0ngjun:
Replace FragmentActivity with XWalkActivity.
XWalkViewShellActivity use FragmentActivity as its base Activity, this prevent it to benefit from XWalkActivity.

### DIFF
--- a/runtime/android/core_shell/src/org/xwalk/core/xwview/shell/XWalkViewShellActivity.java
+++ b/runtime/android/core_shell/src/org/xwalk/core/xwview/shell/XWalkViewShellActivity.java
@@ -8,6 +8,7 @@ import java.util.HashMap;
 
 import android.app.ActionBar;
 import android.app.ActionBar.Tab;
+import android.app.Fragment;
 import android.app.FragmentTransaction;
 import android.content.BroadcastReceiver;
 import android.content.Intent;
@@ -17,7 +18,6 @@ import android.graphics.drawable.ClipDrawable;
 import android.os.Bundle;
 import android.os.Looper;
 import android.os.MessageQueue;
-import android.support.v4.app.FragmentActivity;
 import android.support.v4.view.ViewPager;
 import android.util.Log;
 import android.view.KeyEvent;
@@ -36,13 +36,14 @@ import org.chromium.base.BaseSwitches;
 import org.chromium.base.CommandLine;
 import org.chromium.base.library_loader.LibraryLoader;
 import org.chromium.content.browser.TracingControllerAndroid;
+import org.xwalk.core.XWalkActivity;
 import org.xwalk.core.XWalkNavigationHistory;
 import org.xwalk.core.XWalkPreferences;
 import org.xwalk.core.XWalkResourceClient;
 import org.xwalk.core.XWalkUIClient;
 import org.xwalk.core.XWalkView;
 
-public class XWalkViewShellActivity extends FragmentActivity
+public class XWalkViewShellActivity extends XWalkActivity
         implements ActionBar.TabListener, XWalkViewSectionFragment.OnXWalkViewCreatedListener {
     public static final String COMMAND_LINE_FILE = "/data/local/tmp/xwview-shell-command-line";
     private static final String TAG = XWalkViewShellActivity.class.getName();
@@ -126,7 +127,7 @@ public class XWalkViewShellActivity extends FragmentActivity
 
         mActionBar = getActionBar();
         mActionBar.setNavigationMode(ActionBar.NAVIGATION_MODE_TABS);
-        mSectionsPagerAdapter = new SectionsPagerAdapter(this, getSupportFragmentManager(), mActionBar);
+        mSectionsPagerAdapter = new SectionsPagerAdapter(this, getFragmentManager(), mActionBar);
 
         mViewPager = (ViewPager) findViewById(R.id.pager);
         mViewPager.setAdapter(mSectionsPagerAdapter);
@@ -172,7 +173,7 @@ public class XWalkViewShellActivity extends FragmentActivity
     @Override
     public void onDestroy() {
         super.onDestroy();
-        unregisterReceiver(mReceiver);
+        if (mReceiver != null) unregisterReceiver(mReceiver);
         unregisterTracingReceiver();
     }
 
@@ -325,8 +326,8 @@ public class XWalkViewShellActivity extends FragmentActivity
     @Override
     public void onTabSelected(Tab tab, FragmentTransaction ft) {
         mViewPager.setCurrentItem(tab.getPosition());
-        android.support.v4.app.Fragment fragment = mSectionsPagerAdapter.getItem(tab.getPosition());
-        if (fragment!= null && fragment instanceof XWalkViewSectionFragment) {
+        Fragment fragment = mSectionsPagerAdapter.getItem(tab.getPosition());
+        if (fragment != null && fragment instanceof XWalkViewSectionFragment) {
             mActiveView = ((XWalkViewSectionFragment)fragment).getXWalkView();
         } else {
             mActiveView = null;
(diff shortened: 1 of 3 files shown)


d0ngjun:
@halton @shawngao5 please help reivew, thanks!

halton:
@d0ngjun please explain why this replacing is required in details in git log.

lincsoon:
Shared mode provides backward compatibility with codes written upon old rules. As long as the `XWalkView` objects work on embedded mode, there is no need to extend `XWalkActivity`. However, considering future flexibility, it still make sense to replace `FragmentActivity` with `Fragment`. Actually, this is also what I was going to do before.

d0ngjun:
@lincsoon now XWalkView has combined with XWalkActivity, so it's impossible for XWalkViewShellActivity to benefit from the new feature if it not inherit from XWalkActivity(no matter Shared-mode or Embedded mode).

Considering v4.Fragment is used together with FragmentActivity, we need to replace it with Fragment also.

lincsoon:
I don't agree `XWalkView` is associated with `XWalkActivity`. You could see  `XWalkCoreWrapper.initEmbeddedMode()` in `reflectionInit()` of XWalkView.java, it will help `XWalkView` to work on embedded mode without `XWalkAcvitity`. 

Because XWalkCoreShell is always working on embedded mode(?) and `XWalkActivity`'s features are all about shared mode, we could say it's still available currently, but it doesn't mean the same in the future. So, it's still necessary to replace `FragmentActivity`.

d0ngjun:
@lincsoon, although XWalkCoreWrapper provide initEmbeddedMode for XWalkView initialization, XWalkViewShellActivity can not access it.(XWalkCoreWrapper is not a public class, can not use in other package).

Besides, since LZMA will backport into master, we want XWalkViewShellActivity could benefit from it.

runtime/android/core_shell/src/org/xwalk/core/xwview/shell/XWalkViewShellActivity.java
```diff
@@ -359,4 +314,53 @@ public void onXWalkViewCreated(XWalkView view) {
         mProgressMap.put(view, 0);
         XWalkPreferences.setValue(XWalkPreferences.REMOTE_DEBUGGING, true);
     }
+
+    @Override
+    public void onXWalkReady() {
+        setContentView(R.layout.testshell_activity);
```
lincsoon: It's okay to move the initialization of the activity to onCreate() to make the code easier to understand for someone don't know onXWalkReady() well. You can even create an XWalkView instance before onXWalkReady(), but just don't use it.

lincsoon:
@d0ngjun XWalkViewShellActivity doesn't need to call XWalkCoreWrapper.initEmbeddedMode() directly. This method is called by the method reflectionInit() during the instanciation of each xwalk object automatically. So without XWalkActivity, the xwalk environment will be initialized in the timing that first xwalk instance is created.

Actually, the reason shared mode requires XWalkActivity is that we need to prompt end-user to download the xwalk apk through UI interaction. If the environment of shared mode can be initialized in background silently, just like LZMA, the XWalkActivity is not necessary anymore. So the XWalkActivity is all about UI here.

d0ngjun:
@lincsoon Agree. Now we have both working on "how to decouple XWalkView from XWalkActivity", perhaps the statement "Walk is depend on XWalkActivity" is inaccurate. But for XWalkViewShellActivity, it's  a simple fact, if remove the XWalkActivity, we need to do a lot of work to keep it worked in both shared-mode and embedded-mode, just the same thing XWalkActivity already does.

Considering XWalkViewActivity is an example to present how convenient to use embedding API, it's better to just inherit to XWalkActivity.

d0ngjun:
Hi @lincsoon, I've tried move the setContentView into method onCreate, it will cause the app crash(seems it trigger view access in other view/framgment), I think it's safer to keep view related methods in onXWalkReady.

lincsoon:
Could you provide the crash log for me?

d0ngjun:
@lincsoon , base on my LZMA backport code change, if I move setContentView(R.layout.testshell_activity); into getXWalkReady, XWalkCoreShell will crash, the log is here:
E/AndroidRuntime( 3641): FATAL EXCEPTION: main
E/AndroidRuntime( 3641): java.lang.NullPointerException
E/AndroidRuntime( 3641):    at org.xwalk.core.xwview.shell.XWalkViewShellActivity.onXWalkViewCreated(XWalkViewShellActivity.java:319)
E/AndroidRuntime( 3641):    at org.xwalk.core.xwview.shell.XWalkViewSectionFragment.onCreateView(XWalkViewSectionFragment.java:42)
E/AndroidRuntime( 3641):    at android.app.Fragment.performCreateView(Fragment.java:1695)
E/AndroidRuntime( 3641):    at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:885)
E/AndroidRuntime( 3641):    at android.app.FragmentManagerImpl.moveToState(FragmentManager.java:1057)
E/AndroidRuntime( 3641):    at android.app.BackStackRecord.run(BackStackRecord.java:682)
E/AndroidRuntime( 3641):    at android.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:1435)
E/AndroidRuntime( 3641):    at android.app.FragmentManagerImpl.executePendingTransactions(FragmentManager.java:474)
E/AndroidRuntime( 3641):    at android.support.v13.app.FragmentPagerAdapter.finishUpdate(FragmentPagerAdapter.java:145)
E/AndroidRuntime( 3641):    at android.support.v4.view.ViewPager.populate(ViewPager.java:1073)
E/AndroidRuntime( 3641):    at android.support.v4.view.ViewPager.populate(ViewPager.java:919)
E/AndroidRuntime( 3641):    at android.support.v4.view.ViewPager.setAdapter(ViewPager.java:447)
E/AndroidRuntime( 3641):    at org.xwalk.core.xwview.shell.XWalkViewShellActivity.onXWalkReady(XWalkViewShellActivity.java:333)
E/AndroidRuntime( 3641):    at org.xwalk.core.XWalkActivity.onXWalkLibraryMatched(XWalkActivity.java:156)
E/AndroidRuntime( 3641):    at org.xwalk.core.XWalkActivity$DecompressTask.onPostExecute(XWalkActivity.java:621)
E/AndroidRuntime( 3641):    at org.xwalk.core.XWalkActivity$DecompressTask.onPostExecute(XWalkActivity.java:590)
E/AndroidRuntime( 3641):    at android.os.AsyncTask.finish(AsyncTask.java:632)
E/AndroidRuntime( 3641):    at android.os.AsyncTask.access$500(AsyncTask.java:177)
E/AndroidRuntime( 3641):    at android.os.AsyncTask$InternalHandler.handleMessage(AsyncTask.java:648)
E/AndroidRuntime( 3641):    at android.os.Handler.dispatchMessage(Handler.java:108)
E/AndroidRuntime( 3641):    at android.os.Looper.loop(Looper.java:152)
E/AndroidRuntime( 3641):    at android.app.ActivityThread.main(ActivityThread.java:5726)
E/AndroidRuntime( 3641):    at java.lang.reflect.Method.invokeNative(Native Method)
E/AndroidRuntime( 3641):    at java.lang.reflect.Method.invoke(Method.java:511)
E/AndroidRuntime( 3641):    at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:793)
E/AndroidRuntime( 3641):    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:560)
E/AndroidRuntime( 3641):    at dalvik.system.NativeStart.main(Native Method)

lincsoon:
I tested you code moving all code in `onXWalkReady()` back to `onCreate()` and it works well. Please check 3bbe7e2e4bece0be3e48a423a03bd885e7e2712a

lincsoon:
LGTM

PR #2956, opened 2015-04-08
[Android] Fix failed to query contacts with "is" filter.

LaneRuan:
Correct the function name mapping, and check if  contacts list is empty.

BUG=XWALK-2451

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/contacts/ContactConstants.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/contacts/ContactConstants.java
@@ -38,7 +38,7 @@ public class ContactConstants {
     // e.g. When to find "givenName=John", we actually need to find "John" in "givenNames".
     public static final Map<String, String> findFieldMap = createStringMap(new String[]{
         "familyName", "familyNames",
-        "givenName", "givenNames",
+        "givenNames", "givenNames",
         "middleName", "middleNames",
         "additionalName", "additionalNames",
         "honorificPrefix", "honorificPrefixes",
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/contacts/ContactFinder.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/contacts/ContactFinder.java
@@ -235,10 +235,13 @@ public class ContactFinder {
         // 60  David2 Smith   name
         String where = null;
         String[] whereArgs = null;
+        JSONArray returnArray = new JSONArray();
         if (contactIds.size() != 0) {
             where = Data.CONTACT_ID + " in (" + ContactUtils.makeQuestionMarkList(contactIds) + ")";
             whereArgs = contactIds.toArray(new String[contactIds.size()]);
         }
+        else
+            return returnArray;
         Cursor c = null;
         Map<Long, ContactData> dataMap = null;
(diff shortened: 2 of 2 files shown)


runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/contacts/ContactConstants.java
```diff
@@ -38,7 +38,7 @@
     // e.g. When to find "givenName=John", we actually need to find "John" in "givenNames".
     public static final Map<String, String> findFieldMap = createStringMap(new String[]{
         "familyName", "familyNames",
-        "givenName", "givenNames",
+        "givenNames", "givenNames",
```
rakuco: This looks fishy, you're duplicating the second item here.

LaneRuan: Sorry for reply you late.According to idl-def-ContactName in w3c and the crosswalk-test-suite,there is no use for "givenName".So we should change the function name mapping.

rakuco: I have to admit that I'm confused by the whole API implementation, so help here is appreciated. http://www.w3.org/TR/contacts-api/ just says the spec has been discontinued and people are not supposed to implement it. Then there is http://www.w3.org/TR/2011/WD-contacts-api-20110616/, which is an older version of the spec, but the `find()` method there does not resemble what our implementation looks like (and all the fields are mentioned in the singular form, like "givenName" instead of "givenNames"). Can you please point to which spec is actually being implemented here?

LaneRuan: http://www.w3.org/2012/sysapps/contacts-manager-api/ is being implemented here.

LaneRuan:
@halton

runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/contacts/ContactConstants.java
halton: @shawngao5 please involve this review.

rakuco: Ah, thanks. It finally makes sense to me now.

shawngao5:
LTGM

rakuco:
Please backport to the appropriate branches.

halton:
@rakuco do we have criteria what PR need to be back ported? This is a P2 bug. CC @baleboy

rakuco:
I thought the criteria was basically "if it's a bug fix and it's easy enough to backport, then backport", though I don't think this has ever been formally discussed.

halton:
Backport by Cherry-pick seems simple, but for developer is not that simple. Developers need create canary/beta/stable workspace and ensure the backport build pass and unit test pass.

How about we learn chromium upstream experience, just critical bugs (P1 like) and security bugs need to be back port definitely. For others, let's do that case by case or on-demand. 

For this PR, I do not think we need spend time on back porting unless QA or customer complain.

halton:
After sync with @tinazhao @baleboy , we agree to back port this fix the beta branch. @shawngao5 please guide @LaneRuan to do that.

PR #2958, opened 2015-04-08
[Linux] Fix open external URL in browser.

seanlong:
If the default browser is Chrome, it will start zygote process with
SUID sandbox hepler which fail to check the root privilege. As by
default the prctl() has made the newly started process do not grant
any new privileges.

Also see upstream CL: https://codereview.chromium.org/235983009.

BUG=XWALK-3690


seanlong:
@pozdnyakov @rakuco

rakuco:
lgtm, thanks

pozdnyakov:
lgtm

PR #2959, opened 2015-04-08
rebase to chromium 42.0.2311.68

Bysmyyr:
also marked OnRequestFocusTest flaky because bug=XWALK-3863

### DIFF
--- a/test/android/core/javatests/src/org/xwalk/core/xwview/test/OnRequestFocusTest.java
+++ b/test/android/core/javatests/src/org/xwalk/core/xwview/test/OnRequestFocusTest.java
@@ -8,6 +8,7 @@ import android.test.suitebuilder.annotation.SmallTest;
 
 import org.chromium.base.test.util.Feature;
 
+import org.chromium.base.test.util.DisabledTest;
 /**
  * Test suite for OnRequestFocus().
  */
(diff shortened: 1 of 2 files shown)


darktears:
@Bysmyyr you need to rebase your roll out. DEPS.xwalk was changed :). It should be quick.

test/android/core/javatests/src/org/xwalk/core/xwview/test/OnRequestFocusTest.java
```diff
@@ -8,6 +8,8 @@
 
 import org.chromium.base.test.util.Feature;
 
+import android.test.FlakyTest;
```
rakuco: The test you're disabling below is not flaky, it just needs to be disabled. Flaky tests are tests that randomly fail because of race conditions or because they are dependent on the timing or a certain ordering of events. In this case, we just want to skip the test, as it deterministically fails 100% of the time.

rakuco:
Thanks, it's almost there. We still need 01org/ozone-wayland#371 to be merged first, and then you need to roll oz-wl in your patch.

test/android/core/javatests/src/org/xwalk/core/xwview/test/OnRequestFocusTest.java
```diff
@@ -19,9 +21,12 @@ public void setUp() throws Exception {
         super.setUp();
         mOnRequestFocusHelper = mTestHelperBridge.getOnRequestFocusHelper();
     }
-
+/*
     @SmallTest
     @Feature({"OnRequestFocus"})
+    BUG XWALK-3863
+    */
+     @FlakyTest
```
lincsoon: I think just `@DisabledTest` is okay

PR #2960, opened 2015-04-08
Roll DEPS.xwalk

darktears:
Blink-Crosswalk :
e4aaba Fix XDK code so that it uses size_t instead of int.

V8-Crosswalk :
3b7454  [SIMD.js] Fix no exception thrown when get invalid index of
Float32x4Array
7c2b1f1 [SIMD.js] Fix offset of SIMD.Float32x4.load by adding
tarray.byteOffset.
a60b892 Fix XDK code so that it uses size_t instead of int.


rakuco:
lgtm

PR #2965, opened 2015-04-09
[Android] make_apk: Fix Python 3 regression after 2e723a9.

rakuco:
Commit 2e723a9 ("[Android] make_apk: Simplify TryCodeBaseFromVersionName()") introduced a regression when run with Python 3.

In Python 3, `map()` returns an iterator, so we cannot call `extend()` on it. Fix it by using a list comprehension instead of `map()`.

BUG=XWALK-3903


rakuco:
@darktears @lincsoon

darktears:
LGTM

PR #2966, opened 2015-04-09
Fix compilation on Windows 64 bits

darktears:
Code usually uses std functions which returns a size_t and size_t
is not going to fit into an int on 64 bits. Today it triggers a
warning of potential data loss.

We could disable the warning but it's better to fix the code.
Please note that we have various casts, these are mandatory (though
unfortunate) at the moment because the API in Chromium requires
explicitely an int for example. This can be fixed upstream and
bugs have been reported (e.g. crbug.com/167187).

### DIFF
--- a/application/browser/application_security_policy.cc
+++ b/application/browser/application_security_policy.cc
@@ -8,6 +8,7 @@
 #include <map>
 #include <string>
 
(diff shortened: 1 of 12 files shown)


rakuco:
lgtm

PR #2969, opened 2015-04-11

wuhengzhi:
listener was not called.

BUG=XWALK-3647

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
@@ -396,6 +396,10 @@ class XWalkContentsClientBridge extends XWalkContentsClient
                                 String contentDisposition,
                                 String mimeType,
                                 long contentLength) {
+        if (mDownloadListener != null) {
+            mDownloadListener.onDownloadStart(
+                    url, userAgent, contentDisposition, mimeType, contentLength);
+        }
     }
 
     @Override


axinging:
LGTM. Thanks hengzhi.

rakuco:
can this be covered by a test?

wuhengzhi:
@axinging Yes, it's no need to implement ```XWalkView.setDownloadListener```.
@rakuco If we add test case for ```onDowanloadStart```, we should override ```onDownloadStart``` in ```XWalkDownloadListenerImpl```, but ```XWalkDownloadListenerImpl``` is not public.
Two other ways:
1. Extends from XWalkContentsClient, override ```onDowanloadStart```.
2. Extends from ```DownloadListener``` like as ```XWalkDownloadListenerImpl```, override ```onDowanloadStart```.
Does these two ways make sense? Thanks.

rakuco:
I'd appreciate some feedback from @lincsoon here.
On a broader level, isn't the JIRA ticket you reference in the PR message about making the API public anyway?

lincsoon:
The code is LGTM. But as @rakuco saied, the description of the ticket is confusing, please update it.

rakuco:
Can you please describe what's changed? The latest version of the patch is much bigger than the original one, so what's new? Can the last two commits be squashed together?

wuhengzhi:
The last two commits is to implement ```XWalkView.setDownloadListener``` and add test case. They can be squashed together. Thanks.

lincsoon:
I'm confused too. We'll expose a listener, then implement the interface internally, and expose the class of the internal implementation as well? Let's just submit bug fixing and new API separately.

wuhengzhi:
I will split this PR, thanks.

wuhengzhi:
@lincsoon This PR could first be merged. Thanks.

rakuco:
does it need to be backported?

wuhengzhi:
Yes, I will backport to Crosswalk-14 and Crosswalk-13.

PR #2971, opened 2015-04-13
[Android] make_apk: Stop suggesting architectures without downloads.

rakuco:
Even though we do not provide official Crosswalk downloads for ARM64 and
x86_64, we were still nagging users to provide APKs for both
architectures whenever they used make_apk.py in embedded mode.

Stop doing that and only suggest the architectures for which we provide
official downloads (ARM and x86). The code supporting ARM64 and x86_64
remains in place, and we do allow users to specify them in the
command-line if they have build Crosswalk for those architectures
themselves.

While here, reword and improve the message listing all the APKs that
have been created by the make_apk invocation.

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -23,7 +23,7 @@ from customize import VerifyPackageName, CustomizeAll, \
                       ParseParameterForCompressor, CustomizeManifest
 from extension_manager import GetExtensionList, GetExtensionStatus
 from handle_permissions import permission_mapping_table
-from util import AllArchitectures, CleanDir, GetVersion, RunCommand, \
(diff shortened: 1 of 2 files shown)


rakuco:
@baleboy @bspencer @shawngao5 @darktears

rakuco:
Grrmbl. This approach doesn't make the x86_64 bot very happy...

rakuco:
I'm now taking a less aggressive approach: users can still specify x86_64 and arm64 in the command line and make_apk will work if they have produced shared libraries for those architectures themselves, but make_apk itself will not complain if APKs for those architectures have not been produced.

shawngao5:
LGTM

PR #2973, opened 2015-04-13
crosswalk-13: Roll Chromium 42.0.2311.68.

rakuco:
This also requires us to disable OnRequestFocusTest on Android due to
XWALK-3863.


rakuco:
@baleboy @darktears

darktears:
Lgtm

PR #2975, opened 2015-04-14
[Linux] Add locale support for minimal-ui buttons.

seanlong:
The xwalk_strings could also be used by other string localization usages.

### DIFF
--- a/runtime/app/xwalk_main_delegate.cc
+++ b/runtime/app/xwalk_main_delegate.cc
@@ -95,7 +95,15 @@ void XWalkMainDelegate::InitializeResourceBundle() {
 
   pak_file = pak_dir.Append(FILE_PATH_LITERAL("xwalk.pak"));
 #endif
+
+#if !defined(OS_ANDROID) && !defined(OS_TIZEN)
+  ui::ResourceBundle::InitSharedInstanceWithLocale(
+      "en-US", nullptr, ui::ResourceBundle::DO_NOT_LOAD_COMMON_RESOURCES);
+  ResourceBundle::GetSharedInstance().AddDataPackFromPath(
+      pak_file, ui::SCALE_FACTOR_NONE);
+#else
   ui::ResourceBundle::InitSharedInstanceWithPakPath(pak_file);
+#endif
 }
 
 content::ContentBrowserClient* XWalkMainDelegate::CreateContentBrowserClient() {
--- a/resources/resource_ids
+++ b/resources/resource_ids
@@ -30,4 +30,8 @@
   "xwalk/sysapps/sysapps_resources.grd": {
     "includes": [32175],
   },
+
+  "xwalk/runtime/resources/xwalk_strings.grd": {
+    "messages": [32200],
+  },
 }
--- a/runtime/browser/ui/native_app_window_desktop.cc
+++ b/runtime/browser/ui/native_app_window_desktop.cc
@@ -7,7 +7,9 @@
 
 #include "base/strings/utf_string_conversions.h"
 #include "content/public/browser/web_contents.h"
+#include "grit/xwalk_strings.h"
 #include "third_party/skia/include/core/SkPaint.h"
+#include "ui/base/l10n/l10n_util.h"
 #include "ui/views/controls/webview/webview.h"
 #include "ui/gfx/canvas.h"
 #include "ui/views/background.h"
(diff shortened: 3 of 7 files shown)


runtime/app/xwalk_main_delegate.cc
```diff
@@ -95,7 +95,15 @@ void XWalkMainDelegate::InitializeResourceBundle() {
 
   pak_file = pak_dir.Append(FILE_PATH_LITERAL("xwalk.pak"));
 #endif
+
+#if defined(OS_LINUX) && !defined(OS_TIZEN)
+  ui::ResourceBundle::InitSharedInstanceWithLocale(
+      "en-US", NULL, ui::ResourceBundle::DO_NOT_LOAD_COMMON_RESOURCES);
```
pozdnyakov: nit: nullptr

pozdnyakov:
looks sane, but bots aren't happy

pozdnyakov:
From Tizen Common bot:

[  124s] ../../../home/abuild/rpmbuild/BUILD/crosswalk/src/xwalk/runtime/browser/ui/native_app_window_desktop.cc:10:32: fatal error: grit/xwalk_strings.h: No such file or directory
[  124s]  #include "grit/xwalk_strings.h"

native_app_window_desktop.cc should not be build for Tizen

rakuco:
Is there any specific reason why this shouldn't work on Tizen?

As far as the build system goes, why copy the files around instead of generating them at their final destination?

rakuco:
> As far as the build system goes, why copy the files around instead of generating them at their final destination?

If that doesn't work, take a look at `xwalk_application.gypi` and how it uses `copies` instead of adding a script just for that.

seanlong:
> As far as the build system goes, why copy the files around instead of generating them at their final destination?

The `grit_out_dir` needs to be located in intermediate dir to store generated string id header file.

> take a look at xwalk_application.gypi and how it uses copies instead of adding a script just for that.

I looked at `copies`, seems it doesn't support recursively copy files in a directory. Would you provide more detailed example on how will it work on this case? Thanks.

seanlong:
ping @rakuco @kenchris

rakuco:
Sorry, I had to spend some time looking at what Chromium currently does first. As far as I can see, it does something more complex that involves repacking several pak files for some reason. So a few questions and comments:

* Is the repacking logic not needed for us?
* Is there a reason for providing another grd file instead of adding the new messages to `runtime/resources/xwalk_resources.grd`?
* I got gyp to work without the new Python script by listing which files need to be copied instead of just doing a recursive copy. Chromium sort of cheats because it hardcodes the locale list in `build/common.gypi` and then runs a script that spits out the entire paths that need to be copied based on that. I think we don't need something like this yet. This is what I did, see if it looks sane to you:
```python
    {
      'target_name': 'xwalk_strings',
      'type': 'none',
      'variables': {
        'grit_out_dir': '<(SHARED_INTERMEDIATE_DIR)/xwalk/locales',
      },
      'actions': [
        {
          'action_name': 'generate_xwalk_strings',
          'variables': {
            'grit_grd_file': 'runtime/resources/xwalk_strings.grd',
            'grit_resource_ids': 'resources/resource_ids',
          },
          'includes': [ '../build/grit_action.gypi' ],
        },
      ],
      'copies': [
        {
          'destination': '<(PRODUCT_DIR)/locales',
          'files': ['<(grit_out_dir)/en-US.pak',
                    '<(grit_out_dir)/zh-CN.pak'],
        },
      ],
    },
```

seanlong:
> Is the repacking logic not needed for us?

If we have multiple locale string resource paks like Chromium then we may need the similar repacking logic. Currently we only have one locale resource file then repack is not needed.

> Is there a reason for providing another grd file instead of adding the new messages to runtime/resources/xwalk_resources.grd?

I want to separate string resource and other resource like JS wrapper for extensions. It will be inefficient and weird for each pak files in locales/* to include JS resources not needed for localization.

After this change we will have 2 types of pak files: xwalk.pak for all general resources, pak files in locales/ which contains localization string resources. Crosswalk will always load the first one and load the send one depends on system language setting.

> I got gyp to work without the new Python script by listing which files need to be copied instead of just doing a recursive copy. Chromium sort of cheats because it hardcodes the locale list in build/common.gypi and then runs a script that spits out the entire paths that need to be copied based on that. I think we don't need something like this yet. This is what I did, see if it looks sane to you:

However we need to add each pak file in the gyp by this way, if we add supports for other 20 languages, the list will be long...

Anyway I will change this way if you think we should start from it.

rakuco:
Thanks for the explanation!
Please go ahead with the version without `copy_locales.py`; while in principle it's more practical to do a recursive copy I don't think it justifies rewriting `cp -R` in Python.

PR #2978, opened 2015-04-15
[Android] Add Halton and SunLin as owner for Android port

halton:
Including 8 'OWNERS' files for android related sub folders.


halton:
@darktears This PR is take effect that @lincsoon and myself nominated as crosswalk Android owners.

PR #2979, opened 2015-04-15
crosswalk-13: [Android] make_apk: Stop suggesting architectures without downloads.

rakuco:
Even though we do not provide official Crosswalk downloads for ARM64 and
x86_64, we were still nagging users to provide APKs for both
architectures whenever they used make_apk.py in embedded mode.

Stop doing that and only suggest the architectures for which we provide
official downloads (ARM and x86). The code supporting ARM64 and x86_64
remains in place, and we do allow users to specify them in the
command-line if they have build Crosswalk for those architectures
themselves.

While here, reword and improve the message listing all the APKs that
have been created by the make_apk invocation.

(cherry picked from commit fc30d8ad1938fe9010a5e2fd7e5fdd3f5816dcd1)

Conflicts:
	app/tools/android/make_apk.py
	app/tools/android/util.py

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -23,13 +23,28 @@ from customize import VerifyPackageName, CustomizeAll, \
                       ParseParameterForCompressor, CustomizeManifest
 from extension_manager import GetExtensionList, GetExtensionStatus
 from handle_permissions import permission_mapping_table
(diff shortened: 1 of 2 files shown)


rakuco:
@baleboy @darktears @shawngao5 @bspencer

rakuco:
re-ping

boogunote:
LGTM

shawngao5:
LGTM

PR #2980, opened 2015-04-15
Move to Chromium 42.0.2311.90.

rakuco:
This is the first stable M42 release. In addition, we are now tracking
ozone-wayland's Milestone-Mazama, which is oz-wl's branch tracking M42.


rakuco:
@darktears @baleboy @kenchris

PR #2981, opened 2015-04-15
crosswalk-13: Move to Chromium 42.0.2311.90, the first M42 stable release.

rakuco:
While here, also make ozone-wayland track Milestone-Mazama, its
M42-based branch.


rakuco:
@darktears @kenchris @baleboy

PR #2982, opened 2015-04-15
DEPS.xwalk: Roll blink-crosswalk (2a0323e -> 6e19c2e).

rakuco:
* 6e19c2e Merge pull request #52 from rakuco/revert-faulty-revert
* d00a85f Revert "Revert "Merge 192508 "Focus a page only if the page is different from cu..."""

This roll is to revert a reverted upstream commit that was not supposed
to be pushed to blink-crosswalk (it was reverted temporarily while we
were debugging XWALK-3863).


rakuco:
@darktears @kenchris

PR #2983, opened 2015-04-16
[Android] Implement the embedding API for setDownloadListener.

wuhengzhi:
This patch is to implement the API for setDownloadListener in XWalkView.
Customer will implement the download listener as needed and override the
default listener in XWalkViewInternal.
Add test case for downloading file.

BUG=XWALK-3958


wuhengzhi:
@axinging @lincsoon PTAL, thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -971,10 +971,8 @@ public void setXWalkWebChromeClient(XWalkWebChromeClient client) {
         mContent.setXWalkWebChromeClient(client);
     }
 
-    /**
-     * @hide
-     */
-    public void setDownloadListener(DownloadListener listener) {
+    @XWalkAPI
+    public void setDownloadListener(XWalkDownloadListenerInternal listener) {
```
axinging: Please help to add document:
/**
* Registers the interface to be used when content can not be handled by
* the rendering engine, and should be downloaded instead. This will replace
* the current handler.
* @param listener an implementation of XWalkDownloadListenerInternal 
* @since 5.0
*/

axinging:
LGTM.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkDownloadListenerImplInternal.java
```diff
@@ -23,10 +23,11 @@
 import java.io.OutputStream;
 
 import org.xwalk.core.internal.AndroidProtocolHandler;
-import org.xwalk.core.internal.DownloadListener;
 import org.xwalk.core.internal.R;
+import org.xwalk.core.internal.XWalkDownloadListenerInternal;
 
-class XWalkDownloadListenerImpl implements DownloadListener {
+@XWalkAPI(impl = XWalkDownloadListenerInternal.class, createExternally = true)
```
lincsoon: We don't need to expose XWalkDownloadListenerImpl. But this class could be the default behavior if the developer doesn't override XWalkDownloadListener

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkDownloadListenerImplInternal.java
```diff
@@ -23,10 +23,10 @@
 import java.io.OutputStream;
 
 import org.xwalk.core.internal.AndroidProtocolHandler;
-import org.xwalk.core.internal.DownloadListener;
 import org.xwalk.core.internal.R;
+import org.xwalk.core.internal.XWalkDownloadListenerInternal;
 
-class XWalkDownloadListenerImpl implements DownloadListener {
+class XWalkDownloadListenerImplInternal extends XWalkDownloadListenerInternal {
```
lincsoon: It's okay to keep the name `XWalkDownloadListenerImpl`. If a class name is postfixed with `Internal`, it generally means we intend to expose it and use it as embedding API

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkDownloadListenerInternal.java
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import android.content.Context;
+
+@XWalkAPI(createExternally = true)
+public abstract class XWalkDownloadListenerInternal {
```
lincsoon: Please use interface instead of abstract class

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkDownloadListenerInternal.java
```diff
@@ -0,0 +1,32 @@
...
+     * Notify the host application that a file should be downloaded.
+     * @param url The full url to the content that should be downloaded
+     * @param userAgent the user agent to be used for the download.
+     * @param contentDisposition Content-disposition http header, if present.
+     * @param mimetype The mimetype of the content reported by the server
+     * @param contentLength The file size reported by the server
+     * @since 5.0
+     */
+    @XWalkAPI
+    public abstract void onDownloadStart(String url, String userAgent,
```
lincsoon: No need to declare the method as `abstract`

lincsoon:
Because the reflection generator doesn't support the interface declaration, we use abstract class temporarily.

PR #2984, opened 2015-04-16
Add a new hook to create Google Play Services's directory structure.

rakuco:
Ever since M42, building Chromium (and consequently Crosswalk) for
Android has required installing the Google Play Services library via the
`android` tool first. However, doing so involves manually accepting an
End User License Agreement (EULA), which makes things very difficult to
automate and also breaks our checkout+configuration process for users
who are building Crosswalk for the first time.

Since that library is actually needed only by Chromium but not
Crosswalk, we have been working around this requirement. When Crosswalk
started tracking Chromium M42, we added a few changes to
chormium-crosswalk. One of them was making gyp not fail if some
directories created by the Google Play Services library were not
present.

This solution needs to be rethought for the upcoming M43, as there is
now a build system check in the android_tools repository to fail if a
certain file installed by that library is not found.

The script being added by this change is the answer to this, and is
being landed before M43 itself because it also works independently of
it. If the android_tools repository is checked out, it tries to create
the Google Play Services directories gyp looks for as well as an empty
file with the name that the android_tools build system checks. This way,
we get rid of the corresponding change in chromium-crosswalk, prepare
for the upcoming change in android_tools and make most of these things
be controlled from the crosswalk repository itself.

Related to: XWALK-3955

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -44,6 +44,16 @@ hooks = [
     "action": ["python", "src/xwalk/tools/fetch_deps.py", "-v"],
   },
   {
+    # Make src/third_party/android_tools/android_tools.gyp shut up and not exit
+    # if the Google Play Services library is not installed (after
+    # https://chromium-review.googlesource.com/#/c/247861). We do not use it on
+    # Crosswalk, and installing it is a manual process that involves accepting
+    # an EULA.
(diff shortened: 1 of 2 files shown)


rakuco:
@darktears @kenchris @mrunalk @halton

PR #2985, opened 2015-04-17
Add SSL error page to handle SSL certificate error

PeterWangIntel:
When SSL certificate error occurs, show an Interstitial Page to give user a chance to choose to proceed or just close the App.

Add "xwalk::SSLErrorPage" as a child class of "content::InterstitialPageDelegate" with these methods overridden: 
(1) Show()
     Creates the instance of "content::InterstitialPage" and show it.
(2) GetHTMLContents()
     Provides the html for the interstitial page. The web page includes a link to allow user to ignore the SSL error and proceed the network request.
(3)  OnProceed()
     When user chooses to proceed, this function will be invoked to call the "callback" provided by "content".

BUG=https://crosswalk-project.org/jira/browse/XWALK-3668

### DIFF
--- a/runtime/browser/ssl_error_page.cc
+++ b/runtime/browser/ssl_error_page.cc
@@ -1,0 +1,91 @@
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/runtime/browser/ssl_error_page.h"
+
+#include "base/strings/string_number_conversions.h"
+#include "content/public/browser/interstitial_page.h"
+#include "content/public/browser/web_contents.h"
+
+using content::WebContents;
+using net::SSLInfo;
+
+namespace xwalk {
+
+enum UserCommandsFromPage {
+  CMD_DONT_PROCEED = 0,
+  CMD_PROCEED = 1,
+};
+
+SSLErrorPage::SSLErrorPage(WebContents* web_contents,
+                           int cert_error,
+                           const SSLInfo& ssl_info,
+                           const GURL& request_url,
+                           const base::Callback<void(bool)>& callback)
+    : web_contents_(web_contents),
+      cert_error_(cert_error),
+      ssl_info_(ssl_info),
+      request_url_(request_url),
+      callback_(callback),
+      interstitial_page_(nullptr) {
+}
+
+SSLErrorPage::~SSLErrorPage() {
+}
+
+void SSLErrorPage::Show() {
+  DCHECK(!interstitial_page_);
+  interstitial_page_ = content::InterstitialPage::Create(
+      web_contents_, true, request_url_, this);
+  interstitial_page_->Show();
+}
+
+void SSLErrorPage::OnProceed() {
+  // Allow certificate
+  callback_.Run(true);
+  callback_.Reset();
+}
+
+void SSLErrorPage::OnDontProceed() {
+  // Deny certificate
+  callback_.Run(false);
+  callback_.Reset();
+}
+
+void SSLErrorPage::CommandReceived(const std::string& command) {
+  int cmd = 0;
+  bool retval = base::StringToInt(command, &cmd);
+  DCHECK(retval);
+  switch (cmd) {
+    case CMD_DONT_PROCEED:
+      interstitial_page_->DontProceed();
+      break;
+    case CMD_PROCEED:
+      interstitial_page_->Proceed();
+      break;
+    default:
+      break;
+  }
+}
+
+// TODO(Peter Wang): Provide a more user-friendly page including
+// icons, localized strings, and the details of SSL error etc.
+std::string SSLErrorPage::GetHTMLContents() {
+  std::string proceed_link =
+      "<p><a href=\"javascript:proceed()\"> Proceed to " +
+      request_url_.spec() + "</a></p>";
+  return "<!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">"
+         "<html><head><title>Untrusted Connection</title>"
+         "<script>function proceed() {"
+         "window.domAutomationController.setAutomationId(1);"
+         "window.domAutomationController.send(1);}"
+         "</script></head>"
+         "<body><h1>This Connection is Untrusted</h1>"
+         "<p>This site uses an invalid security certificate.</p>"
+         "<p>Close or proceed if your understand the risks</p>" +
+         proceed_link +
+         "</body></html>";
+}
+
+}  // namespace xwalk
--- a/runtime/browser/ssl_error_page.h
+++ b/runtime/browser/ssl_error_page.h
@@ -1,0 +1,55 @@
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef XWALK_RUNTIME_BROWSER_SSL_ERROR_PAGE_H_
+#define XWALK_RUNTIME_BROWSER_SSL_ERROR_PAGE_H_
+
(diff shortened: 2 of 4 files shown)


runtime/browser/ssl_error_page.h
```diff
@@ -0,0 +1,67 @@
...
+  void OnDontProceed() override;
+  void CommandReceived(const std::string& command) override;
+  std::string GetHTMLContents() override;
+
+  content::WebContents* web_contents() const { return web_contents_; }
+  std::string SSLErrorDetails();
+  GURL request_url() const { return request_url_; }
+  content::InterstitialPage* interstitial_page() const {
+    return interstitial_page_;
+  }
```
halton: Above getter seems not used at all, suggest to remove.

PeterWangIntel:
Below is the snapshot of the SSL error page.

runtime/browser/ssl_error_page.cc
```diff
@@ -0,0 +1,86 @@
...
+         "<html><head><title>Untrusted Connection</title>"
+         "<script>function proceed() {"
+         "window.domAutomationController.setAutomationId(1);"
+         "window.domAutomationController.send(1);}"
+         "</script></head>"
+         "<body><h1>This Connection is Untrusted</h1>"
+         "<p>This site uses an invalid security certificate.</p>"
+         "<p>Close or proceed if your understand the risks</p>" +
+         proceed_link +
+         "</body></html>";
```
halton: There is concern about localization here, I'm okay to add TODO.

PeterWangIntel: ok.

runtime/browser/ssl_error_page.h
PeterWangIntel: ok

PeterWangIntel:
@halton @pozdnyakov @rakuco @kenchris 
The patch has been updated, please help to review, thank you very much.

runtime/browser/ssl_error_page.h
```diff
@@ -0,0 +1,62 @@
...
+}
+
+namespace xwalk {
+
+class SSLErrorPage : public content::InterstitialPageDelegate {
+ public:
+  enum UserCommandsFromPage {
+    CMD_DONT_PROCEED = 0,
+    CMD_PROCEED = 1,
+  };
```
halton: I do not think this need to be public unless you have plan to make SSLErrorPage as parent class in future. If not that case, move this enum into .cc

runtime/browser/ssl_error_page.h
```diff
@@ -0,0 +1,62 @@
...
+  SSLErrorPage(content::WebContents* web_contents,
+               int cert_error,
+               const net::SSLInfo& ssl_info,
+               const GURL& request_url,
+               const base::Callback<void(bool)>& callback);
+
+  ~SSLErrorPage() override;
+
+  // Show an interstitial page to let user to choose the next action
+  void Show();
```
halton: Why not move code in `Show()` to ctor?

runtime/browser/ssl_error_page.h
```diff
@@ -0,0 +1,62 @@
...
+
+ protected:
+  // Overridden methods of InterstitialPageDelegate
+  void OnProceed() override;
+  void OnDontProceed() override;
+  void CommandReceived(const std::string& command) override;
+  std::string GetHTMLContents() override;
+
+  std::string SSLErrorDetails();
```
halton: This func seems not called at anywhere, am I missing something?

runtime/browser/xwalk_content_browser_client.cc
```diff
@@ -275,6 +276,20 @@ void XWalkContentBrowserClient::AllowCertificateError(
                                   &cancel_request);
   if (cancel_request)
     *result = content::CERTIFICATE_REQUEST_RESULT_TYPE_DENY;
+#else
+  content::RenderFrameHost* render_frame_host =
+      content::RenderFrameHost::FromID(render_process_id, render_frame_id);
```
halton: Need not valid render_frame_host?

runtime/browser/ssl_error_page.h
PeterWangIntel: ok.

runtime/browser/ssl_error_page.h
PeterWangIntel: If so, we have to always be careful to put 'interstitial_page_'  as the last member of class 'SSLErrorPage'. Since 'SSLErrorPage' is the delegate of 'interstitial_page_' (its class is 'content::InterstitialPage'), many memebers (e.g. ssl_info_) of  'SSLErrorPage' will be referred (e.g., through GetHTMLContents), so we have make sure all members being initialized before 'interstitial_page_'.

runtime/browser/ssl_error_page.h
PeterWangIntel: Actually, it should be invoked in 'GetHTMLContents()' in future when the UI of SSL error page is designed.

runtime/browser/xwalk_content_browser_client.cc
PeterWangIntel: In the next statement, 'WebContents::FromRenderFrameHost' is pretty strong, even feed  'nullptr' to him.

runtime/browser/ssl_error_page.h
halton: Then add `GetHTMLContents()` with TODO or delete this unused and almost empty func. It is your call to choose either one.

runtime/browser/xwalk_content_browser_client.cc
halton: ok

runtime/browser/ssl_error_page.h
halton: Still can not understand your reason:
1. Why `interstitial_page_` must be as last member? Any compiler error?
2. What real problem you referred private members in `GetHTMLContents()`? An example could be helpful

runtime/browser/ssl_error_page.h
PeterWangIntel: ok. It'll removed in new update.

runtime/browser/ssl_error_page.h
PeterWangIntel: Did you mean you hope the code like this
SSLErrorPage::SSLErrorPage {
...
Show();
...
}
The 'content::InterstitialPage::Show' will invoke 'xwalk::SSLErrorPage::GetHTMLContents', and the last one will refer to, for example, 'SSLErrorPage::ssl_info_' to get the details of error (how detailed is depending on our design of UI) to show.
So, we have to add a comment to provent to initialize related memebers after it.

PeterWangIntel:
@halton 
The patch has been updated according to the comments, please have a look. Thank you.

runtime/browser/ssl_error_page.cc
```diff
@@ -0,0 +1,93 @@
...
+using net::SSLInfo;
+
+namespace xwalk {
+
+namespace {
+enum UserCommandsFromPage {
+  CMD_DONT_PROCEED = 0,
+  CMD_PROCEED = 1,
+};
+}  // namespace
```
halton: I think this anonymous namespace is not needed.

halton:
Much much better, thanks for rework, LGTM with above nit comment.

runtime/browser/ssl_error_page.cc
PeterWangIntel: ok.

halton:
lgtm, ping @darktears @kenchris

runtime/browser/ssl_error_page.cc
pozdnyakov: yeah, and it would be nice to load it from resources rather than hardcode it here. It's OK to do this later.

pozdnyakov:
lgtm

PR #2986, opened 2015-04-20
[Android] Provide interface for initializing XWalkView asynchronously

lincsoon:
The XWalkView's environment is initialized at the moment first XWalkView
object is instanciated. The initialization is kind of a heavy task which
includes loading native library, staring processes and so on, may cause
the UI thread freezed for a moment.

So this patch provides a interface to initialize the environment in
background and notify the host application once ready. This patch also
does the following things:

- Dismisses XWalkLibraryListener and integrates it into XWalkCoreWrapper
- Adds new error type for the case that the CPU architecture
  mismatches. Removes unused code with same function from XWalkViewInternal
- Makes the init method of XWalkViewDelegate can take the argument with
  the type Context

BUG=XWALK-3971


lincsoon:
@halton @d0ngjun @shawngao5 please help to review!

lincsoon:
@axinging @fujunwei

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,70 @@
...
+        public abstract void onXWalkInitFinished();
+        public abstract void onXWalkInitFailed(XWalkInitResult result);
+    }
+
+    public enum XWalkInitResult {
+        SUCCESS,
+        ARCHITECTURE_MISMATCH,
+    }
+
+    public static void initAsync(Context context, XWalkInitListener listener) {
```
shawngao5: Where is this function called from?

halton:
@lincsoon are we target the lite branch instead of master?

lincsoon:
The lite branch is using the old implementation of XWalkActivity. Since the implementation of shared mode has just settled down today, I'll transplant new implementation to the lite branch soon. But before that, please review whether the async interface fits your needs.

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,70 @@
...
+
+        @Override
+        protected XWalkInitResult doInBackground(Void... params) {
+            try {
+                XWalkCoreWrapper.initEmbeddedMode();
+            } catch (VerifyError e) {
+                return XWalkInitResult.ARCHITECTURE_MISMATCH;
+            }
+            publishProgress(50);
+
```
d0ngjun: Seems the decompressing work will inject here, since now we don't know how long it will consume, I'm not sure whether this progress update is meaningful.

lincsoon: Exactly! It should be inserted at the start of `doInBackground()`. If we can't get the progress, it's okay to remove the `onXWalkInitProgressed()`.

lincsoon:
A sample here:

```
public class MyXWalkActivity extends Activity implements XWalkInitializer.XWalkInitListener {
    @Override
    protected void onCreate(Bundle savedInstanceState) {
        XWalkInitializer.initAsync(this, this);
    }

    @Override
    public void onXWalkInitFinished() {
        setContentView(R.layout.activity_xwalkview);
        mXWalkView = (XWalkView) findViewById(R.id.xwalkview);
        mXWalkView.load("file:///android_asset/index.html", null);
    }

    @Override
    public void onXWalkInitFailed(XWalkInitResult result) {
        Assert.fail();
    }
```

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,56 @@
...
+public class XWalkInitializer {
+    public interface XWalkInitListener {
+        public void onXWalkInitFinished();
+        public void onXWalkInitFailed(XWalkInitResult result);
+    }
+
+    public enum XWalkInitResult {
+        SUCCESS,
+        ARCHITECTURE_MISMATCH,
```
halton: nit, remove comma at end

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public static void initAsync(Context context, XWalkInitListener listener) {
+        new InitializationTask(context, listener).execute();
+    }
+
+    private static class InitializationTask
+            extends AsyncTask<Void, Integer, XWalkInitResult> {
```
halton: I think amend above two lines won't exceed 120 chars, so keep it as one line.

halton:
@lincsoon What is difference with `org.xwalk.core.XWalkActivity.onXWalkReady()`? I think this might confuse people.

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,56 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core;
+
+import android.content.Context;
+import android.os.AsyncTask;
+
+public class XWalkInitializer {
+    public interface XWalkInitListener {
```
d0ngjun: XWalkActivity need to implement XWalkInitListener also?

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
lincsoon: Aren't the line length limit is 100 for java files?

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
lincsoon: No. All of `XWalkInitializer` is for external use.

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
halton: Sorry, I wanted to say 100 char, it only 96 chars.

lincsoon:
`XWalkAcvitiy` and `XWalkInitializer` are two irrelevant interfaces, they can't be used together.

1. `XWalkActivity` is a UI wrapper, `XWalkInitializer` is a non-UI wrapper.
2. `XWalkActivity` takes over everything, including initialization and downloading and so on. `XWalkInitializer` is only about initialization.
3. `XWalkActivity` supports both of embedded mode and shared mode, `XWalkInitializer` only supports embedded mode.

- To use shared mode, XWalkActivity is the only way.
- To use LZMA, there are two choices, XWalkActivity and XWalkInitializer.
- To use simple embedded mode, the developer can use `XWalkActivity` and `XWalkInitializer`, or just use `XWalkView` directly.

lincsoon:
If we could subdivide the interface of `XWalkInitializer` to several steps of initialization, we can let `XWalkActivity` use `XWalkInitializer`. But now `XWalkInitializer` provides an all-in-one interface and always tries to link the embedded library, this kind of interface doesn't fit the needs of `XWalkActivity`.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -108,25 +109,25 @@ public final void onMethodCallMissed(ReflectMethod method) {
     }
 
     @Override
-    public final void onXWalkLibraryStartupError(LibraryStatus status, Throwable e) {
```
d0ngjun: Could we implement the XWalkCoreListener in a subclass? since we don't want this mixed with other XWalkActivity callbacks or expose it to developers either.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -16,16 +16,33 @@
...
     public static final String XWALK_APK_PACKAGE = "org.xwalk.core";
     public static final String WRAPPER_PACKAGE = "org.xwalk.core";
     public static final String BRIDGE_PACKAGE = "org.xwalk.core.internal";
 
     private static XWalkCoreWrapper sProvisionalInstance;
     private static XWalkCoreWrapper sInstance;
-    private static XWalkLibraryListener sListener;
+    private static XWalkCoreListener sListener;
 
     private static final String TAG = "XWalkActivity";
```
d0ngjun: Should we really need recognized as "XWalkActivity"?

lincsoon:
@fujunwei @axinging please help to review

runtime/android/core/src/org/xwalk/core/XWalkLibraryLoader.java
```diff
@@ -0,0 +1,266 @@
...
+    private static final String TAG = "XWalkLib";
+
+    private static AsyncTask<Void, Integer, Integer> sActiveTask;
+    private static boolean sPreparedToInit = false;
+    private static boolean sCheckedDecompression = false;
+
+    public static void prepareToInit() {
+        sPreparedToInit = true;
+        XWalkCoreWrapper.handlePreInit();
+    }
```
d0ngjun: could remove prepareToInit, remove sPreparedToInit, and init XWalkCoreWrapper.sReservedObjects in its definition, as:
private static LinkedList<Object> sReservedObjects = new LinkedList<Object>();

Then check for sPreparedToInit in line 45 could remove also.

runtime/android/core/src/org/xwalk/core/XWalkLibraryLoader.java
```diff
@@ -0,0 +1,266 @@
...
+            return false;
+        }
+
+        sActiveTask = new DecompressionTask(listener, context);
+        sActiveTask.execute();
+        return true;
+    }
+
+    public static void cancelDecompression() {
+        DecompressionTask task = (DecompressionTask) sActiveTask;
```
d0ngjun: Do not need this assign, can invoke method cancel directly.
Besides, I think add null check for sActiveTask would be better.

runtime/android/core/src/org/xwalk/core/XWalkLibraryLoader.java
```diff
@@ -0,0 +1,305 @@
...
+                int totalSize = cursor.getInt(totalIdx);
+                int downloadSize = cursor.getInt(downloadIdx);
+                if (totalSize > 0) publishProgress(downloadSize, totalSize);
+
+                int statusIdx = cursor.getColumnIndex(DownloadManager.COLUMN_STATUS);
+                int status = cursor.getInt(statusIdx);
+                if (status == DownloadManager.STATUS_FAILED ||
+                        status == DownloadManager.STATUS_SUCCESSFUL) {
+                    return status;
+                } else if (status == DownloadManager.STATUS_PAUSED) {
```
d0ngjun: Do we need these xxxCount?
I think this could only incur unpredicated failure, could remove this, user will cancel the downloading if they thought it's too long.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -13,194 +12,146 @@
...
-import org.xwalk.core.XWalkLibraryListener.LibraryStatus;
+/**
+ * <p>The appropriate call order is:
+ * handlePreInit() -> initXWalkLibrary() -> initXWalkEnvironment() -> handlePostInit() -> over</p>
+ */
+class XWalkCoreWrapper {
+    private static final String XWALK_APK_PACKAGE = "org.xwalk.core";
+    private static final String WRAPPER_PACKAGE = "org.xwalk.core";
+    private static final String BRIDGE_PACKAGE = "org.xwalk.core.internal";
+    private static final String TAG = "XWalkLib";
```
d0ngjun: use "XWalkCoreWrapper" instead.

lincsoon: This tag is shared by multiple classes to indicate that it's related to the library

runtime/android/core/src/org/xwalk/core/XWalkLibraryLoader.java
lincsoon: Removed total timeout but keep retry timeout remaining. For the timeout error, we can provide the user with specific message and prompt to retry.

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,132 @@
...
+    @Override
+    public final void onInitializationCompleted() {
+        sInstance = null;
+        mListener.onXWalkInitCompleted();
+    }
+
+    private void start() {
+        if (XWalkLibraryLoader.startDecompression(this, mContext)) return;
+
+        final int status = XWalkLibraryLoader.initXWalkLibrary(mContext);
```
d0ngjun: startDecompression will execute an AsyncTask, is it too soon to call initXWalkLibrary after it?

lincsoon: startDecompression returns false means that the decompression is not started.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -24,192 +21,202 @@
...
-        mCoreWrapper = null;
-
-        mIsXWalkReady = false;
-        mIsVisible = false;
-        mReservedObjects = new LinkedList<Object>();
-        mReservedMethods = new LinkedList<ReflectMethod>();
-
-        XWalkCoreWrapper.reset(null, this);
-    }
+    private static class XWalkLibraryListener
```
d0ngjun: since XWalkLibraryListener is a private inner-class, we can remove the final modifier now. :-)

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,148 @@
...
+        if (sInstance == null) return false;
+        Log.d(TAG, "Cancel initializing Crosswalk environment");
+        return sInstance.cancel();
+    }
+
+    private XWalkLibraryListener mLibraryListener;
+    private XWalkInitListener mInitListener;
+    private Context mContext;
+
+    private static class XWalkLibraryListener
```
d0ngjun: Same with XWalkActivity.XWalkLibraryListener.

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -0,0 +1,148 @@
...
+
+import org.xwalk.core.XWalkLibraryInterface.DecompressionListener;
+import org.xwalk.core.XWalkLibraryInterface.InitializationListener;
+
+/**
+ * XWalkInitializer is a helper to initialize Crosswalk environment without UI interaction. Due to
+ * this limitation, XWalkInitializer works only in embedded mode, or when appropriate Crosswalk
+ * library is already installed on the device in shared mode.
+ */
+public class XWalkInitializer {
```
d0ngjun: IMPORTANT:
You need to ensure all methods about XWalkInitListener must invoked at UI thread.

lincsoon: Yes, it's ensured by AsyncTask.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: Done

d0ngjun:
LGTM

halton:
I'm ryly on d0ongjun's lgtm, will merge

PR #2964, opened 2015-04-09
To implement GTK File Picker of Crosswalk for Linux

PeterWangIntel:
Two classes added:
(1) xwalk::Gtk2UI
     It's factory-class of xwalk::SelectFileDialogGTK2
(2) xwalk::SelectFileDialogGTK2
     Child class of "ui::SelectFileDialog" and a wrraper to create Gtk standard file picker. Most code of it is taken from Chrome.

This is a transitional patch working until our WebUI-based file picker is mature.

PR #2988, opened 2015-04-21
To link libgtk2ui of Chrome directly

PeterWangIntel:
Maybe, we can make GTK2-related UI as an outside library and link to it.
At current stage, we can directly use 'libgtkui.a' (target 'gtk2ui') built by Chrome, in future, we can move these code out as an independent project to add our own code or WebUI eventraully.
With this patch, Crosswalk for Linux can support GTK-based 'File Picker', and input method context, and font to save our effort to move the code case by case.

### DIFF
--- a/runtime/browser/ui/gtk2_ui.h
+++ b/runtime/browser/ui/gtk2_ui.h
@@ -1,0 +1,13 @@
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef XWALK_RUNTIME_BROWSER_UI_GTK2_UI_H_
+#define XWALK_RUNTIME_BROWSER_UI_GTK2_UI_H_
+
+#include "ui/views/linux_ui/linux_ui.h"
+
+// Return the unique instance of class 'Gtk2UI'.
+extern views::LinuxUI* BuildGtk2UI();
+
+#endif  // XWALK_RUNTIME_BROWSER_UI_GTK2_UI_H_
--- a/runtime/browser/xwalk_browser_main_parts.cc
+++ b/runtime/browser/xwalk_browser_main_parts.cc
@@ -27,6 +27,9 @@
 #include "xwalk/application/browser/application_system.h"
 #include "xwalk/extensions/browser/xwalk_extension_service.h"
 #include "xwalk/extensions/common/xwalk_extension_switches.h"
+#if defined(USE_GTK_UI)
+#include "xwalk/runtime/browser/ui/gtk2_ui.h"
+#endif
 #include "xwalk/runtime/browser/xwalk_runner.h"
 #include "xwalk/runtime/common/xwalk_runtime_features.h"
 #include "xwalk/runtime/common/xwalk_switches.h"
@@ -138,6 +141,10 @@ void XWalkBrowserMainParts::PreEarlyInitialization() {
 #if defined(USE_WEBUI_FILE_PICKER)
   ui::LinuxShellDialog::SetInstance(BuildWebUI());
   wm_state_.reset(new wm::WMState);
+#elif defined(USE_GTK_UI)
+  views::LinuxUI* gtk2_ui = BuildGtk2UI();
+  gtk2_ui->Initialize();
+  views::LinuxUI::SetInstance(gtk2_ui);
 #endif
 #endif
 }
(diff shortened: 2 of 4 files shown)


PeterWangIntel:
Move the head of git to SHA:d10289d26 (Bump version to 14.42.328.0), apply my patch, build, the xwalk works well. 
@rakuco @kenchris @joone @seanlong 
What do you think about this solution?

halton:
If #2964 closed, can this PR be closed as well?

seanlong:
+1 as this component seems has no dependencies on other parts of src/chrome.
BTW, should we add these resources needed to xwalk.pak?

PeterWangIntel:
@seanlong 
By my investigation, no need. The resources involved come from GTK.

pozdnyakov:
lgtm, @rakuco so we're OK to add a GTK dep?

kenchris:
@pozdnyakov GTK dep on what platform?

PeterWangIntel:
Update patch to enable it on Linux by default.  Once  the WebUI-based File Picker mature, we can revert this small patch.
@halton @rakuco also.
It needs 'libgtk-dev' for Linux trybot.

runtime/browser/ui/gtk2_ui.h
```diff
@@ -0,0 +1,13 @@
...
+
+#ifndef XWALK_RUNTIME_BROWSER_UI_GTK2_UI_H_
+#define XWALK_RUNTIME_BROWSER_UI_GTK2_UI_H_
+
+#include "ui/views/linux_ui/linux_ui.h"
+
+// Return the unique instance of class 'Gtk2UI'.
+extern views::LinuxUI* BuildGtk2UI();
+
+#endif  // XWALK_RUNTIME_BROWSER_UI_GTK2_UI_H_
```
halton: Since this header is simple a including header and extern, and only used in `xwalk_browser_main_parts.cc`, I'd suggest move the logic to `xwalk_browser_main_parts.cc` directly.

seanlong:
@kenchris @rakuco could this be merged now? Other features like IME have dependency on this.

> @pozdnyakov GTK dep on what platform?

Linux platform except Tizen, currently only Linux desktop I think.

runtime/browser/ui/gtk2_ui.h
PeterWangIntel: @halton 
My thought was that we might need to fork the "libgtk2ui" from Chromium in future as an independent library to add some own modifications, and the "gtk2_ui.h" is kept as the interface file of library. 
But, it's ok to drop this file right now. We can add it when there is a real need. 
The patch will be updated soon.

halton: Ah, now I understood, you're keeping same file structure as `src/chrome/browser/ui/libgtk2ui/gtk2_ui.h`, then I'm okay as current.

halton:
lgtm, but you need rebase your code.

runtime/browser/ui/gtk2_ui.h
PeterWangIntel: thank you.

PeterWangIntel:
ok, will be updated soon.

halton:
LGTM, ping @rakuco

halton:
As we decided in PRC-FIN meeting, we should have GTK version of file picker for Linux first. Cause the WebUI implementation (https://github.com/crosswalk-project/chromium-crosswalk/pull/218) is not mature enough, even upstream stick on GTK version for that reason. Let's merge this PR first. 

At the same time, we should keep improve webUI impl, maybe upstream is best place. When it is mature, let's drop the GTK one.

Add @baleboy to confirm.

runtime/browser/xwalk_browser_main_parts.cc
```diff
@@ -27,6 +27,9 @@
 #include "xwalk/application/browser/application_system.h"
 #include "xwalk/extensions/browser/xwalk_extension_service.h"
 #include "xwalk/extensions/common/xwalk_extension_switches.h"
+#if defined(USE_GTK_UI)
```
kenchris: is this really the best FLAG? I think that it would be better with a more specific one

kenchris:
lgtm

runtime/browser/xwalk_browser_main_parts.cc
PeterWangIntel: Since all the GTK dependencies are summarized in 'libgtk2', including File Picker and some Input Methods supporting, and I believe we will revert this patch eventually, so, can we just use the name by now? Or any suggestion is fine for me.

PeterWangIntel:
@halton @pozdnyakov @kenchris @rakuco ,
Then can we merge this patch now? Since there is another patch (Input Method related) is blocked by this patch, it's really painful to rebase. 
Thank you very much for understanding.

rakuco:
let's go for it then

PR #2989, opened 2015-04-21
Revert "Add a new hook to create Google Play Services's directory structure."

rakuco:
The script needs to be in chromium-crosswalk itself, otherwise our
Content Shell builds will break since they do not check out Crosswalk at
all.

This is already done in our M43 chromium-crosswalk, so we only need to
remove the script from this repository.

This reverts commit f4f6b7599628912275b8d3533947b3b4d1b00346.

Related to: XWALK-3955


rakuco:
@darktears @kenchris

kenchris:
lgtm

PR #2990, opened 2015-04-22
[Android] Refactoring .so file decompressor.

shawngao5:
Extract the code for decompressing .so file from internal module to a
specify class.

BUG=XWALK-3971


shawngao5:
@lincsoon @halton @d0ngjun

shawngao5:
I keep the decompressing code in XWalkCompressUtil.java and XalkViewDelegate.java to prevent the compiling error. So, I would say the safe way to work with @lincsoon's code is,
* Firstly, merge my PR.
* Then, @lincsoon uses this decompressor class and remove all references to XWalkCompressUtil and XalkViewDelegate.
* The final step, I delete all legacy decompressing code.

PR #2991, opened 2015-04-22
[Android] Fix filepicker crash issue.

wuhengzhi:
This patch is to fix the filepicker crash issue on some customized ROMs.
In some ROMs, "/sdcard/Pictures" was removed, verify this directory
before create image file.

BUG=XWALK-3983

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -155,6 +155,7 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
     static final String PLAYSTORE_DETAIL_URI = "market://details?id=";
     public static final int INPUT_FILE_REQUEST_CODE = 1;
     private static final String TAG = XWalkViewInternal.class.getSimpleName();
+    private static final String PATH_PREFIX = "file:";
 
     private XWalkContent mContent;
     private Activity mActivity;
@@ -679,7 +680,10 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
                     if (dataString != null) {
                         results = Uri.parse(dataString);
                     }
+                    deleteImageFile();
                 }
+            } else if (Activity.RESULT_CANCELED == resultCode) {
+                deleteImageFile();
             }
 
             mFilePathCallback.onReceiveValue(results);
@@ -1059,18 +1063,11 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
         Intent takePictureIntent = new Intent(MediaStore.ACTION_IMAGE_CAPTURE);
         if (takePictureIntent.resolveActivity(getActivity().getPackageManager()) != null) {
             // Create the File where the photo should go
-            File photoFile = null;
-            try {
-                photoFile = createImageFile();
-                takePictureIntent.putExtra("PhotoPath", mCameraPhotoPath);
-            } catch (IOException ex) {
-                // Error occurred while creating the File
-                Log.e(TAG, "Unable to create Image File", ex);
-            }
-
+            File photoFile = createImageFile();
             // Continue only if the File was successfully created
             if (photoFile != null) {
-                mCameraPhotoPath = "file:" + photoFile.getAbsolutePath();
+                mCameraPhotoPath = PATH_PREFIX + photoFile.getAbsolutePath();
+                takePictureIntent.putExtra("PhotoPath", mCameraPhotoPath);
                 takePictureIntent.putExtra(MediaStore.EXTRA_OUTPUT,
                         Uri.fromFile(photoFile));
             } else {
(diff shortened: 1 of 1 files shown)


wuhengzhi:
@lincsoon @axinging @fujunwei Please help to review, thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1103,8 +1103,21 @@ private File createImageFile() throws IOException {
         // Create an image file name
         String timeStamp = new SimpleDateFormat("yyyyMMdd_HHmmss").format(new Date());
         String imageFileName = "JPEG_" + timeStamp + "_";
-        File storageDir = Environment.getExternalStoragePublicDirectory(
-                Environment.DIRECTORY_PICTURES);
+        String state = Environment.getExternalStorageState();
```
lincsoon: It's better to check the start at the start.

```
private File createImageFile() throws IOException {
    String state = Environment.getExternalStorageState();
    if (!state.equals(Environment.MEDIA_MOUNTED)) {
        Log.e(TAG, Error Log);
        return null;
    }

    ......

    File storageDir = Environment.getExternalStoragePublicDirectory(
            Environment.DIRECTORY_PICTURES);
    if (!storageDir.exists()) storageDir.mkdirs();
```

lincsoon:
About the code calling `createImageFile()` in `showFileChooser()`, I have some questions although it's old problems.

1. Is it necessary to call `takePictureIntent.putExtra("PhotoPath", mCameraPhotoPath);` if the photoFile is null
2. Is it okay to put `mCameraPhotoPath` into `takePictureIntent` before changing the path.
3. The code is bloated. If we don't change the logic, the code could be:
```
    try {
        File photoFile = createImageFile();
        takePictureIntent.putExtra("PhotoPath", mCameraPhotoPath);
        mCameraPhotoPath = "file:" + photoFile.getAbsolutePath();
        takePictureIntent.putExtra(MediaStore.EXTRA_OUTPUT, Uri.fromFile(photoFile));
    } catch (IOException | NullPointerException ex) {
        // Error occurred while creating the File
        Log.e(TAG, "Unable to create Image File", ex);
        takePictureIntent = null;
    }
```

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1103,8 +1103,21 @@ private File createImageFile() throws IOException {
...
+
+        // FIXME: If the external storage state is not "MEDIA_MOUNTED", we need to get
+        // other volume paths by "getVolumePaths()" when it was exposed.
+        if (state.equals(Environment.MEDIA_MOUNTED)) {
+            storageDir = Environment.getExternalStoragePublicDirectory(
+                    Environment.DIRECTORY_PICTURES);
+        }
+        if (storageDir == null) {
+            Log.e(TAG, "Can not get the storage directory.");
+            return null;
```
lincsoon: If we use the exception, then throw an exception here too, the return value should be always valid.
Or we can catch the IOException inside this method, and return null if the IOException happened, so that the caller doesn't have to deal with the exception.

wuhengzhi:
1. Is it necessary to call takePictureIntent.putExtra("PhotoPath", mCameraPhotoPath); if the photoFile is null.<br>It's no necessary.
2. Is it okay to put mCameraPhotoPath into takePictureIntent before changing the path.<br>It's okay but it's better to change the path at first.
<br>Thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1099,20 +1096,44 @@ public boolean showFileChooser(ValueCallback<Uri> uploadFile, String acceptType,
...
+            );
+        } catch (IOException ex) {
+            // Error occurred while creating the File
+            Log.e(TAG, "Unable to create Image File", ex);
+        }
         return imageFile;
     }
 
+    private boolean deleteImageFile() {
+        if (mCameraPhotoPath != null && mCameraPhotoPath.contains(PATH_PREFIX)) {
```
lincsoon:
```
if (mCameraPhotoPath == null || !mCameraPhotoPath.contains(PATH_PREFIX)) {
    return false;
}
```

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1099,20 +1096,44 @@ public boolean showFileChooser(ValueCallback<Uri> uploadFile, String acceptType,
...
-        );
+        if (!storageDir.exists()) storageDir.mkdirs();
+        File imageFile = null;
+        try {
+            imageFile = File.createTempFile(
+                    imageFileName,  /* prefix */
+                    ".jpg",         /* suffix */
+                    storageDir      /* directory */
+            );
+        } catch (IOException ex) {
```
lincsoon:
```
try {
    return File.createTempFile(imageFileName, ".jpg", storageDir);
} catch (IOException ex) {
}
return null;
```

wuhengzhi:
@lincsoon Done, Thanks.

PR #2995, opened 2015-04-24
Re-enable SIMD.js after M43 rebasing.

huningxin:
BUG=XWALK-4008


rakuco:
lgtm

rakuco:
The Linux test failures look problematic, can you resubmit or check if you can reproduce them locally?

huningxin:
All bots are green. @rakuco , thanks!

PR #2997, opened 2015-04-28

fujunwei:

BUG=XWALK-2132

(cherry picked from commit 6d89b1d4dc988bb47e8244f9e7ee08108baaf01d)


fujunwei:
@rakuco Back port commits for XWALK-2132 to crosswalk-13

PR #2998, opened 2015-04-28

fujunwei:

BUG=XWALK-2132

(cherry picked from commit 6d89b1d4dc988bb47e8244f9e7ee08108baaf01d)

Conflicts:
	runtime/browser/xwalk_content_browser_client.cc

### DIFF
--- a/runtime/browser/xwalk_content_browser_client.cc
+++ b/runtime/browser/xwalk_content_browser_client.cc
@@ -47,6 +47,7 @@
 #endif
 
 #if defined(OS_ANDROID)
+#include "base/android/locale_utils.h"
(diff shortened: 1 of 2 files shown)


fujunwei:
@rakuco Back port commits for XWALK-2132 to crosswalk-12

denisazevedo:
Any plans to build the next Crosswalk 12 with this fix?
It's working correctly on +13.42.319.9 but it's not working on 12.41.296.9.

rakuco:
I don't think there are any plans to make further Crosswalk 12 releases. Our current stable version is 13 (with 14 being beta and 15 canary), so this means older branches don't get new releases.

PR #3001, opened 2015-04-29
[Android] Back port commits for dropped IndexedDB  to Crosswalk 11

fujunwei:
Because of multiple profile support, from crosswalk 9 IndexedDB
will be put under profile's subfolder.
Issue is reported from https://github.com/MobileChromeApps/mobile-chrome-apps/issues/554.

BUG=XWALK-4047

(cherry picked from commit 36231eaa85f54622210f9999819cb68994d64432)


rakuco:
Crosswalk 11 is unmaintained.

PR #3004, opened 2015-04-29
Lzma backport

d0ngjun:
This involve 2 commits here:

1. LZMA compression support
   Use lzma compress libxwalkcore.so to reduce apk size.
This feature is backport from lite.

2. add support for runtime build flag.
    Add --enable_lzma parameter into make_apk.py:
    True: compress the libxwalkcore.so into libxwalkcore.so.$arch, then move it into apk's res/raw directory.
    False: do not compress libxwalkcore.so (default behavior).
Besides, also include some refactor work, for the shared-mode/embedded mode/embedded mode with lzma support is too much complicated when they involved in one script.

@halton @shawngao5 @lincsoon please help review, thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,7 @@ Thumbs.db
 cscope.*
 Session.vim
 UPSTREAM.blink
+third_party/lzma_sdk/*.bz2
+third_party/lzma_sdk/*.7z
+third_party/lzma_sdk/src
+raw/
--- a/DEPS
+++ b/DEPS
@@ -38,6 +38,12 @@ hooks = [
     "action": ["python", "src/xwalk/tools/generate_gclient-xwalk.py"],
   },
   {
+    # Fetch and prepare SevenZip LZMA SDK
+    "name": "download-lzma-sdk",
+    "pattern": ".",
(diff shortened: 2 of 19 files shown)


darktears:
Do this includes a way for use to customized the "Uncompressing" screen?

kenchris:
@darktears you mean that a uncompressing screen will be shown? Is it not possible to do so as part of the install process?

darktears:
Yes there is a screen shown on the first launch. I don't think you can hook up something when an APK is installed, so we need to uncompress at first run.

halton:
> Yes there is a screen shown on the first launch. I don't think you can hook up something when an APK is installed, so we need to uncompress at first run.

Thanks for @darktears 's explanation. This is correct.

> Do this includes a way for use to customized the "Uncompressing" screen?

Yes, this is a future feature which is tracked in https://crosswalk-project.org/jira/browse/XWALK-3475

kenchris:
ok lgtm then

darktears:
This broke Windows badly in the python code. `file` and `tar` commands do not exist. We should leverage python infrastructure when possible.

PR #3006, opened 2015-04-29
Fix build on Windows 64 bits.

darktears:
Compilation was failing in some ffmpeg headers with _rmdir not
defined. This function should be defined by os_support.h in ffmpeg
however it requires a set of defines to be properly set by
config.h. Some changes in blink now add blink headers
in the include path and the config.h of Blink was picked up by
the compiler instead of the config.h of ffmpeg. Note that this
behavior is also in 32 bits but only Windows 64 bits requires
a special set of defines. Fix is to remove the content dependency
which is not needed and depend only on blink_headers instead of the
full blink.

### DIFF
--- a/extensions/extensions.gyp
+++ b/extensions/extensions.gyp
@@ -5,11 +5,10 @@
       'type': 'static_library',
       'dependencies': [
         '../../base/base.gyp:base',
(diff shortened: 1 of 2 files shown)


rakuco:
lgtm

PR #3008, opened 2015-05-06
DEPS.xwalk: Roll v8-crosswalk (7c73519->108d89f)

chuan9:


rakuco:
your checkout is _very_ old, please sync with master and then update the patch.

PR #3009, opened 2015-05-06
DEPS.xwalk: Roll v8-crosswalk(f08c04f->108d89f)

chuan9:


rakuco:
thanks!

PR #3012, opened 2015-05-06
crosswalk-13: [Android] Fix make_apk fail to pack app when API level is lower than 21

rakuco:
Some attributes in layout files introduced by upstream is added from
API level 21. These resources are also included in xwalk core library.
When building app by make_apk, command "ant release" will fail if API
level is lower than 21.

BUG=XWALK-3219
BUG=XWALK-4103

(cherry picked from commit 541b2f1769871ce5e214ba1615453dc3944662d8)


rakuco:
@baleboy @lincsoon @halton

halton:
lgtm

PR #3013, opened 2015-05-06
Fix build on Windows in relation of lzma.

darktears:
Python code was using 'file' and 'tar' commands. These are not
available on Windows, we should use the Python classes instead.

### DIFF
--- a/third_party/lzma_sdk/download_lzma_sdk.py
+++ b/third_party/lzma_sdk/download_lzma_sdk.py
@@ -12,8 +12,8 @@
 import os
 import shlex
 import shutil
-import subprocess
 import sys
+import tarfile
 import urllib2
 
 SDK_URL = 'http://downloads.sourceforge.net/sevenzip/lzma920.tar.bz2'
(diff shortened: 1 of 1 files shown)


darktears:
@rakuco : what do you say?

third_party/lzma_sdk/download_lzma_sdk.py
```diff
@@ -32,24 +33,15 @@ def DownloadFromURL(url, dest_file):
...
-    cmd = ['tar', 'xjf', lzma_file, '-C', SDK_DIR, 'Java']
-  elif mime_type == 'application/x-7z-compressed':
-    cmd = ['7zr', 'x', '-o' + SDK_DIR, lzma_file, 'Java']
-  else:
-    print('Error: mime type of %s is not supported.' % lzma_file)
-    return False
-
+  tar = tarfile.open(lzma_file)
+  java_dir = [tarinfo for tarinfo in tar.getmembers()
+  	     if tarinfo.name.startswith("Java/")]
```
d0ngjun: use os.sep instead of '/', or just check if it's a directory.

third_party/lzma_sdk/download_lzma_sdk.py
```diff
@@ -14,6 +14,7 @@
 import shutil
 import subprocess
```
halton: subprocess can be deleted.

halton: This script should return immediately on non-linux host cause the lzma sdk is not needed for Android target so far.

rakuco:
@halton @d0ngjun, is there a reason why this needs to be downloaded instead of just being added to the tree?

d0ngjun:
Hi @rakuco , we do not need to maintain these lzma code, and we can easily change the url if we want to use a different version of LZMA SDK.

rakuco:
That's orthogonal to whether the code should be kept in the tree or not -- replacing it is a matter of just changing the files. Right now downloading the code adds additional complexity for no real gain. I think it's better to just add those small files to a separate directory (`third_party/lzma_sdk`, for example, with a README similar to Chromium's `README.chromium`s) and avoid relying on the network.

darktears:
Other solution is implemented here : https://github.com/crosswalk-project/crosswalk/pull/3014

halton:
LGTM, thanks for @rakuco and @darktears driving this.

PR #3014, opened 2015-05-07
Commit the lzma SDK instead of fetching it.

darktears:
Remove the hooks to fetch the SDK, there is no need to download it,
it's small enough to import the files directly in our repository.

Added README.xwalk to describe the licensing.


rakuco:
We also need a `README.xwalk` similar to the `README.chromium` files that Chromium adds to its third-party libraries, with information such as where the code came from, the license, what it does etc. See the `README.chromium`s in `src/third_party`, for example.

darktears:
If we go that road, in order for the infra, as well as dev checkouts to not fail we would need to land https://github.com/crosswalk-project/crosswalk/pull/3015 first.

rakuco:
ping @halton @d0ngjun

d0ngjun:
LGTM

rakuco:
given the bot errors are unrelated, let's merge

PR #3015, opened 2015-05-07
Remove lzma SDK from .gitignore.

darktears:
This is the first step to land the LZMA SDK part of Crosswalk
source tree. Taking the files away from .gitignore will make gclient
clean up the source tree of the old extracted files coming from the old
way to fetch the SDK (downloading it and then extracting it).

We should be able to import the SDK in a following commit without
the infra complaining about overwriting untracked files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,7 +40,4 @@ Thumbs.db
 cscope.*
 Session.vim
 UPSTREAM.blink
-third_party/lzma_sdk/*.bz2
-third_party/lzma_sdk/*.7z
-third_party/lzma_sdk/src
 raw/


rakuco:
ping @halton @d0ngjun

d0ngjun:
Hi @rakuco, our primary design is just like what you suggest: move these decompressing file into our repo, but after thorough consideration, we decide to download the code from their official website, since:
1. these code is a stable version published by 7-zip community, we can avoid take the effort to review it(besides, we're not experts in this field either).
2. mixing up their code into our repo means we need to maintain them, and merge the code change manually if we need someday.
3. these java files have different coding style, we have to refactor the code to make them complied with ours.
4. download do not need additional network access.
in DEPS, you can see other resources need to download also. Actually, download_lzma_sdk.py  referenced other download script in our chromium third_party repo.

BTW, the license of these code is public domain, we can add it if really needed, it does not take much effort, refer to $chromium/src/third_party/README.chromium.template.

rakuco:
Let me try to clear up some of the confusion. Your decision might have been based on some wrong assumptions, so please take a look at my responses below and reconsider it.

> 1. these code is a stable version published by 7-zip community, we can avoid take the effort to review it(besides, we're not experts in this field either).

Importing third-party code does not mean the code needs to be closely reviewed by experts like if it was written by us. The fact that it is written by somebody else is fine, and it is just another external piece of code we rely on, just like code coming from Chromium or Blink.

Conversely, if the reviewing effort was really supposed to work like that, the fact that the code is downloaded from the network would not make it exempt from such review process.

> 2. mixing up their code into our repo means we need to maintain them, and merge the code change manually if we need someday.

That is true, and the decision of whether to import some new third-party library needs to be done on a case by case basis. In this specific case, 7zip makes very infrequent releases, and we do not need to update every time there is a new version anyway if the existing version is working.

So this should be OK.

> 3. these java files have different coding style, we have to refactor the code to make them complied with ours.

Nope. Third-party code is imported as-is, it is not supposed to be adapted. Think of the pain involved with every update from upstream if that were to be the case.

> 4. download do not need additional network access.
> in DEPS, you can see other resources need to download also. Actually, download_lzma_sdk.py referenced other download script in our chromium third_party repo.

Just like Chromium downloads some files in its hooks, they are generally not auxiliary libraries. Libraries are in `src/third_party`, and the things being downloaded are normally toolchains (NaCl, clang etc) and other kinds of tarballs.

> BTW, the license of these code is public domain, we can add it if really needed, it does not take much effort, refer to $chromium/src/third_party/README.chromium.template.

Yes, it's precisely what is done in #3014.

halton:
@rakuco Either download and importing code is not so different. The Crosswalk-Lite team discussed the two solutions months ago, the reasons that @d0ngjun mentioned does make a little bit more sense, we choose that. Actually, this is very trivial thing thing, we should not waste more time on this. 

What PR #3013 that @darktears is trying to fix is the windows problem. I've already give my comment that we can simply disable download/extract on Windows. Which is a simple way to fix.

My opinion is to close #3014 and this one and continue to fix the python issue on Windows in #3013.

rakuco:
Given that at least some of the reasons for making the decision appear to have been based on false assumptions, I still question the value of not adding the code to the tree, and would like to go with #3014 instead of #3013.

rakuco:
To anyone else watching this pull request, Alexis, Halton, Dongjun and I have spoken privately and agreed to move on with this one and #3015 now that #3013 has landed and the Windows build is working.

rakuco:
lgtm, waiting for @halton or @d0ngjun before merging.

d0ngjun:
LGTM

PR #3017, opened 2015-05-11
[Android] Make the version code for IA higher than for ARM

lincsoon:
Many x86 ABI devices can also run ARM binaries. This patch reorders
the version codes so that the binary with best performance runs on
each device. For example, the x86 APK has the highest version code,
followed by ARMv7. Thus x86 binaries will be preferred for x86
devices and ARMv7 binaries preferred for ARMv7 devices

### DIFF
--- a/xwalk.gyp
+++ b/xwalk.gyp
@@ -939,19 +939,19 @@
       'variables': {
         'variables': {
           'conditions': [
-            ['android_app_abi=="x86"', {
+            ['android_app_abi=="armeabi"', {
               'version_code_shift%': 1,
             }],
             ['android_app_abi=="armeabi-v7a"', {
               'version_code_shift%': 2,
             }],
-            ['android_app_abi=="armeabi"', {
+            ['android_app_abi=="arm64-v8a"', {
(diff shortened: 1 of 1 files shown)


lincsoon:
@rakuco @baleboy please help to review

rakuco:
lgtm. is there a related JIRA ticket?

baleboy:
There is this, but I'm not sure it is related https://crosswalk-project.org/jira/browse/XWALK-4177

halton:
LGTM, will merge.

Remember, we're under crosswallk-app-tools switching, this version code change should be included to app-tools as well.

lincsoon:
@rakuco Could you please cherry-pick this patch to the Crosswalk-14 branch?

rakuco:
> Remember, we're under crosswallk-app-tools switching, this version code change should be included to app-tools as well.

@r0b5t4: ^

baleboy:
@lincsoon you should submit the cherry-pick yourself

lincsoon:
@baleboy I see. Because rakuco has helped with this patch, I'll do that next time.

PR #3019, opened 2015-05-12
crosswalk-14: [Android] Make the version code for IA higher than for ARM

rakuco:
Many x86 ABI devices can also run ARM binaries. This patch reorders
the version codes so that the binary with best performance runs on
each device. For example, the x86 APK has the highest version code,
followed by ARMv7. Thus x86 binaries will be preferred for x86
devices and ARMv7 binaries preferred for ARMv7 devices

(cherry picked from commit 73fce35624f950acfbd79136ef27feaf0c3c11c5)

### DIFF
--- a/xwalk.gyp
+++ b/xwalk.gyp
@@ -939,19 +939,19 @@
       'variables': {
         'variables': {
           'conditions': [
-            ['android_app_abi=="x86"', {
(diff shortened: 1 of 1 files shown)


rakuco:
@lincsoon @halton

halton:
LGTM

PR #3020, opened 2015-05-12
XWALK-3569

rob-close:
[Android] Correctly handle package name change at build time.

Sometimes there is a need to change the package name of an application at
build time.  This is usually done when more than one version of an application
is required.  For example, a free vs a paid version of an application.
An application package name can be changed with aapt, using the
--rename-manifest-package parameter
See this page for further description:
 http://www.piwai.info/renaming-android-manifest-package/

This change has that working correctly.

BUG=XWALK-3569

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
@@ -161,8 +161,13 @@ class XWalkViewDelegate {
         }
 
         ResourceExtractor.setMandatoryPaksToExtract(MANDATORY_PAKS);
-        final int resourcesListResId = context.getResources().getIdentifier(
-                XWALK_RESOURCES_LIST_RES_NAME, "array", context.getPackageName());
+        int resListResId = context.getResources().getIdentifier(
+                XWALK_RESOURCES_LIST_RES_NAME, "array", context.getClass().getPackage().getName());
+        if (resListResId == 0) {
+            resListResId = context.getResources().getIdentifier(
+                    XWALK_RESOURCES_LIST_RES_NAME, "array", context.getPackageName());
+        }
+        final int resourcesListResId = resListResId;
         final AssetManager assets = context.getAssets();
         if (!context.getPackageName().equals(context.getApplicationContext().getPackageName()) ||
                 resourcesListResId != 0) {
@@ -209,7 +214,11 @@ class XWalkViewDelegate {
                     if (resourcesListResId != 0) {
                         String resourceName = resource.split("\\.")[0];
                         int resId = context.getResources().getIdentifier(
-                                resourceName, "raw", context.getPackageName());
+                                resourceName, "raw", context.getClass().getPackage().getName());
+                        if (resId == 0) {
+                            resId = context.getResources().getIdentifier(
+                                    resourceName, "raw", context.getPackageName());
+                        }
                         try {
                             if (resId != 0) return context.getResources().openRawResource(resId);
                         } catch (NotFoundException e) {


fujunwei:
@lincsoon There are lots of using the `getPackageName` in xwalk, should we align with the solution?

axinging:
Hi, if you add 
BUG=XWALK-3569
to the commit logs, JIRA will automatically link to this PR.

baleboy:
Bug ID added. Was this the only update needed for a LGTM?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -209,7 +214,11 @@ public InputStream interceptLoadingForResource(String resource) {
                     if (resourcesListResId != 0) {
                         String resourceName = resource.split("\\.")[0];
                         int resId = context.getResources().getIdentifier(
-                                resourceName, "raw", context.getPackageName());
```
axinging: Better warp these as a function.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -161,8 +161,13 @@ public static void init(XWalkViewInternal xwalkView) throws UnsatisfiedLinkError
         }
 
         ResourceExtractor.setMandatoryPaksToExtract(MANDATORY_PAKS);
-        final int resourcesListResId = context.getResources().getIdentifier(
-                XWALK_RESOURCES_LIST_RES_NAME, "array", context.getPackageName());
+        int resListResId = context.getResources().getIdentifier(
+                XWALK_RESOURCES_LIST_RES_NAME, "array", context.getClass().getPackage().getName());
+        if (resListResId == 0) {
+            resListResId = context.getResources().getIdentifier(
```
lincsoon: Why get the identifier again? Isn't the first `getIdentifier()` enough?

lincsoon:
Please update the commit log(not just the GitHub comment) as following:
```
[Android] Title

Comments

BUG=XWALK-3569
```

lincsoon:
Please also rebase your branch onto the latest version of the master branch

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
rob-close: If the first call to getIdentifier equals zero, then it is called again with a different parameter.

rob-close:
Rebase has been done.

rakuco:
Thanks, Rob. Please also make sure your commit/pull request messages adhere to our conventions, as Lin Sun mentioned a few comments above.

rob-close:
Hi Raphael, unless I am misunderstanding something, it looks like baleboy has updated the message to include the missing information.

rakuco:
He only added the "BUG=XWALK-3569" part to the pull request message. The commit message is still lacking it, and both need a title that is not the bug number and begins with "[Android]".

rob-close:
Hi Raphael, I think I now have everything changed as requested.  Please let me know if there is anything else.

Thanks,
Rob.

rakuco:
We're almost there :-)

You seem to have run `git pull --merge` and now you have two commits doing the same thing and a merge commit on top. Instead, you should have squashed everything into one commit with the right message. Take a look at `git rebase -i`, and let us know if you have any trouble.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
lincsoon: Yes, I know your intention. But could you tell me what exact situation the first call doesn't covers but the second call covers?

rob-close: I cannot tell you the exact situation.  It seems that different tools do the renaming differently.  See this change from Cordova: https://issues.apache.org/jira/browse/CB-5481
They found that they had to do it both ways to have it work correctly with aapt and gradle builds.  We also have to have it work correctly with Mocana ( https://www.mocana.com/ ).  After seeing what worked for Cordova I took their approach.  I know that it works correctly with the tools just mentioned.

rob-close: Here is a different reference to Cordova's approach: https://issues.apache.org/jira/browse/CB-5080
There they try getClass().getPackage().getName(), and if that does not work they try getPackageName(), which is the pattern I followed.

lincsoon:
LGTM

rakuco:
Thanks for all the help, everyone!

rob-close:
Thanks for your help getting this in.  It would really help if the change could be merged into crosswalk 14 as well.

rakuco:
> Thanks for your help getting this in. It would really help if the change could be merged into crosswalk 14 as well.

You can send a pull request with the backport yourself: create a branch off crosswalk-14, use `git cherry-pick -x <this commit>` and send another pull request targeting that branch.

PR #3023, opened 2015-05-13
DEPS.xwalk: Roll v8-crosswalk(ef0cacb->108d89f)

chuan9:


rakuco:
The commit you are moving to is not present in v8-crosswalk's crosswalk-13/42.0.2311.135 branch. One would first need to cherry-pick it to that branch and only then roll DEPS.xwalk.
Given that new Crosswalk 13 releases are only made on demand, can you first check with QA and @baleboy in XWALK-3155 if this V8 commit needs to make it to Crosswalk 13?

chuan9:
er...ok, sorry for my mistake, I will ask for QA's needs.

PR #3026, opened 2015-05-14
[Linux] Crash when remote debugging with Web UI

PeterWangIntel:
The Web UI doesn't have favicon, so don't need to show it in front-end.

And the reason of crash is that the delegate of 'WebContents' related with 'File Picker' is 'WeDialogView' rather than 'Runtime', so 'static_cast&lt;Runtime*&gt;' in 'GetFaviconDataURL' will cause problem.

With this patch the 'File Picker' can be debugged, like attached figure shows.
remote-debug-web-ui

BUG=https://crosswalk-project.org/jira/browse/XWALK-4185

### DIFF
--- a/runtime/browser/devtools/xwalk_devtools_delegate.cc
+++ b/runtime/browser/devtools/xwalk_devtools_delegate.cc
@@ -35,6 +35,8 @@ const char kTargetTypePage[] = "page";
 const char kTargetTypeServiceWorker[] = "service_worker";
 const char kTargetTypeOther[] = "other";
 
+const char kWebUIScheme[] = "chrome";
+
 class Target : public content::DevToolsTarget {
  public:
   explicit Target(scoped_refptr<content::DevToolsAgentHost> agent_host);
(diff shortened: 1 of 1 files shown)


PeterWangIntel:
@pozdnyakov @halton @joone , please help to review, thank you.

rakuco:
"Crush" should be "Crash".

runtime/browser/devtools/xwalk_devtools_delegate.cc
```diff
@@ -82,25 +84,23 @@ class Target : public content::DevToolsTarget {
 
 Target::Target(scoped_refptr<content::DevToolsAgentHost> agent_host)
     : agent_host_(agent_host) {
+  favicon_url_ = GURL();
```
pozdnyakov: I believe this is not needed, favicon_url_ is already constructed as GURL()

PeterWangIntel: yes, sorry.

pozdnyakov:
lgtm

rakuco:
Please backport if necessary.

PR #3027, opened 2015-05-14
[Android] Fixed unit tests for XWalkViewInternal

lincsoon:
This patch is to fix the unit test errors that introduced by commit
021bfe092f22fd7e266947795f0a8a9541db2e73 "[Android] Provide
asynchronous interface for initializing the Crosswalk library"

Make XWalkViewInternal load native library by self if no external
initializer.

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -250,6 +250,8 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
     private static void init(Context context, Activity activity) {
         if (sInitialized) return;
 
+        XWalkViewDelegate.loadXWalkLibrary(null);
+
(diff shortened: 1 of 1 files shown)


halton:
lgtm, will merge after trybot green.

rakuco:
Is this to fix a regression introduced by 021bfe092f22fd7e266947795f0a8a9541db2e73? If so, can you mention that commit (I usually use the format `commit 1234abcd ("first line of the commit message")`) and indicate what kind of error was happening?

lincsoon:
Yes, it's introduced by 021bfe092f22fd7e266947795f0a8a9541db2e73. I modified the comment

rakuco:
You've only seem to have changed the pull request message, so people looking at the commit history will still be confused. If you're not in a big hurry, it would be good to have this fixed before merging, otherwise please keep that in mind for the future, since as it is the commit does not indicate to people what broke and when.

rakuco:
lgtm, thanks!

PR #3029, opened 2015-05-14
[Android] Fixed visual artifacts at startup in shared mode

lincsoon:
Displaying dialog causes irregular visual artifacts in background if the
default window background is null. For this case, set the background to
screen_background_dark temporarily during the initialization of shared
mode.

BUG=XWALK-4155

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
@@ -49,6 +49,7 @@ public abstract class XWalkActivity extends Activity {
     private XWalkLibraryListener mLibraryListener;
     private Dialog mActiveDialog;
     private boolean mIsXWalkReady;
+    private boolean mDecoratedBackground;
     private String mXWalkApkDownloadUrl;
 
     private static class XWalkLibraryListener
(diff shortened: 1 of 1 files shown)


lincsoon:
@halton @fujunwei @axinging Please help to review

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -177,12 +178,21 @@ private void initXWalkLibrary() {
...
+                getWindow().setBackgroundDrawable(null);
+                mDecoratedBackground = false;
+            }
             XWalkLibraryLoader.startInitialization(mLibraryListener);
             return;
         }
 
         if (mActiveDialog != null) return;
 
+        if (getWindow().getDecorView().getBackground() == null) {
```
rakuco: I think it would be good to add a short comment explaining why this is being done.

halton: I think the message in git log is enough, isn't it?

halton:
LGTM, i'm holding to merging until @rakuco is okay with the message/comment.

rakuco:
lgtm

PR #3030, opened 2015-05-14
[Windows] Fix SysApp CPU load API.

darktears:
Implement getting the CPU load from the performance counters API.
This API needs to be initialized so I've added an init() function
which is used only on Windows (other platforms are not doing anything).

### DIFF
--- a/sysapps/device_capabilities/cpu_info_provider.h
+++ b/sysapps/device_capabilities/cpu_info_provider.h
@@ -30,6 +30,8 @@ class CPUInfoProvider {
   // Android (via /proc/loadavg).
   double GetCPULoad() const;
 
+  void init();
+
   int number_of_processors_;
   std::string processor_architecture_;
 
--- a/sysapps/device_capabilities/cpu_info_provider_android.cc
+++ b/sysapps/device_capabilities/cpu_info_provider_android.cc
@@ -20,6 +20,8 @@ const char kProcLoadavg[] = "/proc/loadavg";
 namespace xwalk {
 namespace sysapps {
 
+void CPUInfoProvider::init() {}
+
 double CPUInfoProvider::GetCPULoad() const {
   // Bionic doesn't have a getloadavg() implementation.
   const base::FilePath proc_loadavg(kProcLoadavg);
--- a/sysapps/device_capabilities/cpu_info_provider_win.cc
+++ b/sysapps/device_capabilities/cpu_info_provider_win.cc
@@ -4,14 +4,30 @@
 
 #include "xwalk/sysapps/device_capabilities/cpu_info_provider.h"
 
+#include <pdh.h>
+
 #include "base/logging.h"
 
 namespace xwalk {
 namespace sysapps {
 
+namespace {
+  PDH_HQUERY cpu_query;
+  PDH_HCOUNTER cpu_total;
+}
+
+void CPUInfoProvider::init() {
+  PdhOpenQuery(nullptr, NULL, &cpu_query);
+  PdhAddCounter(cpu_query,
+                L"\\Processor(_Total)\\% Processor Time", NULL, &cpu_total);
+}
+
 double CPUInfoProvider::GetCPULoad() const {
-  NOTIMPLEMENTED();
-  return 0;
+  PDH_FMT_COUNTERVALUE counter_value;
+  PdhCollectQueryData(cpu_query);
+  PdhGetFormattedCounterValue(cpu_total, PDH_FMT_DOUBLE,
+                              nullptr, &counter_value);
+  return counter_value.doubleValue / 100;
 }
 
 }  // namespace sysapps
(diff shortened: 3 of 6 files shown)


sysapps/device_capabilities/cpu_info_provider.h
```diff
@@ -30,6 +30,8 @@ class CPUInfoProvider {
   // Android (via /proc/loadavg).
   double GetCPULoad() const;
 
+  void init();
```
rakuco: How about making it `virtual` and providing an empty implementation here?

darktears: then each backend needs to subclass this which is not the case today.

rakuco: Ah, I didn't see those were not subclasses.

darktears: :) I would have gone for the virtual otherwise :)

pozdnyakov: I'm afraid that wouldn't work as invoked from the ctor :)

sysapps/device_capabilities/cpu_info_provider_win.cc
```diff
@@ -4,14 +4,27 @@
...
 #include "xwalk/sysapps/device_capabilities/cpu_info_provider.h"
 
+#include <pdh.h>
+
 #include "base/logging.h"
 
 namespace xwalk {
 namespace sysapps {
 
+static PDH_HQUERY cpuQuery;
```
pozdnyakov: nameless namespace?

sysapps/device_capabilities/cpu_info_provider_win.cc
```diff
@@ -4,14 +4,27 @@
...
+static PDH_HCOUNTER cpuTotal;
```
pozdnyakov: ditto

sysapps/device_capabilities/cpu_info_provider_win.cc
```diff
@@ -4,14 +4,27 @@
...
+
+void CPUInfoProvider::init() {
+  PdhOpenQuery(NULL, NULL, &cpuQuery);
```
pozdnyakov: nit: nullptr

sysapps/device_capabilities/cpu_info_provider_android.cc
```diff
@@ -20,6 +20,8 @@ const char kProcLoadavg[] = "/proc/loadavg";
 namespace xwalk {
 namespace sysapps {
 
+void CPUInfoProvider::init() {}
```
pozdnyakov: do we anticipate anything here? maybe just put the stuff to the constructor in the windows-specific class (not introduce init() method)?

sysapps/device_capabilities/cpu_info_provider_win.cc
halton: Do we really needed anonymous? `cpuQuery` is not declared in header. BTW, `cpuQeury` -> `cpu_query`

darktears: Anonymous is apparently the pattern in Chromium even in .cc files. Will rename it though.

sysapps/device_capabilities/cpu_info_provider.h
darktears: Yup.

sysapps/device_capabilities/cpu_info_provider_win.cc
darktears: only some can be replaced, the second NULL here is a DWORDPTR and thus doesn't compile with nullptr. Not sure I want to add a cast here.

sysapps/device_capabilities/cpu_info_provider_android.cc
pozdnyakov: so, would it be simpler just to add functional code to the Win-specific constructor?

darktears: Well there is no Win-specific constructor, the _win file is not a subclass if CPUInfoProvider. So what I can do is ifdef out the constructer in cpu_info_provider.cc only for Windows and implement the CPUInfoProvider constructor in the _win class. The other solution is to put the few init() lines inside cpu_info_provider.cc directly protected by a ifdef only on Windows, but then again what's the point of a _win file in the first place.

sysapps/device_capabilities/cpu_info_provider_win.cc
```diff
@@ -4,14 +4,30 @@
...
 
+#include <pdh.h>
+
 #include "base/logging.h"
 
 namespace xwalk {
 namespace sysapps {
 
+namespace {
+  static PDH_HQUERY cpu_query;
```
pozdnyakov: 'static' can be omitted now.

sysapps/device_capabilities/cpu_info_provider_android.cc
pozdnyakov: ah, I see. I'm OK with 'init()' than

pozdnyakov:
lgtm

PR #3034, opened 2015-05-16
M44 Rebase: Update the Remote Debugging and DevTools as per upstream …

mrunalk:
…changes

The DevToolsHttpHandler has been componentized upstream

https://codereview.chromium.org/1086623002
https://codereview.chromium.org/1089413002
https://codereview.chromium.org/1056343004
https://codereview.chromium.org/1114503002

### DIFF
--- a/runtime/browser/android/xwalk_dev_tools_server.cc
+++ b/runtime/browser/android/xwalk_dev_tools_server.cc
@@ -18,12 +18,11 @@
 #include "base/strings/stringprintf.h"
 #include "base/strings/utf_string_conversions.h"
 #include "blink_upstream_version.h"  // NOLINT
+#include "components/devtools_http_handler/devtools_http_handler.h"
+#include "components/devtools_http_handler/devtools_http_handler_delegate.h"
 #include "content/public/browser/android/devtools_auth.h"
 #include "content/public/browser/devtools_agent_host.h"
-#include "content/public/browser/devtools_http_handler.h"
-#include "content/public/browser/devtools_http_handler_delegate.h"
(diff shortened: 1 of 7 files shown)


halton:
Add @PeterWangIntel in loop since he ever did that area.

mrunalk:
@PeterWangIntel can you please take a look?

PeterWangIntel:
@halton @mrunalk , sorry, I just noticed, will have a look and give comments ASAP.

PeterWangIntel:
@mrunalk @halton ,
I'm still fighting to download the code of Chromium... 

I'm not familiar the standard working process of "Rebase" mentioned here, my personal suggestion is that we can disable devtools temporarily, like pull request #3049 shows.

Once the "M44 Rebase" is finished, I believe I can download the stable code to make 'devtools' working within 1~2 days.

rakuco:
The process is essentially:

1. Make sure your chromium/v8/blink/ozone-wayland checkouts are detached (`git checkout --detach`)
1. Switch to Crosswalk's "next" branch
1. `gclient sync`

The build might break if there is code that still needs to be merged, as is probably the case at the moment.

Given the first days after a rebase are usually a bit tumultuous anyway, I think we could go ahead without disabling devtools: it would be good if you could just take a look at the CLs linked in the pull request message and Mrunal's patch and see if there is anything obvious that was overlooked. If not, we can go ahead and fix any regressions afterwards (or during the rebase cycle once the code is buildable).

mrunalk:
Yeah, I just wanted you to review the code. We can fix the regressions later. Considering the deadline is approaching faster I will go ahead and merge this patch for now.

PR #3037, opened 2015-05-17
[Android] Disable gpu rasterization when using animatable xwalk view

fujunwei:
The TextureView of onSurfaceTextureAvailable can't be triggered before PictureLayerImpl::CalculateTileSize
because of using gpu rasterization, so the tile  width is unavailable to render page.

BUG=XWALK-4012

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkSwitches.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkSwitches.java
@@ -11,6 +11,8 @@ public abstract class XWalkSwitches {
     // Native switch - xwalk_switches::kXWalkProfileName
     public static final String PROFILE_NAME = "profile-name";
 
+    public static final String DISABLE_GPU_RASTERIZATION = "disable-gpu-rasterization";
+
     // Prevent instantiation.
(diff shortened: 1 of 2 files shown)


baleboy:
Thanks @fujunwei ! Do you know what are the implications for performance, does it mean that when using the animatable webview the overall performance will decrease?

fujunwei:
Yes, there are some regression for using the animatable xwalk view.

baleboy:
OK, I guess we'll have to live with this for now in order to make things work again, but let's monitor the performance closely. @lincsoon @darktears @halton please comment.

fujunwei:
Because of expanding gpu raster white list from chromium 42 to 43, so the issue is exposed in this version.
Is it necessary to disable animatable xwalk view by default to decrease the effect and then announce  developers.

lincsoon:
I also agree that we live with this for now. We will consider if we could use SurfaceView by default to avoid the regression.

baleboy:
can we LGTM and merge this then? Also cherry pick to beta please

lincsoon:
I think we can merge this first and put aside the performance issue.

lincsoon:
LGTM

PR #3038, opened 2015-05-18
Use package ID as the application ID when launched from XPK.

seanlong:
XPK file will be unzipped to a temporary directory when launch, the
app ID will change as temporary directory changes. Use package ID can
avoid this issue.


seanlong:
@pozdnyakov  thanks.

pozdnyakov:
lgtm

PR #3039, opened 2015-05-18
Refactor runtime features for different platforms. 

seanlong:
Different platforms have different requirements on Crosswalk components,
after this change each platform can define its specific components
without overriding XWalkRunner::CreateXXXComponents().

Also disable all extensions before we have permission management on Linux
desktop platform.


seanlong:
@pozdnyakov @terriko PTAL.

runtime/browser/xwalk_runner.cc
```diff
@@ -144,6 +144,8 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
   std::vector<extensions::XWalkExtension*> ui_thread_extensions;
   std::vector<extensions::XWalkExtension*> extension_thread_extensions;
 
+// Disable bundled extensions for Linux desktop environment.
```
rakuco: The `!defined(OS_LINUX)` part does not seem to be directly related to disabling the bundling extensions on Linux. Did you forget to provide more context in the commit/pull request message or was this an oversight?

kenchris:
I agree that the getManifest API can be removed for now. I don't think anyone is using it and we didn't discuss it for standardization.

We do need some file system api, it is a very requested feature, but it should need some kind of permission. Btw, how far is this API from the PhoneGap one? and how fast it is in comparison?

runtime/browser/xwalk_runner.cc
```diff
@@ -144,6 +144,8 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
...
+#if defined(OS_TIZEN) || !defined(OS_LINUX)
```
pozdnyakov: not lgtm.

instead of random removal of logic parts it would be enough just to move 'AddComponent' calls from XWalkRunner::CreateComponents() to XWalkRunnerTizen::CreateComponents()

terriko:
The decision was that APIs such as the file system API would need some sort of permissions management system so they could be enabled/disabled (Android, for example, does this with intents, but the architecture of Linux gives greater power to applications).  Since the team is hoping to have the Linux version available on a fairly aggressive deadline, it's easier to disable this on a short-term basis.   Support for the file system api can be revisited when someone has time to build an appropriate system to mitigate the risks involved.

runtime/browser/xwalk_runner.cc
seanlong: ```#if defined(OS_TIZEN) || !defined(OS_LINUX)``` is to enable code for all platforms other than Linux desktop.
Anyway I will change like @pozdnyakov mentioned.

runtime/browser/xwalk_runner.cc
seanlong: Thanks for mention this, I will add a new XWalkRunnerLinux to handle the derivation.

pozdnyakov:
Wouldn't it work if you simply remove all the 'AddComponent' calls from XWalkRunner::CreateComponents() (like I wrote previously)? thus there will be no bundled extensions.

seanlong:
> Wouldn't it work if you simply remove all the 'AddComponent' calls from XWalkRunner::CreateComponents() (like I wrote previously)?

Seems no, we need ApplicationComponent, and as ```app_component_``` is a private member, we can't override XWalkRunner::CreateComponents. An alternative way is add app_component_ setter or make it protected, wdyt?

pozdnyakov:
Sorry I missed the app_component issue. Still I've following thoughts on that: 

1) the only way bundled extensions can pass into is 
```c++
void XWalkRunner::CreateComponents() {
  scoped_ptr<ApplicationComponent> app_component(CreateAppComponent());
  app_component_ = app_component.get(); 
  AddComponent(app_component.Pass());

  if (XWalkRuntimeFeatures::isSysAppsEnabled())
    AddComponent(CreateSysAppsComponent().Pass());
  if (XWalkRuntimeFeatures::isStorageAPIEnabled())
    AddComponent(CreateStorageComponent().Pass());
} 
```

2) assuming both  ``` XWalkRuntimeFeatures::isSysAppsEnabled() ``` and ```XWalkRuntimeFeatures::isStorageAPIEnabled()``` return 'false' we've only ```app_component_```

3) going further, the actual extension adding happens at:
```
void ApplicationSystem::CreateExtensions(
    content::RenderProcessHost* host,
    extensions::XWalkExtensionVector* extensions) {
  Application* application =
    application_service_->GetApplicationByRenderHostID(host->GetID());
  if (!application)
    return;  // We might be in browser mode.

  extensions->push_back(new ApplicationRuntimeExtension(application));
  // Register the widget extension only when the application is widget format.
  if (application->data()->manifest_type() == Manifest::TYPE_WIDGET)
      extensions->push_back(new ApplicationWidgetExtension(application));
}

```
4) we could make ``` ApplicationSystem::CreateExtensions ``` virtual and move the existing functionality to ``` ApplicationSystemTizen ```. WDYT?

In general my point is that we have already enough opportunities to tune extension loading.

seanlong:
Disable these components make more sense than disable these components' extensions. I just followed the same pattern used in XWalkRunnerAndroid.

However currently the XWalkRuntimeFeatures is only determined by command line switches. I think if we add platform-to-feature mapping it would be better to handle this cases.

seanlong:
```
../../xwalk/runtime/common/xwalk_runtime_features.cc:47:1: error: could not convert '{"SysApps", "RawSocketsAPI", "StorageAPI"}' from '<brace-enclosed initializer list>' to 'const std::vector<std::basic_string<char, std::char_traits<char>, std::allocator<char> > >'
 };
 ^
```
Seems Android STL doesn't support braced initialization list?

halton:
@seanlong strange, we should use the same Android NDK version. This error does not appear on your local build?

build/common.gypi
```diff
@@ -4,6 +4,7 @@
     'tizen_mobile%': 0,
     'enable_murphy%': 0,
     'use_webui_file_picker%': 0,
+    'linux_dt%': 0,
```
rakuco: What is this?

seanlong: I think we need to define a build parameter to distinguish default Linux build and LinuxDT.
https://github.com/crosswalk-project/crosswalk/pull/3039/files#diff-a109cb6a8605098d714d08947e753ef2R49 is an example, on the default build experimental features will be enabled, however in a release version(LinuxDT), we want these features disabled.

seanlong:
@pozdnyakov @rakuco @halton PTAL, thanks.

runtime/browser/xwalk_runner_android.cc
```diff
@@ -1,40 +0,0 @@
-// Copyright (c) 2014 Intel Corporation. All rights reserved.
```
pozdnyakov: nice that these files leave :)

runtime/common/xwalk_runtime_features.cc
```diff
@@ -11,20 +11,69 @@
...
+namespace {
+
+// Add new features here with the following fields :
+// - Name of the feature
+// - Name of the command line switch which will be used after the
+// --enable/--disable
+// - Description of the feature
+// - Status of the feature : experimental which is turned off by default or
+// stable which is turned on by default
+using Feature = xwalk::XWalkRuntimeFeatures::Feature;
```
pozdnyakov: using Feature = XWalkRuntimeFeatures::Feature;

runtime/common/xwalk_runtime_features.cc
```diff
@@ -11,20 +11,69 @@
...
+Feature* g_features[] = {
+  new Feature("SysApps", "sysapps",
```
pozdnyakov: why on heap? Feature g_features[] =

runtime/common/xwalk_runtime_features.cc
```diff
@@ -11,20 +11,69 @@
...
-  }
-  const std::string name;
+// Available features on current platform.
+#if defined(OS_ANDROID)
+// FIXME(cmarcelo): The application extensions are currently not fully working
+// for Android, so disable them. See
+// https://crosswalk-project.org/jira/browse/XWALK-674.
+// Android uses a Java extension for device capabilities, so disable the one
+// from sysapps/.
+const char* g_platform_features[] = {
```
pozdnyakov: what's g_platform_features vs. g_features ?

runtime/common/xwalk_runtime_features.cc
```diff
@@ -11,20 +11,69 @@
...
+  "SysApps", "RawSocketsAPI", "StorageAPI", "DialogAPI"
 };
+#elif defined(OS_LINUX_DT)
```
pozdnyakov: don't think we need OS_LINUX_DT, let's just disable bundled features by default (enable for android & tizen)

runtime/common/xwalk_runtime_features.cc
```diff
@@ -11,20 +11,69 @@
...
+namespace xwalk {
+
+XWalkRuntimeFeatures::Feature::Feature(
+    const std::string& name,
+    const std::string& cmdline,
+    const std::string& description,
+    Status status,
+    bool enabled)
+    : name(name),
+      cmdline(cmdline),
```
pozdnyakov: cmd_line

runtime/common/xwalk_runtime_features.cc
```diff
@@ -45,49 +94,34 @@ void XWalkRuntimeFeatures::Initialize(const base::CommandLine* cmd) {
...
-                                 const char* description,
-                                 RuntimeFeatureStatus status) {
-  RuntimeFeature feature;
-  feature.name = name;
-  feature.description = description;
-  feature.command_line_switch = command_line_switch;
-  feature.status = status;
-  feature.enabled = false;
-
+void XWalkRuntimeFeatures::AddFeature(Feature* feature) {
```
pozdnyakov: why heap? AddFeature(const Feature& feature)

rakuco:
I disagree with `linux_dt` in gyp and `OS_LINUX_DT`. At the very least it means we won't be testing one configuration; besides that, Linux FooBarBaz should still be Linux, making a distinction between different flavors will only cause unnecessary confusion. If some behavior needs to be different for one reason or another, it should not be done with ifdefs.

runtime/common/xwalk_runtime_features.cc
seanlong: should be ```AddFeature(Feature& feature)``` but non-const reference argument will fail our cpplint.

runtime/common/xwalk_runtime_features.cc
seanlong: will change, thanks.

seanlong:
@pozdnyakov @rakuco if we disable these extensions by default on Linux, the related tests will lose too. I don't think it a good idea to protect the extensions quality.

If we want to ship binaries with different capabilities is there any better choice than ifdef?
An alternative ifdef is "ENABLE_EXTENSIONS_FOR_TESTS" instead of "OS_LINUX_DT", but it is still ifdef.

For example, on general build Linux, $ xwalk --list-features-flags should show:
```
Available runtime features flags : 
  Command Line Switch                Description                                                   Status       
----------------------------------------------------------------------------------------------------------------
--disable-application-api            JavaScript support for Widget and Manifest APIs               Stable
--disable-raw-sockets                JavaScript support for using TCP and UDP sockets              Stable
--enable-dialog-api                  JavaScript support for dialog APIs                            Experimental
--disable-sysapps                    Master switch for the SysApps category of APIs                Stable
--disable-device-capabilities        JavaScript support for peeking at device capabilities         Stable
--disable-storage                    JavaScript support to file system beyond W3C spec             Stable
```

On LinuxDT version, $xwalk --list-features-flags should show:
```
Available runtime features flags : 
  Command Line Switch                Description                                                   Status       
----------------------------------------------------------------------------------------------------------------
```

pozdnyakov:
maybe instead we can introduce smth like "DISABLE_BUNDLED_EXTENSIONS" @rakuco

seanlong:
@pozdnyakov @rakuco PTAL, thanks.

runtime/common/xwalk_runtime_features.cc
```diff
@@ -45,49 +98,34 @@ void XWalkRuntimeFeatures::Initialize(const base::CommandLine* cmd) {
...
-  AddFeature("StorageAPI", "storage",
-             "JavaScript support to file system beyond W3C spec", Stable);
+
+  std::vector<std::string> platform_features(
+      g_features, g_features + arraysize(g_features));
+  for (auto feature : g_all_features) {
+    if (std::find(platform_features.begin(),
+                  platform_features.end(),
+                  feature.name) != platform_features.end())
+      AddFeature(feature);
```
pozdnyakov: can we just do
#if !defined(DISABLE_BUNDLED_EXTENSIONS)
#if defined(OS_ANDROID)
AddFeature("SysApps", "sysapps",  "Master switch for the SysApps category of APIs", Stable);
....
#else
AddFeature(....)
....
#endif
#endif 

so that we drop g_all_features & g_features.

LGTM otherwise

seanlong: done.

pozdnyakov:
lgtm

PR #3040, opened 2015-05-18
[Android] Fixed visual artifacts at startup in shared mode

lincsoon:
Displaying dialog causes irregular visual artifacts in background if the
default window background is null. For this case, set the background to
screen_background_dark temporarily during the initialization of shared
mode.

BUG=XWALK-4155
(cherry picked from commit d4bd7d8357bc0e7830dffc55a0b901dc279b4bab)


lincsoon:
@rakuco cherry-picked from the master branch

PR #3041, opened 2015-05-18
[Android] Check shared library in onResume instead of onStart

lincsoon:
Originally, the application checks shared library in onStart() in order
to initialize the Crosswalk library after jumped into the install page
and switched back to the application. But on tablets, the install
process is not in full screen, so only onResume() get called after
switched back to the application. In this case, the check method won't
be invoked. This patch is to fix this issue.

BUG=XWALK-3953

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
@@ -149,8 +149,8 @@ public abstract class XWalkActivity extends Activity {
     }
 
     @Override
-    protected void onStart() {
-        super.onStart();
+    protected void onResume() {
+        super.onResume();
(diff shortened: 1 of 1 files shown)


lincsoon:
@rakuco @halton

halton:
According your description in log, this fix will fix non fullscreen on tablet. So how about small size screen?   Fullscreen or not? Will onResume will be always called whatever fullscreen or non-fullscreen?

lincsoon:
On the normal mobile devices, the end-use will navigate to other activity, so onRestart()->onStart()->onResume() are invoked when returns back to current activity. On tablets, only dialog boxes pop up upon current activity, only onResume() is invoked when returns to current activity.

So onResume() is the safest solution.

halton:
Thanks for your explanation, LGTM and will merge. Please remember to backport to beta if needed.

PR #3042, opened 2015-05-19
[Android] Disable gpu rasterization when using animatable xwalk view

fujunwei:
The TextureView of onSurfaceTextureAvailable can't be triggered before PictureLayerImpl::CalculateTileSize
because of using gpu rasterization, so the tile  width is unavailable to render page.
There are some regression for the animatable xwalk view.

BUG=XWALK-4012

(cherry picked from commit a275a49c85e21c37713a3c623a8c0463abeda6b7)


fujunwei:
@halton back port the commit to crosswalk 14, PTAL, thanks.

PR #3043, opened 2015-05-19
Ignore checking out reference chrome build

halton:
The reference chrome builds (win/linux/mac) is not used for Crosswalk, so
simply not checking them out to save gclient sync time.


halton:
@rakuco @darktears please review. The sync of chrome_win is very slow on windows, that is the intent of this PR.

rakuco:
great, lgtm!

PR #3044, opened 2015-05-19
[Android] Check shared library in onResume instead of onStart

lincsoon:
Originally, the application checks shared library in onStart() in order
to initialize the Crosswalk library after jumped into the install page
and switched back to the application. But on tablets, the install
process is not in full screen, so only onResume() get called after
switched back to the application. In this case, the check method won't
be invoked. This patch is to fix this issue.

BUG=XWALK-3953
(cherry picked from commit 66fa3c91148209001246d6b17b77185984d0df74)


lincsoon:
@rakuco check-pick from the master branch

PR #3046, opened 2015-05-20
[Anrdoid] Mark shared library with the version of embedding API

lincsoon:
Replace the naming of "sdk" with "api". The original naming is confusing and hard to be associated with the interface of embedding API.

Shift the version of embedding API to 5, so this will be the first version of shared library.

Also remove the target_arch parameter which is not used anymore.


lincsoon:
@rakuco @baleboy @halton Please help to review. This is a big change for the version of shared mode.

tools/reflection_generator/reflection_generator.py
```diff
@@ -125,11 +123,10 @@ def main(argv):
...
                                  'placed.'))
   option_parser.add_option('--stamp', help='the file to touch on success.')
-  option_parser.add_option('--target-arch', help='Target Architecture')
-  option_parser.add_option('--sdk-version', help='API Version')
-  option_parser.add_option('--min-sdk-version', help='Min API Version')
+  option_parser.add_option('--api-version', help='API Version')
+  option_parser.add_option('--min-api-version', help='Min API Version')
   option_parser.add_option('--verify-xwalk-apk', default=0, type='int',
-      help='Verify Crosswalk library APK before loading')
+      help='Verify crosswalk library APK before loading')
```
rakuco: Crosswalk should always be uppercased.

rakuco:
> This patch is just to replace the naming of "sdk" with "api".

I do see more changes than that in the patch though. Some `target_arch` code got removed with no explanation, and the change to `generate_app_packaging_tool.py` does not look like a renaming. Was it a bug before?

lincsoon:
Sorry, I forgot to update the commit log after made some additional change.

`target_arch` is not used anymore, so I removed it.

The change to `generate_app_packaging_tool.py` is to copy the `API_VERSION` file to the Crosswalk package folder so that the users can identify the version of shared library.

rakuco:
lgtm

rakuco:
in case you weren't already planning to, please backport this, otherwise I guess future backports will be annoying to do.

PR #3047, opened 2015-05-20
Adjust gyp for locale string resource generation.

seanlong:
Let's use gyp 'copies' to do recursive directory copy.

### DIFF
--- a/xwalk.gyp
+++ b/xwalk.gyp
@@ -593,15 +593,13 @@
             'grit_resource_ids': 'resources/resource_ids',
           },
           'includes': [ '../build/grit_action.gypi' ],
+          'outputs': [ '<(grit_out_dir)' ]
         },
       ],
       'copies': [
         {
-          'destination': '<(PRODUCT_DIR)/locales',
-          'files': [
-            '<(grit_out_dir)/en-US.pak',
-            '<(grit_out_dir)/zh-CN.pak'
-          ],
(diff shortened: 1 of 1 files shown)


seanlong:
@rakuco, thanks.

rakuco:
Didn't you say `copies` did not do recursive copies before? Why not just change the existing action instead of creating another target?

seanlong:
I misunderstood the "Gyp doesn't have any built-in looping capability" comment in https://code.google.com/p/chromium/codesearch#chromium/src/chrome/tools/build/repack_locales.py. The "copies" does support copy a directory recursively to another directory.

The "copies" source directory has dependency on grit_action.gypi generated locales, so another target is needed. Like other pak resource targets.

rakuco:
> The "copies" source directory has dependency on grit_action.gypi generated locales, so another target is needed. Like other pak resource targets.

But if you specify the two actions one after the order aren't they going to be run in the order you specify, so that the output one produces is available as input to the other?

seanlong:
PR updated.
Seems the missing bit is : ```+          'outputs': [ '<(grit_out_dir)' ]```, otherwise  the "copies" input dependencies will lose: ```ninja: error: 'gen/xwalk/locales', needed by 'locales', missing and no known rule to make it```.

rakuco:
excellent, thanks!

PR #3050, opened 2015-05-20
[Android] Fix the crash issue when launch cordova.

wuhengzhi:
Reserved object was created, but it's content is empty.
The condition of judgment prevents the initialization for embedded mode.

BUG=XWALK-4257

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
@@ -104,7 +104,7 @@ class XWalkCoreWrapper {
      * This method must be invoked on the UI thread.
      */
     public static void initEmbeddedMode() {
-        if (sInstance != null || sReservedObjects != null) return;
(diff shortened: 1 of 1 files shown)


wuhengzhi:
@lincsoon @axinging PTAL, thanks.

axinging:
LGTM

lincsoon:
`CordovaActivity` inherits from `XWalkActivity`. If the developers use `XWalkActivity`, they must follow the rule that invokes embedding API in `onXWalkReady()`. So the usage described in XWALK-4257 is a wrong way, not a bug.

On the other side, `initEmbeddedMode()` is for the class which doesn't use `XWalkActivity` to keep the backward compatibility with old code. It can't coexist with `XWalkActivity`, so the check you concern is to make sure `initEmbeddedMode()` won't be invoked under `XWalkActivity`. Your modification will violate this intention.

My suggestion is that this commit is not necessary, and we should use `CordovaActivity` with the recommended way just like the default template.

wuhengzhi:
Thanks for the explanation, I will close XWALK-4257.

PR #3051, opened 2015-05-21
[Anrdoid] Mark shared library with the version of embedding API

lincsoon:
This patch replaced the naming of "sdk" with "api". The original
naming is confusing and hard to be associated with the interface
of embedding API.

Shifted the version of embedding API to 5, so this will be the
first version of shared library.

This patch also removed the target_arch parameter which is not
used anymore.
(cherry picked from commit 5f1ca18e16f66440f27048c3872901a2511c28f7)


lincsoon:
@rakuco This is cherry picked from the master branch

PR #3054, opened 2015-05-22
 Make gyp check more generious. tizen is subset of ozone

Bysmyyr:
Ozone-wayland desktop build was broken, because it is not Tizen but needs same settings. 
We should check against ozone if gtk can be used.
 Tizen is only part of platforms that needs ozone.


rakuco:
@PeterWangIntel: fyi ^

rakuco:
lgtm

PR #3055, opened 2015-05-22
Move to Chromium 43.0.2357.65, first stable M43

Bysmyyr:
release. It contains some security fixes.

http://googlechromereleases.blogspot.com/2015/05/stable-channel-update_19.html

Thanks for rakuco for guiding through this


rakuco:
lgtm, let's wait for the bots before merging

PR #3057, opened 2015-05-22
[Windows] Enable 'minimal-ui' display mode

pozdnyakov:
Enable 'minimal-ui' display mode on Windows same way as we have it on Linux desktop.

### DIFF
--- a/runtime/browser/ui/native_app_window_views.cc
+++ b/runtime/browser/ui/native_app_window_views.cc
@@ -24,7 +24,7 @@
 #include "ui/views/window/native_frame_view.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


pozdnyakov:
@kenchris @darktears PTAL

rakuco:
> Also this patch adds a warning if a web app is using deprecated 'app.launch.local_path' manifest key.

Can we do that separately?

pozdnyakov:
>>    Also this patch adds a warning if a web app is using deprecated 'app.launch.local_path' manifest key.

>Can we do that separately?

OK, dropping..

rakuco:
lgtm

PR #3060, opened 2015-05-25
Unused include

darktears:


rakuco:
See the warning in the bots' code style check step: the include is there because we use `base::hash_set` and the style checker complains we do not add one of those includes.

darktears:
Then I need to fix it differently for M44.

PR #3061, opened 2015-05-25
Remove unused includes and silence the warning on the style checker.

darktears:
This actually break the build on Windows because hash_set is not
 in ext/. In addition we do not use it in the file but the style checker
 complains we do (we in fact use the base/ version). Silent the warning
 instead.

### DIFF
--- a/application/browser/application_service.cc
+++ b/application/browser/application_service.cc
@@ -4,11 +4,6 @@
 
 #include "xwalk/application/browser/application_service.h"
 
-#include <ext/hash_set>
-#include <set>
-#include <string>
-#include <vector>
-
 #include "base/files/file_util.h"
 #include "base/strings/utf_string_conversions.h"
 #include "content/public/browser/browser_thread.h"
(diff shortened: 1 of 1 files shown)


rakuco:
Why not just revert the `application_service.cc` part of d37f398f65a8c2a2b37ad38bf889e01a9fd03799?

darktears:
because now Android must use ext/ as per https://codereview.chromium.org/951983002. So at the end of the day the only one now using ext/ is now Windows.

rakuco:
Actually none of the platforms use those headers, as we use the typedefs in `base` instead. I'm almost going for just removing the includes and trying to add some NOLINT stuff to avoid the lint warning.

darktears:
right, I can do that instead.

darktears:
@rakuco what about now?

rakuco:
looks good to me. how about sending it against master so the bots can check it? it shouldn't break anything there either, and I can fix the rebasing conflicts in next tomorrow.

PR #3062, opened 2015-05-25
Remove unused includes in application_service.cc

darktears:
We do not use hash_set in the file (we use base/ version)
but the style checker complains we should include it.
Instead we can silence the warning and remove the include. I also
took the opportunity to remove some other unused includes.

### DIFF
--- a/application/browser/application_service.cc
+++ b/application/browser/application_service.cc
@@ -4,15 +4,6 @@
 
 #include "xwalk/application/browser/application_service.h"
 
-#if defined(OS_MACOSX)
-#include <ext/hash_set>
-#else
-#include <hash_set>
-#endif
-#include <set>
-#include <string>
-#include <vector>
-
 #include "base/files/file_util.h"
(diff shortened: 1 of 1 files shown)


darktears:
@rakuco : yeah?

rakuco:
:green_heart: lgtm

Dunno if the bots will fail because the actual header that handles those hash typedefs is not included though.

rakuco:
The `NOLINT` did not seem to have any effect apparently. Since it's a warning and not an error, we could either just leave it there or look for another way to ignore it (maybe something in `CPPLINT.cfg`).

darktears:
worked locally for me : 

_____ do cpp lint
Done processing application/browser/application_service.cc
Total errors found: 0

darktears:
It was missing the NOLINT :)

rakuco:
@_@

rakuco:
lgtm, feel free to merge

PR #3063, opened 2015-05-26

wuhengzhi:
listener was not called.

BUG=XWALK-3647

(cherry picked from commit 5a929a436b054e02428715b549f25440315a8701)


wuhengzhi:
@lincsoon Please merge it to crosswalk-14, thanks.

PR #3064, opened 2015-05-26

wuhengzhi:
listener was not called.

BUG=XWALK-3647

(cherry picked from commit 5a929a436b054e02428715b549f25440315a8701)


wuhengzhi:
@lincsoon Please merge it to crosswalk-13, thanks.

rakuco:
@baleboy, should this go into crosswalk-13?

wuhengzhi:
In https://crosswalk-project.org/jira/browse/XWALK-3647, the affected version is Crosswalk-11. IMO, it' better to fix this issue on current stable version of crosswalk-13.

baleboy:
should be enough to have it in crosswalk 14

baleboy:
Looks like we are creating a new 13 anyway, so this could be cherry picked after all

PR #3065, opened 2015-05-26
[Tizen] Explicitly depend on the content_browser target.

rakuco:
This is in preparation for M44: `tizen_data_fetcher_shared_memory.h` includes `content/` headers, which in M44 end up depending on `blink_headers`. Since there was no explicit dependency on `content_browser` in the `xwalk_tizen_lib` target, Blink's include paths were not being added to the compiler.


rakuco:
@darktears @mrunalk

darktears:
LGTM

PR #3066, opened 2015-05-27
[Android] Implement the embedding API for setDownloadListener.

wuhengzhi:
This patch is to implement the API for setDownloadListener in XWalkView.
Customer will implement the download listener as needed and override the
default listener in XWalkViewInternal.
Add test case for downloading file.

BUG=XWALK-3958

(cherry picked from commit 8690aeb34a653fd6d49153e1d672db4b44e95de5)


wuhengzhi:
@lincsoon I have verified this patch on local device, please merge it to crosswalk-14. Thanks.

PR #3068, opened 2015-05-28
[Android] Extract the initialization process from XWalkActivity

lincsoon:
Some activities might want to integrate the initialization process but
can't extend XWalkActivity due to its limitations. So we extract this
part from XWalkActivity for other's reuse. This is especially for the
integration into Cordova Plugin.

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkActivity.java
@@ -1,42 +1,21 @@
-// Copyright (c) 2014 Intel Corporation. All rights reserved.
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
 package org.xwalk.core;
 
 import android.app.Activity;
-import android.app.AlertDialog;
-import android.app.Dialog;
-import android.app.DownloadManager;
-import android.app.ProgressDialog;
-import android.content.ActivityNotFoundException;
-import android.content.Context;
-import android.content.DialogInterface;
-import android.content.DialogInterface.OnClickListener;
-import android.content.DialogInterface.OnDismissListener;
-import android.content.DialogInterface.OnShowListener;
-import android.content.Intent;
-import android.content.IntentFilter;
-import android.content.pm.ApplicationInfo;
-import android.content.pm.PackageManager;
-import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.res.Resources;
-import android.net.Uri;
 import android.os.Bundle;
-import android.util.Log;
-
-import org.xwalk.core.XWalkLibraryInterface.DecompressionListener;
-import org.xwalk.core.XWalkLibraryInterface.DownloadListener;
-import org.xwalk.core.XWalkLibraryInterface.InitializationListener;
 
 /**
- * <p><code>XWalkActivity</code> helps to execute all procedures for makeing the Crosswalk Project
- * runtime workable and displays dialogs for interacting with the end-user if necessary. The
+ * <p><code>XWalkActivity</code> helps to execute all procedures for makeing the Crosswalk runtime
+ * workable and displays dialogs for interacting with the end-user if necessary. The
  * activities that hold the {@link XWalkView} object might want to extend <code>XWalkActivity</code>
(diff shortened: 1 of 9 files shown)


lincsoon:
@halton @axinging @fujunwei

lincsoon:
@d0ngjun this patch fixes XWALK-4072 indirectly

halton:
@lincsoon if this PR fixes XWALK-4072, please add it in your git log.

lincsoon:
Removed the bug fixing for XWALK-4072, but added a new method `getCrosswalkRuntime` for the `XWalkInitializer`.

halton:
@d0ngjun please review this PR, i'll reply on your lgtm.

runtime/android/core/src/org/xwalk/core/XWalkInitializer.java
```diff
@@ -54,147 +60,175 @@
...
+ *     public void onXWalkInitCancelled() {
+ *         finish();
  *     }
  * }
  * </pre>
  */
 public class XWalkInitializer {
     /**
-     * Interface used to initialize the Crosswalk environment
+     * Interface used to initialize the Crosswalk Project runtime
```
d0ngjun: rewording: Crosswalk runtime

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -91,99 +71,53 @@
...
 
-        @Override
-        public void onInitializationCompleted() {
-            mXWalkActivity.mIsXWalkReady = true;
-            mXWalkActivity.onXWalkReady();
-        }
+    /**
+     * Returns true if the Crosswalk environment is ready, false otherwise
+     */
+    protected boolean isXWalkReady() {
```
d0ngjun: Seems we do not need this.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -91,99 +71,53 @@
...
-        mLibraryListener = new XWalkLibraryListener(this);
-        XWalkLibraryLoader.prepareToInit();
-        XWalkLibraryLoader.startDecompression(mLibraryListener, this);
+        mLibraryDelegate = new XWalkLibraryDelegate();
     }
 
     @Override
     protected void onResume() {
         super.onResume();
-        if (!mIsXWalkReady && !(mActiveDialog instanceof ProgressDialog)) initXWalkLibrary();
```
d0ngjun: Could move all the code in onResume into onXWalkLibraryPrepared, then mIsXWalkReady is useless.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: The user need to take different actions according the status of the Crosswalk runtime sometimes.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
lincsoon: `mIsXWalkReady` is to give the user an option to check the runtime, not only for internal use.

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -91,99 +70,47 @@
...
+
+        mActivityDelegate = new XWalkActivityDelegate(this, completeCommand, cancelCommand);
     }
 
     @Override
     protected void onResume() {
         super.onResume();
-        if (!mIsXWalkReady && !(mActiveDialog instanceof ProgressDialog)) initXWalkLibrary();
+
+        if (!mIsXWalkReady) mActivityDelegate.initialize();
```
d0ngjun: How about move mActivityDelegate.initialize() into onCreate?
Since we only need to invoke this once.
BTW, I think mIsXWalkReady could move into ActivityDelegate, since it's more related.

lincsoon: mIsXWalkReady will make sure `initialize()` is called only once. We have to check shared mode in `onResume()` after installed the apk

d0ngjun:
LGTM

wuhengzhi:
Maybe this PR will lead XWalkCoreShell to crash since the async initialization for xwalk core, I will verify it.

PR #3071, opened 2015-05-28
[Android] Correctly handle package name change at build time.

rob-close:
Sometimes there is a need to change the package name of an application at
build time.  This is usually done when more than one version of an application
is required.  For example, a free vs a paid version of an application.
An application package name can be changed with aapt, using the
--rename-manifest-package parameter
See this page for further description:
 http://www.piwai.info/renaming-android-manifest-package/

This change has that working correctly.

BUG=XWALK-3569

(cherry picked from commit 8d0511b748b0f9e41cf70f89b612090fdc95de70)


rakuco:
lgtm, thanks!

PR #3074, opened 2015-06-01
[Android] Fix the Cordova package failed issue.

wuhengzhi:
Pack cordova 4.0 failed with cordova canary 15.43.355.0, this patch is
to fix it.
The resources copied from components zip file are invalid 9-patch images,
can't be used for Android studio/Maven, so copy these resources from
Chromium source folder directly to CrossWalk.

BUG=XWALK-4313

### DIFF
--- a/build/android/generate_xwalk_core_library.py
+++ b/build/android/generate_xwalk_core_library.py
@@ -226,6 +226,7 @@ def ReplaceCrunchedImage(project_source, filename, filepath):
   """Replace crunched images with source images.
   """
   search_dir = [
+      'components/web_contents_delegate_android/android/java/res',
       'content/public/android/java/res',
       'ui/android/java/res'
   ]


rakuco:
The change looks sane to me. Is there a way to add a test somewhere to prevent this from happening in the future (for example, if we add a dependency on another java target and need to update that same Pytho file again)?

wuhengzhi:
We should get a method to prevent it. This issue blocks the QA tests, we can merge this patch at first, then file a new issue to track the method. What's your opionion? @rakuco
@fujunwei @lincsoon Please help to review.
Thanks.

rakuco:
ok, but please file a ticket so we don't forget to do this

rakuco:
ping @lincsoon and @fujunwei 
looks ok to me at least, and according to a recent posting in crosswalk-dev we might need this in crosswalk-14 too.

fujunwei:
LGTM, thanks.

lincsoon:
LGTM

baleboy:
Please cherry-pick to Crosswalk 14

PR #3076, opened 2015-06-01
[Windows] Use RuntimeUIDelegateDesktop

pozdnyakov:
So that minimal-ui controls actually work.

### DIFF
--- a/runtime/browser/runtime_ui_delegate.cc
+++ b/runtime/browser/runtime_ui_delegate.cc
@@ -7,7 +7,7 @@
 #include "ui/gfx/image/image.h"
 #include "xwalk/runtime/browser/runtime.h"
 
-#if defined(OS_LINUX) && !defined(OS_TIZEN)
+#if defined(OS_WIN) || (defined(OS_LINUX) && !defined(OS_TIZEN))
 #include "xwalk/runtime/browser/runtime_ui_delegate_desktop.h"
 #endif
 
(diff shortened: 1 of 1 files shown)


pozdnyakov:
@darktears ptal

runtime/browser/runtime_ui_delegate.cc
```diff
@@ -7,7 +7,7 @@
 #include "ui/gfx/image/image.h"
 #include "xwalk/runtime/browser/runtime.h"
 
-#if defined(OS_LINUX) && !defined(OS_TIZEN)
+#if defined(OS_WIN) || defined(OS_LINUX) && !defined(OS_TIZEN)
```
rakuco: Although it might not be required, I think it makes sense for clarity to use extra parentheses here: `defined(OS_WIN) || (defined(OS_LINUX) && !defined(OS_TIZEN))`.

runtime/browser/runtime_ui_delegate.cc
```diff
@@ -41,7 +41,7 @@ NativeAppWindow* RuntimeCreateWindow(
 RuntimeUIDelegate* RuntimeUIDelegate::Create(
     Runtime* runtime,
     const NativeAppWindow::CreateParams& params) {
-#if defined(OS_LINUX) && !defined(OS_TIZEN)
+#if defined(OS_WIN) || defined(OS_LINUX) && !defined(OS_TIZEN)
```
rakuco: Ditto.

darktears:
LGTM.

PR #3077, opened 2015-06-02
[Android] Implement the feature to change cache path.

wuhengzhi:
This patch is to add the function to change cache path by adding switch
to command line.
If the user data directory was specified in command line, use this path
instead of the default.

BUG=XWALK-2030

### DIFF
--- a/runtime/browser/xwalk_browser_context.cc
+++ b/runtime/browser/xwalk_browser_context.cc
@@ -109,8 +109,11 @@ XWalkBrowserContext::CreateZoomLevelDelegate(
 base::FilePath XWalkBrowserContext::GetPath() const {
   base::FilePath result;
 #if defined(OS_ANDROID)
-  CHECK(PathService::Get(base::DIR_ANDROID_APP_DATA, &result));
   base::CommandLine* cmd_line = base::CommandLine::ForCurrentProcess();
+  if (cmd_line->HasSwitch(switches::kUserDataDir))
+    result = cmd_line->GetSwitchValuePath(switches::kUserDataDir);
+  if (result.empty())
+    CHECK(PathService::Get(base::DIR_ANDROID_APP_DATA, &result));
   if (cmd_line->HasSwitch(switches::kXWalkProfileName))
     result = result.Append(
         cmd_line->GetSwitchValuePath(switches::kXWalkProfileName));
--- a/runtime/common/xwalk_switches.cc
+++ b/runtime/common/xwalk_switches.cc
@@ -52,4 +52,8 @@ const char kPpapiFlashPath[] = "ppapi-flash-path";
 const char kPpapiFlashVersion[] = "ppapi-flash-version";
 #endif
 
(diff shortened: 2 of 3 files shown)


wuhengzhi:
@lincsoon @axinging @halton Please take a look, thanks.

runtime/common/xwalk_switches.cc
```diff
@@ -32,6 +32,9 @@ const char kXWalkDataPath[] = "data-path";
 #if defined(OS_ANDROID)
 // Specifies the separated folder to save user data on Android.
 const char kXWalkProfileName[] = "profile-name";
+// Specifies the user data directory, which is where the browser will look for
+// all of its state.
+const char kUserDataDir[] = "user-data-dir";
```
halton: I do not see why this switch is android specific, it should be a general one.

wuhengzhi: Yes, it is general. In this patch, the implementation is only for android, so I wrap it with OS_ANDROID. I will adjust it.

halton: I vote do not wrap inside Android logic.

wuhengzhi: Updated it, thanks.

runtime/common/xwalk_switches.h
```diff
@@ -29,6 +29,8 @@ extern const char kPpapiFlashPath[];
 extern const char kPpapiFlashVersion[];
 #endif
 
+extern const char kUserDataDir[];
```
lincsoon: Isn't this parameter only for Android?

wuhengzhi: From http://peter.sh/experiments/chromium-command-line-switches/, we can see it's a general switch.

runtime/browser/xwalk_browser_context.cc
```diff
@@ -109,11 +109,13 @@ XWalkBrowserContext::CreateZoomLevelDelegate(
...
-  CHECK(PathService::Get(base::DIR_ANDROID_APP_DATA, &result));
   base::CommandLine* cmd_line = base::CommandLine::ForCurrentProcess();
-  if (cmd_line->HasSwitch(switches::kXWalkProfileName))
-    result = result.Append(
-        cmd_line->GetSwitchValuePath(switches::kXWalkProfileName));
+  if (cmd_line->HasSwitch(switches::kUserDataDir))
+    result = cmd_line->GetSwitchValuePath(switches::kUserDataDir);
+  if (result.empty())
+    CHECK(PathService::Get(base::DIR_ANDROID_APP_DATA, &result));
+  result = result.Append(
```
lincsoon: Don't need to check HasSwitch for `kXWalkProfileName`?

wuhengzhi: For android, profile name was set as default. This check can be ignored, but the check will increase the robustness, I will add it.

lincsoon:
LGTM

axinging:
LGTM.

PR #3078, opened 2015-06-02
Fix dangling pointer in xwalk extension code.

darktears:
At the time we reach OnRenderProcessHostClosed the IPC messaging filters are already deleted because the IPC Channel has been closed. We do not need to invalidate the ExtensionServerMessageFilter instance because it's already deleted (and the various overloads of MessageFilter are going to invalidate it) and we should not remove from the current channel because it's been already deleted (removing it will cause its deletion). In fact we do not need to hold a reference of that message filter as it's own by the IPC channel and will be deleted.

BUG=XWALK-4295

### DIFF
--- a/extensions/browser/xwalk_extension_data.cc
+++ b/extensions/browser/xwalk_extension_data.cc
@@ -14,8 +14,7 @@ namespace xwalk {
 namespace extensions {
 
 XWalkExtensionData::XWalkExtensionData()
-    : in_process_message_filter_(NULL),
-      extension_thread_(NULL),
+    : extension_thread_(NULL),
       render_process_host_(NULL) {}
 
 XWalkExtensionData::~XWalkExtensionData() {
--- a/extensions/browser/xwalk_extension_data.h
+++ b/extensions/browser/xwalk_extension_data.h
@@ -35,10 +35,6 @@ class XWalkExtensionData {
     return in_process_ui_thread_server_.get();
   }
 
-  ExtensionServerMessageFilter* in_process_message_filter() {
-    return in_process_message_filter_;
-  }
-
   scoped_ptr<XWalkExtensionProcessHost> extension_process_host() {
     return extension_process_host_.Pass();
   }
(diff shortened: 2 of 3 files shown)


darktears:
@halton PTAL.

halton:
LGTM, @darktears many thanks for your quick fix.

soflare:
It works, but not perfect.
OnFilterRemoved, OnChannelClosing and OnChannelError are all called on IO thread. So there is no need to hold a lock for cleanup. Both the lock and the Invalidate method are redundant.

zliang7:
Actually, the root cause of this bug is using a dangling pointer instead of double free.
The message filter is referenced by a plain C pointer in XWalkExtensionData object. It is freed asynchronously when channel is closed. XWalkExtensionData is not aware of the destruction.
So the trap is still there.

darktears:
@zliang7 sure, that's one way to look at it. XWalkExtensionData and the filter are not connected. The one thing we could do is hold a pointer of XWalkExtensionData inside the filter so it can inform that it was destructed, therefore it would clear the reference inside XWalkExtensionData.

zliang7:
Perhaps we can remove the members for message filter in XWalkExtensionData. It is only used in XWalkExtensionService::OnRenderProcessHostClosed() which causes this bug. I'm not very sure since I have not involved the project for 1.5 years. If we have to keep the reference, a weak pointer may be helpful.

darktears:
@zliang7 PTAL

soflare:
LGTM.
Nit: XWalkExtensionData::lock_ should be useless.

darktears:
Assuming you're talking about the lock_ in ExtensionServerMessageFilter, OnMessageReceived is still using it.

zliang7:
Yes, I mean ExtensionServerMessageFilter, sorry for typo and the wrong account.
OnMessageReceived should be only called on IO thread, the lock can be always acquired. It protects nothing.

zliang7:
May not on the IO thread, but always on same thread.

PR #3082, opened 2015-06-08
[Android] Deliver onCreateInputConnection to XWalkView

lincsoon:
onCreateInputConnection is never called even if override it on
XWalkView. Because the view get callbacked is the descendant
ContentView, but not the XWalkView which the user holds.
So pass the callback to the parent view step by step and finally
it can reach to the XWalkView.

BUG=XWALK-3024
(cherry picked from commit 7cdb86c3ed540b291def245922812ccdca26c475)


lincsoon:
@rakuco cherry picked from the master branch

PR #3089, opened 2015-06-11
[Android] Use SurfaceView as XWalkView's default render backend.

axinging:
SurfaceView provids better performance for application which does not need animating the XWalkView. So we use it as default.

BUG=XWALK-4366

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkPreferencesInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkPreferencesInternal.java
@@ -78,11 +78,11 @@ public class XWalkPreferencesInternal {
 
     /**
      * The key string to enable/disable animatable XWalkViewInternal. Default value is
-     * true.
+     * false.
      *
-     * If this key is set to False, then SurfaceView will be created internally as the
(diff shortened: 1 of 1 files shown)


rakuco:
@baleboy: if this gets in it must be in the release notes so people are aware of possible changes in their apps.

axinging:
@lincsoon @fujunwei @kenchris @darktears @halton

fujunwei:
LGTM, thanks.

halton:
@axinging you need update https://github.com/axinging/crosswalk/blob/xwalk-4366-surfaceviewdefault/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkPreferencesInternal.java#L80 so that javadoc can be updated automatically.

axinging:
Thansk @halton. Updated.

halton:
LGTM, merging.

jonathanewerner:
May i ask when we might we able to expect this in beta / core versions? 
Thank you. :)

rakuco:
This commit was only merged to the master branch (Crosswalk 15), and has been present in canaries ever since.
It will make it into betas once we deem Crosswalk 15 OK to be promoted to the beta channel (our schedule has slipped a bit because of all the summer holidays).

PR #3096, opened 2015-06-16
Backport patch to fix an old skia regression

darktears:
Backport https://codereview.chromium.org/1161063003 which fixes
an old Skia performance regression on popular devices like Samsung
Galaxy S3. The above CL was landed in M45 and therefore will be
part in our next rebase. However this regression is a pressing issue
for some stakeholders so they can move to newer versions of Crosswalk
(as opposed to stick with crosswalk 7 the last version where there
is no performance regression). As we do not fork Skia repository
the proposed solution is to apply a patch when running gclient. That
way it is easier to remove when rebasing to M45. Please note that the
provided patch file has been modified to apply on M44.

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -115,6 +115,16 @@ hooks = [
     ],
     'pattern': '.',
     'name': 'upstream_revision'
+  },
+  {
+    #TODO(darktears): Remove when we move to M45 since the patch will be part of skia
+    'action': [
+      'python',
+      'src/xwalk/tools/patch_skia.py',
+      'src/xwalk/packaging/skia-perf-regression-copy-surface-as-draw.patch'
+    ],
+    'pattern': '.',
+    'name': 'patch_skia_regression'
   }
 ]
 
--- a/packaging/skia-perf-regression-copy-surface-as-draw.patch
+++ b/packaging/skia-perf-regression-copy-surface-as-draw.patch
@@ -1,0 +1,1210 @@
+Author : bsalomon <bsalomon@google.com>
+
+Move copy-surface-as-draw fallback to GrGLGpu.
+
+This patch comes from upstream Skia and is a backport
+to version < M45. As we do not fork Skia we apply
+the patch manually with gclient.
+
+CL : https://codereview.chromium.org/1161063003
+Commit : https://skia.googlesource.com/skia/+/6df86409ca586c3cb34f616f03501bd96181f9e4
+
+diff --git a/src/gpu/GrDrawTarget.cpp b/src/gpu/GrDrawTarget.cpp
+index 549ab37..2424992 100644
+--- a/src/gpu/GrDrawTarget.cpp
++++ b/src/gpu/GrDrawTarget.cpp
+@@ -100,13 +100,10 @@ bool GrDrawTarget::setupDstReadIfNecessary(const GrPipelineBuilder& pipelineBuil
+         return false;
+     }
+     SkIPoint dstPoint = {0, 0};
+-    if (this->copySurface(copy, rt, copyRect, dstPoint)) {
+-        dstCopy->setTexture(copy);
+-        dstCopy->setOffset(copyRect.fLeft, copyRect.fTop);
+-        return true;
+-    } else {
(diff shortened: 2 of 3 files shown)


darktears:
@rakuco @baleboy

DEPS.xwalk
```diff
@@ -115,6 +115,15 @@ hooks = [
     ],
     'pattern': '.',
     'name': 'upstream_revision'
+  },
+  {
+    'action': [
```
rakuco: Can you add a "TODO(darktears): Remove when we move to M45 since the patch is part of Skia there"?

packaging/skia-perf-regression-copy-surface-as-draw.patch
```diff
@@ -0,0 +1,1199 @@
+diff --git a/src/gpu/GrDrawTarget.cpp b/src/gpu/GrDrawTarget.cpp
```
rakuco: Can you include the commit message before the patch? Right now it is totally unclear that it comes from upstream, what the CL number is etc.
Plus you can also mention that this is only necessary for M44.

tools/patch_skia.py
```diff
@@ -0,0 +1,36 @@
...
+"""
+patch_skia.py -- Apply patches to skia when running gclient.
+"""
+
+import os
+import subprocess
+import sys
+
+
+def main(argv=None):
```
rakuco: This block manipulating `argv` isn't necessary, you can just reference `sys.argv` directly.

tools/patch_skia.py
```diff
@@ -0,0 +1,36 @@
...
+  xwalk_dir = os.path.dirname(tools_dir)
+  # tools_dir should be at src/xwalk/tools/patch_skia.py
+  # so src is at tools_dir/../../../
+  src_dir = os.path.dirname(xwalk_dir)
+  root_dir = os.path.dirname(src_dir)
+  third_party_dir = os.path.join(src_dir, 'third_party')
+  skia_dir = os.path.join(third_party_dir, 'skia')
+  patch_dir = os.path.join(root_dir, argv[1])
+
+  output = subprocess.check_call(["git", "reset", "--hard"], cwd=skia_dir)
```
rakuco: I think you're looking for `subprocess.call()`. If `check_call()` fails it will raise an exception.

tools/patch_skia.py
```diff
@@ -0,0 +1,36 @@
...
+  if output != 0 :
```
rakuco: Nit: extra space before ':'.

tools/patch_skia.py
```diff
@@ -0,0 +1,36 @@
...
+    return output
+
+  return subprocess.check_call(["git", "apply", patch_dir], cwd=skia_dir)
```
rakuco: Ditto about `call()` vs `check_call()`.

rakuco:
There's also a JIRA issue about this that you should reference in the PR and in the commit message.

rakuco:
Also, do you know if the people who've submitted this change upstream have filed a merge request for M44?

rakuco:
> There's also a JIRA issue about this that you should reference in the PR and in the commit message.

This one: https://crosswalk-project.org/jira/browse/XWALK-3720

packaging/skia-perf-regression-copy-surface-as-draw.patch
darktears: Sure I can put a comment. I'll try to backport it to beta and stable if it makes sense and if possible.

tools/patch_skia.py
darktears: Ok.

darktears:
@kenchris : there is a long discussion going on construct2 forums since Crosswalk 8-9 about regressions in performance related to Skia when using canvas on some devices. We, Crosswalk team, managed to propose a fix some weeks ago but then Google had another fix in the pipe, which is the CL  i'm linking too. The funky title is because it landed first, then got reverted, then re-landed.

kenchris:
OK thanks for the explanation. rs=me sans potential comments from rajuco

PR #3101, opened 2015-06-22
Add code to build Debian desktop packages.

rakuco:
This work is still experimental, so here be dragons!

The whole patch is inspired by the code in Chromium itself in
`src/chrome/installer/linux` that is used to generate RPMs and Debian
packages.

Consequently, this means that at least for now we are not following
Debian's best practices:
- We do not build from a source tarball; instead, the new gyp target
  `xwalk_installer_linux_debian` creates a package out of the files
  generated by the `xwalk` target.
- To achieve that, we make some fairly low-level calls to `dpkg-deb`
  directly.

We may revisit the way we create packages in the future; for now, this
is the quickest way to start having Debian packages and also saves us
some work if/when we decide to ship NaCl's support files, as we do not
need to care about downloading all the toolchain tarball files ourselves
since it is done by `gclient runhooks'.

TODO items:
- Have a `copyright` file.
- Consider whether we should install into `/usr` directly.
- Possibly switch to a more standard Debian packaging workflow.
- Ship NaCl's support files.

Related bug: XWALK-4374

### DIFF
--- a/tools/installer/common/crosswalk.info
+++ b/tools/installer/common/crosswalk.info
@@ -1,0 +1,27 @@
+# Copyright (c) 2009 The Chromium Authors. All rights reserved.
+# Copyright (c) 2015 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+#
+# This file provides common configuration information for building
+# Crosswalk packages for various platforms.
+
+# Base name of the package.
+PACKAGE="crosswalk"
+
+# Filename of the main executable (for generating launcher scripts, etc.)
+PROGNAME="xwalk"
+
+# Base directory for package installation.
+INSTALLDIR="/opt/crosswalk-project"
+
+# Brief package description.
+SHORTDESC="Runtime for hosted and packaged apps"
+
+# Detailed package description.
+FULLDESC="The Crosswalk Project is an HTML5 application runtime, built on open source foundations, which extends the web platform with new capabilities."
+
+# Package maintainer information.
+MAINTNAME="Crosswalk Project"
+MAINTMAIL="crosswalk-dev@lists.crosswalk-project.org"
+PRODUCTURL="https://crosswalk-project.org/"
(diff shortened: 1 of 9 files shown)


rakuco:
ping @baleboy @kenchris @seanlong

rakuco:
Patch v2 is up: I've reworded `SHORTDESC` according to one of Kenneth's suggestions, and we don't hardcode the word _canary_ in the changelog anymore.

tools/installer/common/crosswalk.info
```diff
@@ -0,0 +1,27 @@
...
+PACKAGE="crosswalk"
+
+# Filename of the main executable (for generating launcher scripts, etc.)
+PROGNAME="xwalk"
+
+# Base directory for package installation.
+INSTALLDIR="/opt/crosswalk-project"
+
+# Brief package description.
+SHORTDESC="Runtime for hosted and packaged apps"
```
seanlong: hosted and packaged "web" apps?

tools/installer/debian/build.sh
```diff
@@ -0,0 +1,222 @@
...
+  set -x
+fi
+set -u
+
+# Create the Debian changelog file needed by dpkg-gencontrol. This just adds a
+# placeholder change, indicating it is the result of an automatic build.
+# TODO(mmoss) Release packages should create something meaningful for a
+# changelog, but simply grabbing the actual 'svn log' is way too verbose. Do we
+# have any type of "significant/visible changes" log that we could use for this?
+gen_changelog() {
```
seanlong: We used to use git log to generate changelog: https://github.com/seanlong/crosswalk-package-tool/blob/master/lib/DebProject.js#L44

seanlong:
Good to see we can use these scripts to generate deb instead of using node scripts.
LGTM with some comments.

rakuco:
Thanks, I have some code here to generate a better changelog that I'll finish writing tomorrow and will land this.

rakuco:
Patch v3 is up: I've added sed+awk voodoo to get the list of commits since the last version bump into the changelog. I'll land after the bots go green.

rakuco:
...and patch v4 makes the changelog generation code work if for any reason one is trying to build a package without an actual git checkout.

PR #3104, opened 2015-06-23
Crosswalk 14 bad performance

matrixreal:
hi
i got bad performance with latest intel xdk and crosswalk 12
also the performance became more bad if i use admob banner
any fix for this problem ?


rakuco:
This is a question, not a patch, so there is no point in sending a pull request. Please use the crosswalk-help mailing list or our JIRA bug tracker.

PR #3108, opened 2015-06-26
Work around Skia performance regression

qkmiao:
A Skia CL (https://codereview.chromium.org/607993002 ) made CrossWalk
regress on some devices such as Galaxy S3. Upstream proposed a CL to fix
the performance regression in
https://codereview.chromium.org/1161063003. Alexis backported upstream's
fix to CrossWalk master in
https://github.com/crosswalk-project/crosswalk/pull/3096. But it's hard
to backport to CrossWalk 14 due to many dependency issues in Skia
repo. Not only the real fix is needed to backport but also other
dependent CLs are needed to backport. So I propose to revert the origin
Skia CL which introduced the regression as a workaroud for CrossWalk 14.

Before the regressed CL, Skia use a fallback draw as the first choice
for surface copy.  In this CL I remove other copy methods before the
fallback draw. This is the same logic as before. Also upstream's fix did
the same thing: https://codereview.chromium.org/1161063003. With this
CL, performance of CrossWalk is back. For a test case (propaganda: a
game from community), it can achieve 55FPS compared to previous
5FPS.

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -105,6 +105,15 @@ hooks = [
     ],
     'pattern': '.',
     'name': 'upstream_revision'
+  },
+  {
+    'action': [
+      'python',
+      'src/xwalk/tools/patch_skia.py',
+      'src/xwalk/packaging/skia-perf-regression-copy-surface-as-draw.patch'
+     ],
+     'pattern': '.',
+     'name': 'patch_skia_regression'
   }
 ]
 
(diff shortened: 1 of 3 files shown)


qkmiao:
@rakuco @baleboy @darktears @baleboy , Please help to look at this.

baleboy:
@ds-hwang PTAL

baleboy:
I'm no owner, but LGTM from my point of view. We badly need this to make people move away from Crosswalk 7

baleboy:
@rakuco this will also need the same integration mechanism that @darktears used for a similar patch in Crosswalk 15. Not sure what is required but I hope you can help with that.

ds-hwang:
lgtm, thx!

rakuco:
Francesco/someone else: it's a good idea to mention this PR in the JIRA ticket about performance issues too.

PR #3109, opened 2015-06-26
Remove fix-speex-checkout.py.

rakuco:
This was necessary for people migrating from M38 to M39; we're at M44
now so there has been enough time for everybody to update their
checkouts (plus gclient itself can probably handle this better
nowadays).


kenchris:
ok, lgtm

PR #3111, opened 2015-06-29
Load libxwalkcore library by loadXWalkLibrary.

d0ngjun:
For supporting lzma in make_apk.py,
we can not rely on LibraryLoader to load libxwalkcore,
but load it explictly, no matter lzma is enable or not.

BUG=XWALK-4424


lincsoon:
LGTM

PR #3112, opened 2015-06-30
[Android] Export XWalkCookieManager

fujunwei:
Crosswalk cordova plugin need this class to provide ability of accessing cookie in shared mode. Change the name of XWalkCookieManager to XWalkCookieManagerInternal for generating wrapper and bridge of the class.

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
@@ -7,18 +7,20 @@ package org.xwalk.core.internal;
 import org.chromium.base.JNINamespace;
 
 /**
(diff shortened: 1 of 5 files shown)


fujunwei:
@lincsoon @axinging @halton PTAL, thanks.

axinging:
Several customers use XWalkCookieManager directly. It seems this API not changed, so they can keep using XWalkCookieManager?

About the CookieManagerTest.java, I think use XWalkCookieManager will be better.

fujunwei:
The XWalkCookieManager will be generated by the script of python that will be packaged in jar that will be used by customer.
Use  XWalkCookieManagerInternal instead of XWalkCookieManager  in internal core library.

lincsoon:
LGTM

halton:
LGTM, merging.

PR #3115, opened 2015-06-30
[Windows] Fix patch_skia.py.

rakuco:
Calling "git" on Windows does not work, as the only executable that is
guaranteed to be in `%%PATH%%` because of depot_tools is `git.bat`.

Do the same thing some depot_tools scripts do and prefer the latter when
on Windows.


rakuco:
ping @kenchris @darktears

PR #3128, opened 2015-07-09
Add a swtich to exclude code of devtools

PeterWangIntel:
Build with flag '-Ddisable_devtools=1' to turn on the switch.

### DIFF
--- a/runtime/browser/xwalk_runner.cc
+++ b/runtime/browser/xwalk_runner.cc
@@ -15,7 +15,9 @@
 #include "xwalk/extensions/browser/xwalk_extension_service.h"
 #include "xwalk/extensions/common/xwalk_extension_switches.h"
 #include "xwalk/runtime/browser/application_component.h"
+#ifndef DISABLE_DEVTOOLS
 #include "xwalk/runtime/browser/devtools/remote_debugging_server.h"
+#endif
 #include "xwalk/runtime/browser/storage_component.h"
 #include "xwalk/runtime/browser/sysapps_component.h"
 #include "xwalk/runtime/browser/xwalk_app_extension_bridge.h"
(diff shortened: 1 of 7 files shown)


PeterWangIntel:
@d0ngjun , thanks

runtime/browser/xwalk_runner.cc
```diff
@@ -174,6 +176,9 @@ void XWalkRunner::OnRenderProcessHostGone(content::RenderProcessHost* host) {
 }
 
 void XWalkRunner::EnableRemoteDebugging(int port) {
+#ifdef DISABLE_DEVTOOLS
+  (void) port;
+#else
```
d0ngjun: remove the above two lines.

#else -> #ifndef DISABLE_DEVTOOLS.

xwalk.gyp
```diff
@@ -319,6 +313,19 @@
         },
       },
       'conditions': [
+        ['disable_devtools!=1', {
+          'dependencies': [
+            '../components/components.gyp:devtools_http_handler',
```
d0ngjun: why add a dependency for disable_devtools=0?

xwalk.gyp
```diff
@@ -616,7 +623,7 @@
         'xwalk_resources',
       ],
       'conditions': [
-        [ 'OS!="android"', {
+        [ 'OS!="android"' and 'disable_devtools!=1', {
```
d0ngjun: Seems we don't need to take this 'not android' into consideration.

xwalk.gyp
```diff
@@ -646,7 +653,7 @@
             ],
           },
           'conditions': [
-            [ 'OS!="android"', {
+            [ 'OS!="android"' and 'disable_devtools!=1', {
```
d0ngjun: same as above.

d0ngjun:
Just a reminder:
Don't forget add reducing size info into git log.

runtime/browser/xwalk_runner.cc
PeterWangIntel: yes. thanks.

xwalk.gyp
PeterWangIntel: Means, we only build it when devtools sre not enabled.

xwalk.gyp
PeterWangIntel: We need this dependency only when devtools enabled.

PeterWangIntel:
@d0ngjun @halton

d0ngjun:
LGTM

PR #3130, opened 2015-07-10
[Windows][Extensions]Fix render process crash when sending data

darktears:
from native to JS using the shared memory handle.

This is the second fix for this bug. The previous fix uncovered this
one.

The crash was because the memory handle we send to the extension process
was null. The fact it was null is because on Windows it was failing to
share the handle of the shared memory between the render process and
the extension process. GiveReadOnlyToProcess was failing inside
windows specific code, more exactly in DuplicateHandle, a win API
which was returning an error code of 6. 6 means the handle we're passing
is invalid. Turns out that while we store the renderer process handle
earlier (so we can share the data between the two), the open() call
on the process is too early. It seems like Windows can't work with this
handle unless the process is open right before. Moving the open() call
to the last moment (by storing the PID instead of the handle) doesn't
fully work (we don't fail with error 6 but with 5 which means there
is not enough privileges to open/share with that process). We therefore
open it with extra privileges which means we can share data between
the two. I also added a LOG and an early return when sharing the handle
doesn't work, after all it means it's functionally broken.

While investigating this issue it turns out that the automatic test was
broken because it was always reporting "Pass" even if the actual message
passing with data handle was broken. The fix consists of moving it from
a regular extension test to a true external extension test. The second
part is also making sure we don't report "Pass" until we received all
the messages back from the native extension. If we don't receive one
the test will time out. I've decided to address it in this patch so that
I don't need to fix the test, then disable it to land that patch and
re-enable it again.

It turns out that the fix made the test to fail on Linux (running the test
into a regular xwalk instance was working fine) because the process
couldn't be opened. After debbuging it turns out that the test was
run in a in mode where the extensions are not created the same way as
an external extension (created as in process). This will make our code
for sharing memory to fail because there is no multi-process involved.
The test never worked but the code worked, it was discovered together
with this patch.

### DIFF
--- a/extensions/common/xwalk_extension_server.cc
+++ b/extensions/common/xwalk_extension_server.cc
@@ -25,7 +25,7 @@ const size_t kInlineMessageMaxSize = 256 * 1024;
 
 XWalkExtensionServer::XWalkExtensionServer()
     : sender_(NULL),
-      renderer_process_handle_(base::kNullProcessHandle),
+      renderer_process_pid_(0),
       permissions_delegate_(NULL) {}
 
 XWalkExtensionServer::~XWalkExtensionServer() {
@@ -54,9 +54,7 @@ bool XWalkExtensionServer::OnMessageReceived(const IPC::Message& message) {
 }
 
 void XWalkExtensionServer::OnChannelConnected(int32 peer_pid) {
-  base::Process process = base::Process::Open(peer_pid);
-  renderer_process_handle_ = process.Handle();
-  CHECK(process.IsValid());
+  renderer_process_pid_ = peer_pid;
 }
 
 void XWalkExtensionServer::OnCreateInstance(int64_t instance_id,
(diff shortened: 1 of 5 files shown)


darktears:
@jondong @huningxin please have a look and test that it works fine for you (also that's it's functionally working).

darktears:
Seems like opening the process after fails on Linux. Looks like I need to debug that on an actual Linux machine :(.

huningxin:
I verified it fixes the crash issue on Windows.

jondong:
@darktears I've run the bulkDataExtension test on my Linux and it failed with this patch. It seems the process is not correctly opened with:
base::Process::OpenWithExtraPrivileges(renderer_process_pid_);

darktears:
@jondong this test BulkDataExtension is completely buggy. 

I figured out what was wrong on Linux, in fact the test always passed in the past because the old code was not testing the handle before passing it for sharing. In the test suite it turns out that it runs in single-process which in that case doesn't work properly. A quick debug without the patch on renderer_process_handle_ run while running the test suite shows that it is 0 meaning the call to GiveReadOnlyToProcess fail miserably and as we don't test the return value, it fails to share the memory and thus it supposed to not work, yet the test shows "Pass". When running separately from the test suite (like ningxin mentionned) renderer_process_handle_ gets a correct value and thus the code work (so does with my patch in fact). In that scenario where it's working (both with and without my patch) I got curious and commented the call to Send() where we pass the handle. In both cases the test shows PASS which it shouldn't, after all I'm not passing the message to JS!!! Clearly BulkDataExtension needs love.

What I haven't figured out is why Linux gets confused with the ExtensionServers in single-process mode and the object instance where OnChannelConnected is called is not the same instance where PostMessageToJSCallback is called, something that is not happening on Windows and it also not happening in multi-process mode where OnChannelConnected and PostMessagetoJSCallback are called on the same instance.

darktears:
In fact the difference between when running inside the test and when running like ningxin mentionned is that in the former extensions are registered as in process in the UI thread whereas on the other case those are external extensions.

darktears:
@huningxin @jondong PTAL

jondong:
LGTM, thanks for your effort @darktears !

huningxin:
LGTM. Thanks for the fixing!

PR #3131, opened 2015-07-14
[RemoteDebugging] To fix EnableRemoteDebugging bug

myroot:
To fix EnableRemoteDebugging bug. 

When EnableRemoteDebugging was called with same port number, 
remote_debugging_server was stopped, because it was deleted by reset()

It does not seem to intended code. 
If already remote_debugging_server was created with same port, 
It should just return


myroot:
Abandon this commit, 
The reason of delete the remote_debugging_server was to prevent port already in use error

PR #3132, opened 2015-07-14
Roll Chromium 44.0.2403.81.

rakuco:
This is the latest Linux beta release so far. Thanks to Olli Raula for
starting the work.

FFMPEG had to be temporarily rolled forward so that we can continue
building it as a shared library for the time being.
See https://codereview.chromium.org/1141703002 and
https://codereview.chromium.org/1198653005/ for more information.

A few commits in chromium-crosswalk were also dropped because of this
change, as the code which loaded additional ffmpeg libraries via
--proprietary-codec-lib-path does not work anymore. Users wishing to
provide a different ffmpeg version with additional codecs now need to
replace the version built by Crosswalk instead of adding it alongside
the existing library.

Related to: XWALK-3705
Related to: XWALK-4574

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,8 +17,8 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = 'aa27a06ed316c2a8a77ef256acc7c367de781fdf'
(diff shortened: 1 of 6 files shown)


darktears:
LGTM

PR #3137, opened 2015-07-16
[Android] Fix the test case crash issue on runtime client.

wuhengzhi:
The test case was executed when xwalk was not ready, so the runtime view
sometimes was null. This patch is to wait for the xwalk ready then execute
test case.

BUG=XWALK-4624

(cherry picked from commit b8e1bf1c579eaf612468d26c2351109d3ec6fde2)


wuhengzhi:
@rakuco Please merge this patch to crosswalk-14, thanks.

rakuco:
lgtm, thanks

PR #3146, opened 2015-07-20
Preserve local storage when migrating from webview to Crosswalk

fujunwei:
Copy the data of web storage directory to Crosswalk profile name for
Android Kitkat.
The KK webview and the pre-KK webview use different formate to store data,
but the storage database have implemented the mechanism that it will be upgraded
to V2(BLOB value column) if there was V1 (TEXT value column) database already.

BUG=XWALK-4349

### DIFF
--- a/runtime/browser/xwalk_browser_main_parts_android.cc
+++ b/runtime/browser/xwalk_browser_main_parts_android.cc
@@ -40,6 +40,9 @@
 
 namespace {
 
+const char kPreKitkatDataDirectory[] = "app_database";
+const char kKitkatDataDirectory[] = "app_webview";
+
 base::StringPiece PlatformResourceProvider(int key) {
   if (key == IDR_DIR_HEADER_HTML) {
     base::StringPiece html_data =
@@ -50,12 +53,9 @@ base::StringPiece PlatformResourceProvider(int key) {
   return base::StringPiece();
 }
 
-void MoveUserDataDirIfNecessary(const base::FilePath& user_data_dir,
-                                const base::FilePath& profile) {
-  if (base::DirectoryExists(profile))
-    return;
-
-  if (!base::CreateDirectory(profile))
+void ImportKitkatDataIfNecessary(const base::FilePath& old_data_dir,
+                                 const base::FilePath& profile) {
+  if (!base::DirectoryExists(old_data_dir))
     return;
 
   const char* possible_data_dir_names[] = {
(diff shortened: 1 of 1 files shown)


runtime/browser/xwalk_browser_main_parts_android.cc
```diff
@@ -50,12 +53,9 @@ base::StringPiece PlatformResourceProvider(int key) {
   return base::StringPiece();
 }
 
-void MoveUserDataDirIfNecessary(const base::FilePath& user_data_dir,
+void ImportKitkatDataIfNecessary(const base::FilePath& old_data_dir,
                                 const base::FilePath& profile) {
```
d0ngjun: indent

runtime/browser/xwalk_browser_main_parts_android.cc
```diff
@@ -76,6 +76,40 @@ void MoveUserDataDirIfNecessary(const base::FilePath& user_data_dir,
...
+                                    const base::FilePath& profile) {
+  if (!base::DirectoryExists(old_data_dir))
+    return;
+
+  // Local Storage
+  base::FilePath local_storage_path = old_data_dir.Append("localstorage");
+  if (base::PathExists(local_storage_path)) {
+    if (!base::Move(local_storage_path, profile.Append("Local Storage"))) {
+        NOTREACHED() << "Failed to import previous user data: localstorage";
+      }
```
d0ngjun: indent

fujunwei: Thanks, Done

fujunwei:
@huningxin @lincsoon @halton @d0ngjun , PTAL.

wuhengzhi:
In the comment, ```formate``` -> ```formats```.

d0ngjun:
LGTM

halton:
LGTM

rakuco:
Is there a way for QA to test this? I'd like to have this backported only if we're sure we're not losing or damaging existing user data.

fujunwei:
The test case is  https://github.com/crosswalk-project/crosswalk-test-suite/tree/master/usecase/usecase-wrt-android-tests/samples/ApplicationlocalStorage.  QA will test it.

rakuco:
The test cases are only for localstorage in general, right? I was asking about testing these transitions between different Crosswalk versions and the webview.

PR #3147, opened 2015-07-22
[Android] Enable widevine support

qjia7:
Currently, crosswalk only support clearkey CDM, but doesn't support
widevine CDM. This PR will enable it in crosswalk for android.

The spec can refer https://w3c.github.io/encrypted-media/.

The steps to verify this feature:
Add "enable_widevine=1" to the GYP_DEFINES. Then run gyp and build.
The following link is for DRM testing. If 'com.widevine.alpha' keySystem
is not supported, the below video won't play.
http://shaka-player-demo.appspot.com/?dash;asset=assets/car_segmenttemplate.mpd

### DIFF
--- a/runtime/DEPS
+++ b/runtime/DEPS
@@ -19,5 +19,7 @@ include_rules = [
   "+jni",
   "+components/web_contents_delegate_android",
   "+components/navigation_interception",
+  "+components/cdm/browser",
+  "+components/cdm/renderer",
   "+cc/base/switches.h"
 ]
--- a/runtime/browser/xwalk_content_browser_client.cc
+++ b/runtime/browser/xwalk_content_browser_client.cc
@@ -55,6 +55,7 @@
 #include "base/android/locale_utils.h"
 #include "base/android/path_utils.h"
 #include "base/base_paths_android.h"
+#include "components/cdm/browser/cdm_message_filter_android.h"
 #include "xwalk/runtime/browser/android/xwalk_cookie_access_policy.h"
 #include "xwalk/runtime/browser/android/xwalk_contents_client_bridge.h"
 #include "xwalk/runtime/browser/android/xwalk_web_contents_view_delegate.h"
@@ -202,6 +203,9 @@ void XWalkContentBrowserClient::RenderProcessWillLaunch(
 #endif
   xwalk_runner_->OnRenderProcessWillLaunch(host);
   host->AddFilter(new XWalkRenderMessageFilter);
+#if defined(OS_ANDROID)
+  host->AddFilter(new cdm::CdmMessageFilterAndroid());
+#endif
 }
 
 content::MediaObserver* XWalkContentBrowserClient::GetMediaObserver() {
--- a/runtime/browser/xwalk_permission_manager.cc
+++ b/runtime/browser/xwalk_permission_manager.cc
@@ -55,9 +55,11 @@ void XWalkPermissionManager::RequestPermission(
       callback.Run(content::PERMISSION_STATUS_DENIED);
 #endif
       break;
+    case content::PermissionType::PROTECTED_MEDIA_IDENTIFIER:
+      callback.Run(content::PERMISSION_STATUS_GRANTED);
+      break;
     case content::PermissionType::MIDI_SYSEX:
     case content::PermissionType::NOTIFICATIONS:
-    case content::PermissionType::PROTECTED_MEDIA_IDENTIFIER:
     case content::PermissionType::PUSH_MESSAGING:
       NOTIMPLEMENTED() << "RequestPermission is not implemented for "
                        << static_cast<int>(permission);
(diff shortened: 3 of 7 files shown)


qjia7:
@huningxin Please take a look.

qjia7:
@huningxin The trybot error has been fixed. PTAL.

halton:
@qjia7 have you both check turn on/off enable_widevine? From your change, I guess there should be problem when enable_widevine==0.

And your code use defined(OS_ANDROID) protection? Isn't Cdm corss-platform support?

I know nothhing about widevine/cdm, please forgive me if I asked stupid questions.

qjia7:
>have you both check turn on/off enable_widevine? From your change, I guess there should be problem when enable_widevine==0.

@halton I only checked when enable_widevine=1, whether widevine CDM is supported. The result is even we turn on enable_widevine, the test video still can't be played. So I investigated the reason and send this PR. After adding this patch, the test video can be played well. But I haven't checked to turn off enable_widevine, But I can give a try and provide the result. I think enable_widevine will add some classes and macro. If we disable it, maybe some functions will be blocked which will result in playing failure.  

>And your code use defined(OS_ANDROID) protection? Isn't Cdm corss-platform support?

cdm is cross-platform. But for this patch, I only enabled widevine for android support (I think this is high priority). I can enable other platforms in following PR. How do you think?

halton:
@qjia7 , you need ensure the build fine with enable_widevine on/off. And please decide whether to turn this flag on by default.

Secondly, I know you only tested on Android. But the code seems cross-platform, you suppose benefit upstream already, no need to explicitly add protection for Android unless we really want to support Android only. 

Add @darktears  in cc.

qjia7:
@halton 
> I know you only tested on Android. But the code seems cross-platform, you suppose benefit upstream already, no need to explicitly add protection for Android unless we really want to support Android only.

Chromium have supported widevine. Crosswalk have some issues to directly support it since it depends on  content level. But there are some implementation is in chrome level. 
This PR is only for android not for linux, windows.  There is no error to explicitly add protection for android because those classes are only for android. If we want to support other platforms, we need to do like below:
```
void AddChromeKeySystems(std::vector<KeySystemInfo>* key_systems_info) {
#if defined(ENABLE_PEPPER_CDMS)
  AddExternalClearKey(key_systems_info);

#if defined(WIDEVINE_CDM_AVAILABLE)
  AddPepperBasedWidevine(key_systems_info);
#endif  // defined(WIDEVINE_CDM_AVAILABLE)
#endif  // defined(ENABLE_PEPPER_CDMS)

#if defined(OS_ANDROID)
  cdm::AddAndroidWidevine(key_systems_info);
#endif  // defined(OS_ANDROID)
}

qjia7:
@halton @darktears Any feedback about current implementation?

halton:
LGTM, please rebase your code to fix the merge conflict. Ping @darktears

qjia7:
@halton The patch has been updated. Thanks.

qjia7:
@kenchris Please take a look. Seems that @darktears is on vacation.

darktears:
LGTM

rakuco:
Does one need to ship additional files somewhere if `enable_widevine==1`? I've always been under the impression that some proprietary blob is needed at some point.

qjia7:
>Does one need to ship additional files somewhere if enable_widevine==1?

I think it doesn't need. But I am not sure.  In fact, I am not very clear about your question. Could you please tell me how to verify it in my developing machine? I can check it:)

rakuco:
First of all: I may be asking some stupid questions, but I really don't know much about what this Widevine CDM is supposed to do and how.
With that said, one check would be to launch that URL in the pull request message using Crosswalk on Android and check if it works out of the box. If it does not, it means additional files need to be shipped somewhere.

qjia7:
@rakuco  It should not need additional resource. Apply this patch and build it. I can directly use src/out/Debug/xwalk_app_template.tar.gz alone. Use <code>make_apk.py</code> in that zip file to build a widevine application which will run  http://shaka-player-demo.appspot.com/?dash;asset=assets/car_segmenttemplate.mpd well.

BTW: Widevine cdm is use to decrypt an encrypted html5 video. So you see the spec https://w3c.github.io/encrypted-media/ for the encrypted-media. And some concepts of Key Systems.

For android, widevine cdm will use android sdk API <code>android.media.MediaDrm</code>
For windows, widevine cdm will use google's <code>widevinecdm.dll</code>

halton:
@rakuco okay to merge with @qjia7 's explanation?

rakuco:
OK, so on Android nothing's special is needed, whereas on other platforms a proprietary blob is needed.

js0701:
a question: How about Playready CDM

qjia7:
> a question: How about Playready CDM

Currently, it is not supported. We need to evaluate the status of upstream and consider the next step.

PR #3151, opened 2015-07-23
Fix crash issue when app control event was invoked

myroot:
LaunchFromAppID could returning the NULL pointer.
but it does not checked before used

### DIFF
--- a/application/browser/application_system_tizen.cc
+++ b/application/browser/application_system_tizen.cc
@@ -118,8 +118,10 @@ void application_event_cb(app_event event, void* data, bundle* b) {
 
       Application* app =
           app_service_tizen->LaunchFromAppID(app_id, encoded_bundle);
-      LOG(INFO) << "Application launched with id: " << app->id();
-      handler_data->current_app = static_cast<ApplicationTizen*>(app);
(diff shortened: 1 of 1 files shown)


rakuco:
Thanks for the patch, @myroot. Can you explain under which circumstances you're seeing this behavior (`app` being `NULL`)? Before adding this check, I'd like to understand the situation and make sure we're not just treating a symptom of the problem instead of the root cause.

myroot:
Hello, 
Thanks for your attention. 

First, It was tested in Tizen 3.0 Platform.

The application_system_tizen.cc was using tizen appfw API. 

AE_RESET event was called when the another application was sending a event into this application. 
So, AE_RESET event could be called more than once. 

When AE_RESET event was called, application_system_tizen.cc was use ApplicationServiceTizen::LaunchFromAppID and ApplicationService::Launch method. 

ApplicationService::Launch was returning the application handle. 
But, If Application was already launched, It just returning NULL value. 
and application_system_tizen was use the returning value without NULL checking.

rakuco:
OK, thanks for the explanation. Merging.

PR #3155, opened 2015-07-24
Add XW_MessagingInterface_2

huningxin:
XW_MessagingInterface_2 adds the binary message support.

BUG=XWALK-4615
TEST=xwalk_extensions_browsertest
     --gtest_filter=ExternalExtensionTest.ExternalExtensionMessaging2

### DIFF
--- a/extensions/common/xwalk_external_instance.cc
+++ b/extensions/common/xwalk_external_instance.cc
@@ -34,18 +34,25 @@ XWalkExternalInstance::~XWalkExternalInstance() {
 
 void XWalkExternalInstance::HandleMessage(scoped_ptr<base::Value> msg) {
   XW_HandleMessageCallback callback = extension_->handle_msg_callback_;
-  if (!callback) {
+  XW_HandleBinaryMessageCallback binary_callback =
+      extension_->handle_binary_msg_callback_;
+  if (!callback && !binary_callback) {
     LOG(WARNING) << "Ignoring message sent for external extension '"
                  << extension_->name() << "' which doesn't support it.";
     return;
   }
 
   std::string string_msg;
-  if (!msg->GetAsString(&string_msg)) {
+  const base::BinaryValue* binary_msg = nullptr;
+  if (callback && msg->GetAsString(&string_msg)) {
+    callback(xw_instance_, string_msg.c_str());
+  } else if (binary_callback && msg->GetAsBinary(&binary_msg)) {
+    binary_callback(xw_instance_, binary_msg->GetBuffer(),
+                    binary_msg->GetSize());
+  } else {
     LOG(WARNING) << "Failed to retrieve the message's value.";
-    return;
   }
-  callback(xw_instance_, string_msg.c_str());
+  return;
 }
 
 void XWalkExternalInstance::HandleSyncMessage(scoped_ptr<base::Value> msg) {
@@ -75,6 +82,12 @@ void* XWalkExternalInstance::CoreGetInstanceData() {
 
 void XWalkExternalInstance::MessagingPostMessage(const char* msg) {
   PostMessageToJS(scoped_ptr<base::Value>(new base::StringValue(msg)));
+}
+
+void XWalkExternalInstance::MessagingPostBinaryMessage(const char* msg,
+                                                       const size_t size) {
+  PostMessageToJS(scoped_ptr<base::Value>(
+      base::BinaryValue::CreateWithCopiedBuffer(msg, size)));
 }
 
 void XWalkExternalInstance::SyncMessagingSetSyncReply(const char* reply) {
--- a/extensions/common/xwalk_external_adapter.cc
+++ b/extensions/common/xwalk_external_adapter.cc
@@ -79,6 +79,16 @@ const void* XWalkExternalAdapter::GetInterface(const char* name) {
     return &messagingInterface1;
   }
 
+  if (!strcmp(name, XW_MESSAGING_INTERFACE_2)) {
+    static const XW_MessagingInterface_2 messagingInterface2 = {
+      MessagingRegister,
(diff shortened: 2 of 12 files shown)


huningxin:
@halton @zliang7 @darktears please take a look! Thanks!

huningxin:
ping @halton @zliang7 @darktears , Thanks!

halton:
LGTM, ping @darktears . 

@huningxin I suppose you should have Android part in separate PR?

huningxin:
> @huningxin I suppose you should have Android part in separate PR?

Exactly. Android implementation depends on common part of this PR.

extensions/common/xwalk_external_instance.cc
```diff
@@ -34,18 +34,25 @@ XWalkExternalInstance::~XWalkExternalInstance() {
...
   }
 
   std::string string_msg;
-  if (!msg->GetAsString(&string_msg)) {
+  const base::BinaryValue* binary_msg = nullptr;
+  if (callback && msg->GetAsString(&string_msg)) {
+    callback(xw_instance_, string_msg.c_str());
+  } else if (binary_callback && msg->GetAsBinary(&binary_msg)) {
+    binary_callback(xw_instance_, binary_msg->GetBuffer(),
+                    binary_msg->GetSize());
```
zliang7: This means extension can not use string and binary message simultaneously, and string message is preferred. Is this by design?
I think we should deprecate the old string message interface, and remove it in some day hopefully.
I'm not sure what the owners and architects think about this. I'd better sync with them.

zliang7: If this is by design, it's better to raise an assertion/exception while extensions with string message callback try to register a binary message callback.

huningxin: > This means extension can not use string and binary message simultaneously

This is not true. Extension can use string and binary message at same time. Please see tests for details.



> I think we should deprecate the old string message interface, and remove it in some day hopefully.

I think they are complementary. String message is easy for coding with JSON. Binary message is for binary data transmission. Similar to websocket API, it provides both string and binary message interface.

zliang7: Make sense. Please ignore my previous comments, LGTM

huningxin: Thanks for the review.

huningxin:
@halton @zliang7 , thanks for review. Ping @darktears . Thanks!

rakuco:
Is there a way to make this patch more resilient to being reverted? https://build.crosswalk-project.org/try/builders/Crosswalk%20Linux/builds/3486 is testing an unrelated patch right after this pull request was processed, and it is possible to see how some extension tests are failing because they're trying to access XW_MessagingInterface_2 somehow.

huningxin:
@rakuco , thanks for pointing out this.

This patch introduces a test which builds an external extension to test the XW_MessagingInterface_2. The external extension remains there even if this patch is reverted. Is it possible to remove this external extension after this patch is reverted?

Anyway, I will fix the test external extension to check whether XW_MessagingInterface_2 is supported before using it.

huningxin:
@rakuco , the test external extension is updated to work with xwalk without XW_MessagingInterface_2. I am not able to wait for bots (midnight here). Please take a look once bots are green. I suppose the bots should replace the old buggy `echo_extension_messaging_2.[dll|so]` with this fixed one. Thanks!

huningxin:
ping @darktears , @rakuco . PTAL. Thanks!

huningxin:
@darktears for the PR.
@rakuco for the test case update.

Thanks!

huningxin:
ping @darktears for the PR.
ping @rakuco for the test case update.

Thanks!

rakuco:
Alexis is only coming back next week. I didn't watch the discussion in crosswalk-dev, but at least the test looks fine now.

huningxin:
Thanks, @rakuco . I will ping Alexis next week.

halton:
ping @darktears

huningxin:
@darktears , could you please take a look?

halton:
Since @darktears give lgtm, i'll merge it to delegate him.

chandanusin:
Ca we use the Binary Messaging Interface for iOS as well or is it only an Android extension ?

halton:
@jondong could you anwser @chandanusin 's question?

jondong:
@chandanusin I'm afraid this feature is only available on android, due to the restriction of iOS WKWebView's interface, the message passed between JS and native can only be string based, so this feature is not provided with Crosswalk iOS.

chandanusin:
Hi thanks a lot for reply. 
I had one more query. 

I am trying to render two video sources using crosswalk wkwebview but am not able to play it simultaneously. 
The same is possible on native using AvPlayer. Although Apple Developer documentation says that it is not possible through html, can you, if possible provide a rationale behind it or a work around with cross walk if possible.

Kindly forward this query to the right person in case. Thanks a lot in advance.

halton:
@chandanusin this is not right place to ask questions, strongly recommended you to ask at https://lists.crosswalk-project.org/mailman/listinfo/crosswalk-dev

jondong:
@chandanusin I'm afraid this is not available for Crosswalk iOS either. For video features we totally rely on WKWebView's ability for now. and actually we don't have future plan to enhance this feature within Crosswalk iOS (Which could be huge effort but less benefit from our perspective if implement outside web engine). So I have to say sorry.
And it could be much better for you to get answer and find the right person to ask in the mail thread: https://lists.crosswalk-project.org/mailman/listinfo/crosswalk-help and https://lists.crosswalk-project.org/mailman/listinfo/crosswalk-dev

chandanusin:
really appreciate the help @halton  and @jondong .

PR #3156, opened 2015-07-27
[Android] Fix the crash issue when running file transfer.

wuhengzhi:
This patch is to fix the crash issue when running file transfer tests in
mobilespeck.
The exposed APIs were not initialized when the explicit constructor was
not existed in the internal java file.
Add the default constructor to call init function.

BUG=XWALK-4667

### DIFF
--- a/tools/reflection_generator/java_class.py
+++ b/tools/reflection_generator/java_class.py
@@ -85,6 +85,7 @@ class InternalJavaFileData(object):
     self._imports = []
     self._enums = {}
     self._package_name = ''
+    self._need_default_constructor = True
 
   @property
   def class_name(self):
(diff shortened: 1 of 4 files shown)


wuhengzhi:
@fujunwei @lincsoon Please help to review, thanks.

tools/reflection_generator/java_class.py
```diff
@@ -322,6 +329,7 @@ def ExtractMethods(self, java_content):
           method_name, method_return, method_params,
           method_annotation, method_doc)
       self._methods.append(method)
+      self._need_default_constructor = False
```
lincsoon: A class doesn't need default constructor just because it has a static method? We should check the annotation noInstance instead.

wuhengzhi:
@lincsoon Modified, thanks.

PR #3162, opened 2015-07-29
[Android] Fixed test error or XWalkRuntimeClient

lincsoon:
Added the api related javascripts to the apk of XWalkRuntimeClientShell

BUG=XWALK-4711
(cherry picked from commit 3c0a6bf1d25fba6e5ca287684c071e9c76340dcb)


lincsoon:
@halton It seems okay except the error of update scripts

halton:
@lincsoon The sysapps APIs are deprecated, I do not think we need test them. Presentation API should be cover.

PR #3163, opened 2015-07-29
[Android] Disable the certificate request test case.

wuhengzhi:
Disable the test case for certificate request since the device can not
access network.


wuhengzhi:
@lincsoon @rakuco Please merge it, thanks.

lincsoon:
LGTM

rakuco:
OK, it does make the test stop failing, but what is the plan to actually fix it?

wuhengzhi:
Currently all of test cases which use web url are disabled. The test device should be replaced to enable these test cases.

rakuco:
Well it's not really a matter of replacing the device, but choosing whether we want the devices to have network access or not (the devices themselves can connect to any network just fine).
I'm going to merge this one to get the tree back to green, but we need to devise a plan for those then.

PR #3165, opened 2015-07-31
DEPS.xwalk: Roll chromium-crosswalk (ecadc9->1021d6)

axinging:
1021d6 [Backport] Rather than dropping stale navigations, resurrect the navigation entry.
f19d78 [Backport] Don't discard navigations.

BUG=XWALK-4378

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = 'ecadc9f25c767bb05ed7513d47864b0015772090'
+chromium_crosswalk_rev = '1021d6640d06558bdf664a2ca958cd17511e7b72'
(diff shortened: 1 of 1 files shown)


rakuco:
Please update your Crosswalk checkout; the old hash you're replacing is wrong.

axinging:
@rakuco  Any detail about this?

axinging:
@rakuco

portwatcher:
When can we use this fixed function and on which version can we use?

Harbon:
+1

rakuco:
It is going to be part of the next canary (15.44.382.0).

portwatcher:
when will 15.44.382.0 be stable? a month or so?

rakuco:
.382 was just the first canary to contain this commit. We branched Crosswalk 15 to the beta channel at 15.44.384.0, so all further betas and stable releases will come from that one.
You can already test v15 beta, and I guess the first stable should come in maybe a month.

portwatcher:
@rakuco  Time flies, it's already mid-October now. :smiley:

PR #3169, opened 2015-07-31
interpret android-L as android-20 by modifing GetAndroidApiLevel() in…

chuan9:
… make_apk.py

BUG=XWALK-3233


rakuco:
The code does not work because the regular expression above the change will have filtered out non-numbered entries. In any case, I've asked a few more questions in the ticket.

chuan9:
The code only checks android-L version and Belem zhang has provided some info about this bug. If you have any question, pls give some more comments here. Thanks

rakuco:
It is better to gather all comments in a single place instead of spreading them across JIRA and GitHub, so I'm posting my concerns there.

PR #3172, opened 2015-08-04
[Android] Rewrite merge_jars.py.

rakuco:
Almost all the code that was present in the script is also available in
Chromium itself, so it makes no sense to have the uptenth
reimplementation of GetCommandOutput() or calls to jar(1).

Additionally, the handling of the --jars option was broken: it always
assumed that each path would be properly escaped by gyp and could be
eval()ed directly into Python. Not only does calling eval() in this
context make zero sense, but it also fails with targets such as
`android_support_v13_javalib`, which is passed as an absolute path
without escaped shell quotes. This is a problem starting with M45, as
that target is depended upon by `content_java`.

The --classes-dir option is also useless: instead of creating and
cleaning directories inside the build directory, just use a different
temporary directory each time.

The only difference in the generated jar files is the presence of a
temporary directory with an `.empty` file there that comes from the code
in src/build/android/gyp/jar.py. It should not cause any trouble.

### DIFF
--- a/build/android/merge_jars.py
+++ b/build/android/merge_jars.py
@@ -5,67 +5,33 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-import fnmatch
 import optparse
 import os
 import sys
-import shutil
-import subprocess
 
+GYP_ANDROID_DIR = os.path.join(os.path.dirname(__file__),
+                               os.pardir, os.pardir, os.pardir,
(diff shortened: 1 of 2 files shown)


rakuco:
ping @halton, @darktears, @lincsoon and @mrunalk.

This is required for M45, but I'm landing it separately because it can go into master without the rest of M45.

halton:
LGTM

PR #3174, opened 2015-08-05
Work around Skia performance regression

qkmiao:
A Skia CL (https://codereview.chromium.org/607993002 ) made CrossWalk
regress on some devices such as Galaxy S3. Upstream proposed a CL to fix
the performance regression in
https://codereview.chromium.org/1161063003. Alexis backported upstream's
fix to CrossWalk master in
https://github.com/crosswalk-project/crosswalk/pull/3096. But it caused
regression bug: XWALK-4467. The problem is there are many patches
between Skia roll in CrossWalk and upstream Skia and the backported
patch needed other dependent patches to work correctly.

In this patch, I revert the Skia backport patch and cherry-pick the
workaround in CrossWalk 14
(https://github.com/crosswalk-project/crosswalk/pull/3096). We can remove
this workaround after next Skia roll.

BUG=XWALK-3720, XWALK-4467, XWALK-4589


qkmiao:
@rakuco @baleboy PTAL.

PR #3176, opened 2015-08-06
[Linux] Fix the incorrect icon update issue

XiaosongWei:
Ensure at least the ImageSkiaRep's for the default 1x scale is always
attached to ImageSkia's storage so that HasRepresentation for the
default 1x scale return a meaningful true then the icon updator can get
a chance to retrieve the 1x scale Rep.

BUG=XWALK-4763

### DIFF
--- a/runtime/browser/ui/native_app_window_views.cc
+++ b/runtime/browser/ui/native_app_window_views.cc
@@ -4,6 +4,8 @@
 
 #include "xwalk/runtime/browser/ui/native_app_window_views.h"
 
+#include <vector>
+
 #include "base/command_line.h"
 #include "content/public/browser/notification_service.h"
 #include "content/public/browser/web_contents.h"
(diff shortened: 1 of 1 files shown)


rakuco:
Since there is no associated bug and the PR description only describes what needs to be done, it is still unclear to me what is being fixed.

XiaosongWei:
@rakuco @pozdnyakov 
The issue is found when running test: 
xwalk --app-icon=path_to_icon.png http://www.google.com

rakuco:
OK, at least we have a direction now :-) But it is more helpful if you create a JIRA ticket describing what's wrong and reference it here, so that people do not need to run Crosswalk themselves to find out what is broken (which is going to be even more difficult once the issue is fixed).

pozdnyakov:
Could you pls explain what is fixed from the User perspective?

XiaosongWei:
@pozdnyakov pls find detailed issue description in BUG=XWALK-4763, QA confirmed the issue is reproducible.

runtime/browser/ui/native_app_window_views.cc
```diff
@@ -229,7 +231,17 @@ gfx::ImageSkia NativeAppWindowViews::GetWindowAppIcon() {
...
-  return *icon_.ToImageSkia();
+  const gfx::ImageSkia *result =  icon_.ToImageSkia();
+
+  const std::vector<float> oldSupportedScales =
+      gfx::ImageSkia::GetSupportedScales();
+  std::vector<float> scale = { 1.0f };
+  gfx::ImageSkia::SetSupportedScales(scale);
+  // Ensure at least the ImageSkiaRep for the default 1x scale is attached to
+  // ImageSkia's storage
+  result->EnsureRepsForSupportedScales();
```
pozdnyakov: shouln't 'result' be null-checked?

pozdnyakov:
lgtm, with a question

runtime/browser/ui/native_app_window_views.cc
XiaosongWei: The function ToImageSkia always return a valid pointer, so let's keep the original no null-checked.

PR #3184, opened 2015-08-11
DEPS.xwalk: Roll chromium-crosswalk (6e988a->97959a)

darktears:
cd38a1 : Fix build on Windows when using components build.


darktears:
@rakuco

PR #3186, opened 2015-08-12
Remove use of MessageLoopProxy and deprecated MessageLoop APIs.

rakuco:
Chromium is deprecating MessageLoopProxy and moving towards the TaskRunner API instead. [Bug 465354](http://crbug.com/465354) contains more information.

Since we can already do this in M44, do it separately from moving to M45 so that the size of the M45 patch is smaller. Thanks to Olli Raula and Mrunal Kapade for doing the majority of the work.

Most of the changes are mechanic:
* `base::MessageLoopProxy::current()` becomes `base::ThreadTaskRunnerHandle::Get()`.
* `scoped_refptr<base::MessageLoopProxy>` becomes `scoped_refptr<base::SingleThreadTaskRunner>`.
* `base::Thread::message_loop_proxy()` becomes `base::Thread::task_runner()`.

The only noteworthy change is in `XWalkExtensionFunctionHandler`, caused by the fact that `ThreadTaskRunnerHandle::Get()` enforces an assertion that was commented out in `MessageLoopProxy`. This assertion makes sure the call is being performed in a thread with a valid task runner.

That is not the case for XWalkExtensionFunctionHandlerTest's `PostingResultAfterDeletingTheHandler`. Its call to `XWEFH::HandleMessage()` is done without a runner, so we need to check before calling `ThreadTaskRunnerHandle::Get()`. This is similar to Chromium commits afd978c6 ("gpu: Fix webview parent compositor crash in context loss") and 0ad2c54d ("gpu: Remove use of MessageLoopProxy and deprecated MessageLoop APIs").

### DIFF
--- a/extensions/browser/xwalk_extension_function_handler.cc
+++ b/extensions/browser/xwalk_extension_function_handler.cc
@@ -5,6 +5,7 @@
 #include "xwalk/extensions/browser/xwalk_extension_function_handler.h"
 
 #include "base/location.h"
+#include "base/thread_task_runner_handle.h"
 #include "xwalk/extensions/common/xwalk_external_instance.h"
 
 namespace xwalk {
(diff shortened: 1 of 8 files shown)


rakuco:
@Bysmyyr @pozdnyakov @mrunalk

pozdnyakov:
lgtm

PR #3194, opened 2015-08-14
Cleanup some mess checked in.

darktears:
This should probably not be here.


darktears:
@rakuco

darktears:
So apparently it's needed. What a weird name and place to be.

rakuco:
agreed. I think it was added with the LZMA SDK stuff, and it can certainly be put somewhere more reasonable.

rakuco:
> it can certainly be put somewhere more reasonable.

Done in #3198.

PR #3195, opened 2015-08-14
Crosswalk.ico for Windows

darktears:


darktears:
@@r0b5t4 Can you try it before I merge it?

r0b5t4:
@darktears if that makes it show up in the release zip, LGTM.

rakuco:
Some more context would be appreciated. The file won't be automatically part of anything we release. What problem is this solving and where should it show up?

r0b5t4:
The android release ships the crosswalk icon, so if a users did not provide one in the web manifest, we can default to that, and still build an app that's good enough for being run and tested (android would probably fall back to the built-in icon anyway).

On windows it would be good if we were able to do the same, and use the crosswalk icon by default. Windows only accepts .ico files, so when someone wants to casually rebuild an android app for win, it would be good to have a default icon.

kenchris:
can't we generate the ico file somehow? like https://gist.github.com/pfig/1808188

What does manifoldjs.com do ?

r0b5t4:
@kenchris depending on imagemagick might be a possibility on linux, on OS X and Windows not so much. What's the issue with putting the icon into the release zip? It's 38K.

kenchris:
None... I was understanding that you need to use .ico files, even for what is in the manifest, so we much convert them no?

r0b5t4:
For the record:

(15:20:15) @kenchris: Answer [...]: actually, windows apps only support png, gif and jpg. .ico are not supported.
(15:23:21) @kenchris: it changed in windows 8

We need to work out how to go about this, possibly building separate packages for Windows 7 and 8+.

darktears:
@r0b5t4 : We don't really target Windows 7 so I'm thinking that we should be fine on that part. Let's focus on Windows 8+. So if you can put a PNG/JPG then we're kind of fine.

@rakuco to answer your question about the release part, when we will create the zip file to release Windows (containing xwalk.exe and the various other files) then we could add this one so that crosswalk-app-tools has a default icon to use BUT maybe it's not needed. Right now you can use the ones of Android though they are deep into Android specific directories and of course on Windows, crosswalk-app-tools can't access it. What about actually having Crosswalk-app-tools having the default icon?

r0b5t4:
@darktears Crosswalk-app-tools already has the png default icon, so we're good for Windows 8+, and this PR is probably moot.

PR #3197, opened 2015-08-17
[Tizen] Stop depending on the extensions_browser gyp target.

rakuco:
The dependency was introduced when we moved to M42, as there was a
layering violation that introduced a dependency on extensions/ that we
had to provide ourselves (see comment #30 in http://crbug.com/449919).

This has since been fixed upstream (http://crbug.com/477096), so we can
remove the workaround and get rid of a lot of unnecessary dependencies.


pozdnyakov:
lgtm

kenchris:
lgtm

PR #3198, opened 2015-08-17
[Android] Move dummy/dummy.cc to a reasonable location, rename.

rakuco:
A top-level directory called "dummy" is a terrible choice that confuses
pretty much anyone that looks at it -- there is only a single-line
explanation in the only file there giving a hint that it is something
Android-specific.

Move the directory to runtime/android and call it "dummy_lib" instead;
this way, the association with Android is clear and people can at least
know that it is a library.


rakuco:
ping @darktears @lincsoon @halton @d0ngjun

d0ngjun:
LGTM.

halton:
LGTM

lincsoon:
LGTM

PR #3200, opened 2015-08-18
[Linux] Add XWalk's download support.

XiaosongWei:
[Linux] Add XWalk's download support.
    
    This change includes file saving file picker and download UI. The default
    download path on Linux will be the user's ~/Downloads/.
    
    Each NativeAppWindowDesktop contains one DownloadBarView which contains one or multiple
    DownloadItemView. Each DownloadItemView will be associated with a DownloadItem and update
    its look when download status updated.


XiaosongWei:
@rakuco @pozdnyakov 

This is requested by Deepin. Previously we use platform's default download manner (xdg-open) to download the given item but this approach may lead to download failure when the server requests client to send back some information for generating contents dynamically.  Therefore, we have to use XWalk's internal download to handle all download cases.

XiaosongWei:
@rakuco @pozdnyakov Please take a look, Thanks.

runtime/browser/ui/native_app_window_desktop.h
```diff
@@ -21,11 +21,17 @@ namespace xwalk {
 class NativeAppWindowDesktop : public NativeAppWindowViews,
                                public views::ButtonListener {
  public:
+  enum ViewID {
+    VIEW_ID_NONE = 0,
```
pozdnyakov: Usually a window class does not need to know about its implementations

pozdnyakov: These Ids better to be declared outside the class

runtime/browser/ui/native_app_window_desktop.cc
```diff
@@ -215,4 +228,19 @@ void NativeAppWindowDesktop::SetAddressURL(const std::string& url) {
   }
 }
 
+void NativeAppWindowDesktop::UpdateWebViewPreferredSize() {
+  int height = this->height();
+  if (toolbar_view_) {
+    height -= toolbar_view_->GetPreferredSize().height();
+    // Minus vertical paddings.
+    height = height - 2 - 5 - 3;
```
pozdnyakov: - 2 - 5 - 3;  magic numbers are not acceptable

pozdnyakov: btw, isn't it possible that 'height' will be negative here?

runtime/browser/ui/native_app_window_desktop.cc
```diff
@@ -215,4 +228,19 @@ void NativeAppWindowDesktop::SetAddressURL(const std::string& url) {
   }
 }
 
+void NativeAppWindowDesktop::UpdateWebViewPreferredSize() {
+  int height = this->height();
```
pozdnyakov: this-> can be omitted

runtime/browser/ui/native_app_window_desktop.cc
kenchris: Yeah that is not code that I would easily understand when I read it

runtime/browser/runtime_download_manager_delegate.h
```diff
@@ -51,10 +89,23 @@ class RuntimeDownloadManagerDelegate
...
+  void FileSelectionCanceled(void* params) override;
+#endif
+
   content::DownloadManager* download_manager_;
   base::FilePath default_download_path_;
   bool suppress_prompting_;
 
+#if defined(OS_LINUX) && !defined(OS_TIZEN)
+  scoped_refptr<ui::SelectFileDialog> select_file_dialog_;
+  std::map<content::WebContents*, RuntimeDownloadView*> download_views_;
```
pozdnyakov: std::map\<content::WebContents*, scoped_ptr\<RuntimeDownloadView>> ?

runtime/browser/runtime_download_manager_delegate.cc
```diff
@@ -40,7 +76,12 @@ RuntimeDownloadManagerDelegate::RuntimeDownloadManagerDelegate()
   AddRef();
 }
 
-RuntimeDownloadManagerDelegate::~RuntimeDownloadManagerDelegate() {}
+RuntimeDownloadManagerDelegate::~RuntimeDownloadManagerDelegate() {
+#if defined(OS_LINUX) && !defined(OS_TIZEN)
+  for (auto& key_view : download_views_)
+    delete(key_view.second);
```
pozdnyakov: 'delete' is not a function.
smart ptrs should be called, see comment below

runtime/browser/runtime_download_manager_delegate.cc
```diff
@@ -55,20 +96,16 @@ bool RuntimeDownloadManagerDelegate::DetermineDownloadTarget(
...
-    platform_util::OpenExternal(download->GetURL());
-    return false;
-  }
-#endif
   // This assignment needs to be here because even at the call to
   // SetDownloadManager, the system is not fully initialized.
   if (default_download_path_.empty()) {
+#if defined(OS_LINUX) && !defined(OS_TIZEN)
+    default_download_path_ =
+        base::nix::GetXDGUserDirectory("DOWNLOAD", "Downloads");
```
pozdnyakov: where this "Downloads" comes from?

runtime/browser/runtime_download_manager_delegate.h
```diff
@@ -40,6 +51,33 @@ class RuntimeDownloadManagerDelegate
...
+    DownloadSelectFileParams(content::DownloadItem* item,
+                             const content::DownloadTargetCallback& callback);
+    ~DownloadSelectFileParams();
+
+    content::DownloadItem* item;
+    content::DownloadTargetCallback callback;
+  };
+
+#if defined(OS_LINUX) && !defined(OS_TIZEN)
+  class RuntimeDownloadView : public content::WebContentsObserver {
```
pozdnyakov: DownloadManagerDelegate is not aware of UI and it should not. pls take a look at 'xwalk/runtime/browser/runtime_ui_delegate_desktop.h', the 'RuntimeUIDelegateDesktop' class is the place where UI should be "attached" to a Runtime instance.

pozdnyakov: Let the 'RuntimeUIDelegateDesktop' class contain DownloadBarView member

runtime/browser/runtime_download_manager_delegate.cc
XiaosongWei: this is a fallback directory name if user's "DOWNLOAD" dir not found.

runtime/browser/runtime_download_manager_delegate.h
XiaosongWei: agree that UI stuffs should be placed within UIDelegate/NativeAppWindow. I will refactor the code based on this idea. Thanks for your suggestions.

XiaosongWei:
@pozdnyakov code refactored, please take a look. Thanks!

runtime/browser/runtime_download_manager_delegate.cc
```diff
@@ -178,12 +189,31 @@ void RuntimeDownloadManagerDelegate::ChooseDownloadPath(
...
   if (GetSaveFileName(&save_as))
     result = base::FilePath(std::wstring(save_as.lpstrFile));
-#else
-  NOTIMPLEMENTED();
-#endif
 
   callback.Run(result, content::DownloadItem::TARGET_DISPOSITION_PROMPT,
                content::DOWNLOAD_DANGER_TYPE_NOT_DANGEROUS, result);
+#elif defined(OS_LINUX) && !defined(OS_TIZEN)
+  content::WebContents* contents = item->GetWebContents();
```
pozdnyakov: I dislike that we create a dialog here. 
Let's instead obtain 'Runtime' and invoke 'runtime->AddDownloadItem(item); // further UIDelegate is handling everything (including dialog creation)'

runtime/browser/runtime.cc
```diff
@@ -342,4 +342,18 @@ void Runtime::LoadProgressChanged(content::WebContents* source,
...
 }
 
+void Runtime::AddDownloadItem(content::DownloadItem* download_item,
+    const content::DownloadTargetCallback& callback,
+    const base::FilePath& suggested_path) {
+  if (ui_delegate_) {
+    ui_delegate_->AddDownloadItem(download_item, callback, suggested_path);
+  } else {
+    const base::FilePath empty;
+    callback.Run(empty,
```
pozdnyakov: why this code is necessary?

pozdnyakov: I mean why not simply:
{
if (ui_delegate_) 
    ui_delegate_->AddDownloadItem(download_item, callback, suggested_path);
}

runtime/browser/ui/download_views.cc
```diff
@@ -0,0 +1,271 @@
...
+const SkColor kDownloadCompleteColor = SkColorSetARGB(128, 165, 214, 167);
+const SkColor kDownloadCancelledColor = SkColorSetARGB(128, 239, 154, 154);
+
+}  // namespace
+
+DownloadBarView::DownloadBarView(NativeAppWindow* app_view)
+    : download_image_(nullptr),
+      ellipsis_label_(nullptr),
+      close_button_(nullptr) {
+  parent_ = static_cast<NativeAppWindowDesktop*>(app_view);
```
pozdnyakov: better to pass as NativeAppWindowDesktop*

runtime/browser/runtime.cc
XiaosongWei: Per the comment in DownloadManagerDelegate::DetermineDownloadTarget(), If the download should be canceled, |callback| should be invoked with an empty |target_path| argument.

runtime/browser/ui/download_views.cc
XiaosongWei: Yes, agree.

runtime/browser/ui/download_views.h
```diff
@@ -0,0 +1,85 @@
...
+#include "ui/views/controls/image_view.h"
+#include "ui/views/controls/label.h"
+
+namespace xwalk {
+
+class DownloadItemView;
+class NativeAppWindow;
+class NativeAppWindowDesktop;
+
+class DownloadBarView : public views::ButtonListener,
```
pozdnyakov: let's create a 'runtime/browser/ui/desktop' folder and put this class there

runtime/browser/ui/download_views.h
```diff
@@ -0,0 +1,85 @@
...
+                        public views::View {
+ public:
+  explicit DownloadBarView(NativeAppWindow* app_view);
```
pozdnyakov: NativeAppWindowDesktop* app_view

runtime/browser/ui/download_views.h
XiaosongWei: Do you mean to put download_views.h/cc under "runtime/browser/ui/desktop"?

pozdnyakov: yeah, since this is part of desktop-specific UI

runtime/browser/runtime.cc
pozdnyakov: let 'AddDownloadItem' return bool (both at Runtime and UI delegate) and invoke |callback| on upper level (if Runtime::AddDownloadItem() == false )

runtime/browser/ui/download_views.cc
```diff
@@ -0,0 +1,271 @@
...
+}
+
+void DownloadBarView::Layout() {
+  views::View::Layout();
+
+  int x, y, width, height;
+  gfx::Size image_size = download_image_->GetPreferredSize();
+  width = image_size.width() + 2 * kDownloadImageHorizontalPadding;
+  height = image_size.height() + 2 * kDownloadImageVerticalPadding;
+  y = (height - image_size.height()) / 2;
```
pozdnyakov: isn't y==kDownloadImageVerticalPadding ??

runtime/browser/ui/download_views.cc
```diff
@@ -0,0 +1,271 @@
...
+  download_image_->SetBounds(kDownloadImageHorizontalPadding, y,
+                             image_size.width(), image_size.height());
+
+  gfx::Size btn_size = close_button_->GetPreferredSize();
+  x = this->width() - btn_size.width() - kCloseButtonRightPadding;
```
pozdnyakov: this-> can be omitted

runtime/browser/ui/download_views.cc
```diff
@@ -0,0 +1,271 @@
...
+  y = (height - btn_size.height()) / 2;
+  close_button_->SetBounds(x, y, btn_size.width(), btn_size.height());
+
+  int max_item_x = x - kCloseButtonLeftPadding - kMaxItemTextWidth;
```
pozdnyakov: how you make sure you are not getting negative values here (and within other calculations)?

runtime/browser/ui/download_views.cc
```diff
@@ -0,0 +1,271 @@
...
+  x = width;
+  for (auto& item : download_item_views_) {
```
pozdnyakov: since we know the type, it's better to put it explicitly (not 'auto')

runtime/browser/ui/download_views.cc
XiaosongWei: yes, it is. use kDownloadingImageVerticalPadding explicitly.

runtime/browser/ui/download_views.cc
XiaosongWei: width() is conflict with int width, so prefer to keep it here.

runtime/browser/ui/download_views.cc
XiaosongWei: max_item_x doesn't need to be positive as we just use it for position comparison. For others, negative values are acceptable, for example, origin (x,y) in SetBounds, as it doesn't impact the actual layout.

runtime/browser/ui/download_views.cc
XiaosongWei: it's fine to use explicit type, however it seems auto is conciser.

pozdnyakov:
lgtm

runtime/browser/ui/desktop/download_views.cc
```diff
@@ -0,0 +1,271 @@
...
+  gfx::Size image_size = download_image_->GetPreferredSize();
+  width = image_size.width() + 2 * kDownloadImageHorizontalPadding;
+  height = image_size.height() + 2 * kDownloadImageVerticalPadding;
+  download_image_->SetBounds(kDownloadImageHorizontalPadding,
+      kDownloadImageVerticalPadding,
+      image_size.width(), image_size.height());
+
+  gfx::Size btn_size = close_button_->GetPreferredSize();
+  x = this->width() - btn_size.width() - kCloseButtonRightPadding;
+  y = (height - btn_size.height()) / 2;
```
pozdnyakov: could it be that 'height < btn_size.height()' ? Please make sure you always have valid (positive) values for the layout metrics

XiaosongWei: No. images used in download_image_ and close_button are fixed size and the former height is always > the latter.

pozdnyakov: pls put DCHECK for  every '-' operation result in this code (in case image or button will be accidentally modified)

pozdnyakov: I mean DCHECK 'x', 'y', 'max_item_x' and others..

PR #3203, opened 2015-08-18
[Android] Fix the javascript was not executed issue.

wuhengzhi:
Sometimes XWalkView was created in onXWalkReady(), if the activity which
owns XWalkView was destroyed, pauseTimers() will be invoked, then
JavaScript timer was also paused. Reentry this activity, the
resumeTimers() was not called since XWalkView creation was later than
onResume().
Invoke resumeTimers() explicitly when XWalkView was created.

BUG=XWALK-4676

(cherry picked from commit 421d1dfbba5f4bac76a504a48083367073f243e3)


wuhengzhi:
@lincsoon @halton Please merge it to crosswalk-15, thanks.

PR #3211, opened 2015-08-24
[Android]Support external extensions written in C++ on Android.

minggangw:
Before this patch, Crosswalk only supports the extensions written in
Java.

With this patch, Crosswalk can load native C++ extensions by calling a
Java API, which will tell the aboslute location to the XWalkExtensionService.
Then when the extension service starts, it will load the extension from
the proper path. To achieve this purpose, add a new option "--native-extensions"
for make_apk.py. By setting it to the location of the folder, which
contains the c++ extension library (.so), the make_apk will pack them into
the final APK file.

BUG=XWALK-4749


halton:
@minggangw why close this PR?

PR #3213, opened 2015-08-24
[Android]Support external extensions written in C++ on Android

minggangw:
Before this patch, Crosswalk only supports the extensions written in
Java.

With this patch, Crosswalk can load native C++ extensions by calling a
Java API, which will tell the aboslute location to the XWalkExtensionService.
Then when the extension service starts, it will load the extension from
the proper path. To achieve this purpose, add a new option
"--native-extensions" for make_apk.py. By setting it to the location of
the folder, which contains the c++ extension library, the make_apk will
pack them into the final APK file.

Also, this patch creates two unit tests for native external extensions
for runtime client, including shared/embedded modes.

BUG=XWALK-4749

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -10,6 +10,7 @@ import optparse
 import os
 import re
 import shutil
+import struct
 import subprocess
 import sys
 import tempfile
@@ -62,6 +63,16 @@ def ConvertArchNameToArchFolder(arch):
   }
   return arch_dict.get(arch, None)
 
+def ConvertArchNameToArchValue(arch):
+  # The processor types represented in ELF file header follow
+  # http://www.sco.com/developers/gabi/latest/ch4.eheader.html
+  arch_dict = {
+      'x86': 3,
+      'x86_64': 50,
+      'arm': 40,
+      'arm64': 183
+  }
+  return arch_dict.get(arch, None)
 
 def AddExeExtensions(name):
   exts_str = os.environ.get('PATHEXT', '').lower()
@@ -95,6 +106,21 @@ def GetAndroidApiLevel(android_path):
   targets.extend([-1])
   return max(targets)
 
+def GetNativeLibraryArchValueInELFHeader(path):
+  library = open(path, 'rb')
+  # The offset of the architecture value is 0x12 from the ELF header,
+  # the length is 2 bytes.
+  header = library.read(20)
+  return struct.unpack_from('H', header, 18)[0]
+
+def CheckValidationOfExpectedLibraryArch(path, expected_arch):
+  extension_path_list = FindNativeExtensionFiles(path);
+  for extension_path in extension_path_list:
+    if not ConvertArchNameToArchValue(expected_arch) == \
+        GetNativeLibraryArchValueInELFHeader(extension_path):
+      print('Invalid CPU arch of %s, %s is expected.' % (extension_path,
+                                                         expected_arch))
+      sys.exit(10)
 
 def ContainsNativeLibrary(path):
   return os.path.isfile(os.path.join(path, NATIVE_LIBRARY))
@@ -328,6 +354,23 @@ def CopyCompressedLibrary(native_path, library_path, raw_path, arch):
 def CopyNativeLibrary(native_path, library_path, raw_path, arch):
   shutil.copytree(native_path, os.path.join(library_path, arch))
 
+def CopyNativeExtensionFile(suffix, src_path, dest_path):
+  if os.path.exists(src_path):
+    for afile in os.listdir(src_path):
+      if afile.endswith(suffix):
+        shutil.copy(os.path.join(src_path, afile), dest_path)
+
+def FindNativeExtensionFiles(path):
+  ''' Find all .so files for native external extensions. '''
+  extension_list = []
+  if not os.path.exists(path):
+    return extension_list
+
+  for afile in os.listdir(path):
+    if os.path.isfile(os.path.join(path, afile)) and afile.endswith('.so'):
+      extension_list.append(os.path.join(path, afile))
+
+  return extension_list
 
 def Execution(options, app_info):
   # Now we've got correct app_version and correct ABI value,
@@ -447,12 +490,30 @@ def Execution(options, app_info):
     for dir_name in os.listdir(library_path):
       clean_library(library_path, dir_name)
 
+    if options.native_extensions:
+      CheckValidationOfExpectedLibraryArch(options.native_extensions,
+                                           options.arch)
+      CopyNativeExtensionFile('.so', os.path.join(options.native_extensions,
+                                                  arch),
+                              native_path)
+
     if contains_library(native_path):
       copy_library(native_path, library_path, raw_path, arch)
     else:
       print('No %s native library has been found for creating a Crosswalk '
             'embedded APK.' % arch)
       sys.exit(10)
+  else:
+    if options.native_extensions:
+      for arch_name in ALL_ARCHITECTURES:
+        arch = ConvertArchNameToArchFolder(arch_name)
+        extension_path = os.path.join(app_dir, SHARED_LIBRARY, 'libs', arch)
+        library_path = os.path.join(options.native_extensions, arch)
+        CheckValidationOfExpectedLibraryArch(library_path, arch_name)
+        os.mkdir(extension_path)
+        CopyNativeExtensionFile('.so', os.path.join(options.native_extensions,
+                                                    arch),
+                                extension_path)
 
   if options.project_only:
     print (' (Skipping apk package creation)')
@@ -757,6 +818,14 @@ def main(argv):
           'Mac OS respectively. For example, '
           '--extensions=/path/to/extension1:/path/to/extension2.')
   group.add_option('--extensions', help=info)
+  info = ('The native external extension path. Put the architecture-specific '
+          'library file in the corresponding architecture folder '
+          '(armeabi-v7a/arm64-v8a/x86/x86_64) under '
+          'the path. For example, '
+          '--native-extensions=/path/to/native_extension, '
+          'architecture-specific folders, which contain the library, can be '
+          'found under the path.')
+  group.add_option('--native-extensions', help=info)
   group.add_option('-f', '--fullscreen', action='store_true',
                    dest='fullscreen', default=False,
                    help='Make application fullscreen.')
--- a/extensions/common/android/xwalk_native_extension_loader_android.cc
+++ b/extensions/common/android/xwalk_native_extension_loader_android.cc
@@ -1,0 +1,29 @@
+// Copyright (c) 2015 Intel Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/extensions/common/android/xwalk_native_extension_loader_android.h"
+
+#include <string>
+
+#include "jni/XWalkNativeExtensionLoaderAndroid_jni.h"
+#include "xwalk/runtime/browser/xwalk_browser_main_parts_android.h"
+#include "xwalk/runtime/browser/xwalk_content_browser_client.h"
+
+namespace xwalk {
+namespace extensions {
+
+void RegisterExtensionInPath(JNIEnv* env, jclass jcaller, jstring path) {
+  const char *str = env->GetStringUTFChars(path, 0);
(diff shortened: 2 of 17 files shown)


minggangw:
@halton please review this PR, thanks!

xwalk_core_library_android.gypi
```diff
@@ -26,6 +26,7 @@
           '>(reflection_gen_dir)/wrapper/org/xwalk/core/XWalkJavascriptResult.java',
           '>(reflection_gen_dir)/wrapper/org/xwalk/core/XWalkNavigationHistory.java',
           '>(reflection_gen_dir)/wrapper/org/xwalk/core/XWalkNavigationItem.java',
+          '>(reflection_gen_dir)/wrapper/org/xwalk/core/XWalkNativeExtensionLoader.java',
```
rakuco: Please keep this alphabetically sorted.

runtime/browser/xwalk_browser_main_parts_android.cc
```diff
@@ -259,4 +259,12 @@ XWalkExtension* XWalkBrowserMainPartsAndroid::LookupExtension(
   return NULL;
 }
 
+void XWalkBrowserMainPartsAndroid::RegisterExtensionInPath(
+    const std::string& path) {
+  if (extension_service_) {
```
rakuco: Is it actually possible for `extension_service_` to be `NULL` at this point?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkNativeExtensionLoaderInternal.java
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) 2015 The Chromium Authors. All rights reserved.
```
rakuco: Copyright Intel?

extensions/common/android/xwalk_native_extension_loader_android.cc
```diff
@@ -0,0 +1,31 @@
...
+#include "jni/XWalkNativeExtensionLoaderAndroid_jni.h"
+#include "xwalk/runtime/browser/xwalk_browser_main_parts_android.h"
+#include "xwalk/runtime/browser/xwalk_content_browser_client.h"
+
+namespace xwalk {
+namespace extensions {
+
+void RegisterExtensionInPath(JNIEnv* env, jclass jcaller, jstring path) {
+  const char *str = env->GetStringUTFChars(path, 0);
+  std::string p(str);
```
rakuco: This doesn't seem to be used anywhere.

xwalk_core_library_android.gypi
minggangw: Will change the order.

runtime/browser/xwalk_browser_main_parts_android.cc
minggangw: Yes, extension_service_ is impossible to be NULL, the 'if' is redundant.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkNativeExtensionLoaderInternal.java
minggangw: I'll change the copyright :)

extensions/common/android/xwalk_native_extension_loader_android.cc
minggangw: This local variable should be removed here, thank you!

rakuco:
Thanks, the code looks better now. I still have a few comments/ideas:
* Your actual commit message is very different from the pull request message. The former only says "XWALK-4749" without anything else. You should use the latter instead.
* You should probably make the help message for the new make_apk option more informative: it does not say that the extensions need to be in architecture-specific directories inside the directory being passed, and their names are not mentioned anywhere. Adding some more information like "Found native extension of architecture XYZ" when doing the processing is also helpful.
* The lack of tests makes it prone to regressions in the future. You should at least add a test native extension to verify everything is working as expected.

minggangw:
Thank you for your suggestions, I will add some tests to verify this new feature.

minggangw:
Hi, @rakuco . I did some investigations on how to write an automation test to verify the native extension on Android (I supposed the test you mentioned is an automation test) and found some problems:

1. In order to test the native extension, we have to make a real apk file and run it on emulator/device. I think to achieve this we have a lot of things to do.
2. The native C++ extension mechanism itself have been tested on other platform like the  "echo_extension". Here, we just reuse it. The function doesn't need to be tested again.

I discussed with Halton about it and have a sample solution. We can reuse the current native extension C++ test case, write the detailed description on how to test it manually in the commit log. So what is your opinion, thanks!

rakuco:
> In order to test the native extension, we have to make a real apk file and run it on emulator/device. I think to achieve this we have a lot of things to do.

The whole setup to build and run device tests is already in place, and that's how we run all the tests [here](https://build.crosswalk-project.org/builders/Crosswalk%20Tests%20%28Android%20x86%29), for example. IIRC we even test Java extensions this way already. Isn't it just a matter of building another simple extension the way and extending what we already do?

> The native C++ extension mechanism itself have been tested on other platform like the "echo_extension". Here, we just reuse it. The function doesn't need to be tested again.
>
> I discussed with Halton about it and have a sample solution. We can reuse the current native extension C++ test case, write the detailed description on how to test it manually in the commit log. So what is your opinion, thanks!

The problem with this idea is that it basically assumes that:
- The code will not substantially change in the future, and thus the way the C++ extensions work and how they are plugged into the Android code will always work.
- People will somehow know that in this specific commit from a random date in the project's history there are instructions on how to test the C++/Java integration of the extensions in the commit message. Even if the instructions are in a comment in the code instead, John Doe who starts contributing to the project 2 years from now will have no idea that the instructions are there.

Which is why it would be nice to do these things automatically: you just run your usual test suite and know if your changes broke something or not.

IIRC @wuhengzhi added several Android test cases in the past, so he might be able to help you with this. If you still have problems, I can also try to give you a hand.

minggangw:
Ok, I will contact @wuhengzhi .

minggangw:
@rakuco , I've updated the patch by your comments, adding 2 tests and some helpful informations. The instrumentation tests have passed by running test_runner.py instrumentation --release --test-apk=XWalkRuntimeClientEmbeddedTest/XWalkRuntimeClientTest. Please have a look, thank you. And thanks for @wuhengzhi 's help :)

minggangw:
Don't know why the PR closed, just reopen it.

halton:
ping @rakuco

halton:
@rakuco how is your review status on this PR?

halton:
@wuhengzhi , you're right. @minggangw please fix the style thing.

minggangw:
Ok, I will fix that.

halton:
@rakuco the style issues are fixed, please continue your review.

rakuco:
Sorry, I'm focused on M45 at the moment and don't have much time to context switch back to this PR. Overall it looks fine to me now; if it's OK to you and @wuhengzhi then feel free to go ahead.

xwalk_android_tests.gypi
```diff
@@ -26,6 +26,17 @@
       'includes': [ '../build/java.gypi' ],
     },
     {
+      'target_name': 'native_echo_extension',
+      'type': 'loadable_module',
+      'include_dirs': [
+        '..',
+      ],
+      'sources': [
+        'test/android/data/nativeechoextension/native_echo_extension.c',
```
halton: I presume the content of native_echo_extension.c is same as `src/xwalk/extensions/test/echo_extension.c`. Can we re-use the the code and target instead of copy?

xwalk_android_tests.gypi
```diff
@@ -248,12 +259,18 @@
       'dependencies': [
         'xwalk_app_runtime_java',
         'xwalk_runtime_client_test_utils_java',
+        'libxwalkdummy',
```
halton: I do not think this change is related. If you're trying to fix some bug, please submit separate PR

xwalk_android_tests.gypi
```diff
@@ -440,6 +459,7 @@
           '<(PRODUCT_DIR)/runtime_client_test/assets/echo.html',
           '<(PRODUCT_DIR)/runtime_client_test/assets/echoSync.html',
           '<(PRODUCT_DIR)/runtime_client_test/assets/messaging_mini.html',
+          '<(PRODUCT_DIR)/runtime_client_embedded_test/assets/nativeEcho.html',
```
halton: ditto, please re-use the echo.html instead of copy.

xwalk_android_tests.gypi
minggangw: In the current RuntimeClinetShell/RuntimeClientEmbeddedShell test, there is already a Java external extension named "echo" as the js interface, see "test/android/data/extensions-config.json". 
"/test/echo_extension.c" defined the same "echo" as the extension name. So, here I redefine "nativeEcho" as the name of the extension to distinguish the one written in Java.  Most of the codes in echo_extension.c and native_echo_extension.c are the same, excepted the name of the js interface.

xwalk_android_tests.gypi
minggangw: I will check if 'libxwalkdummy' is necessary here, thanks!

minggangw:
After discussing with @halton about the conflicts of the existing naming "echo" js interface in ExternalExtensionTest/NativeExternalExtensionTest tests (one for Java external extension, the other is native c++ extension), I prefer to solve it:

1. Rename the current test case resources (htmls, js interface) of Android Java to something like, xxx_java to indicate that this is a tc for Java extension. Then commit it as the first patch.
2. Commit the test case for native c++ extension as the second patch, which will reuse the current test/echo_extension.c and its test case.

By doing this, we can reduce repetitive test code of c++ and just rename the current Java test case to avoid conflicting.

minggangw:
Follow my previous comment, I submitted the fist patch to rename the js interface. To ensure these changes of renaming don't break the original instrumentation tests, XWalkCoreInternalTest/XWalkCoreTest/XWalkRuntimeClientTest/XWalkRuntimeEmbeddedClinetTest were tested. Please @halton have a look, thanks!

wuhengzhi:
In the commit log, ```echoSync_jave``` -> ```echoSync_java```, ```chaged``` -> ```changed```.

halton:
@minggangw you should open a new PR for the renaming thing and leave this PR for the feature you worked on. Because all the discussion of this PR is about your feature.

minggangw:
Sorry for the wrong spelling, I will open a new PR to do the naming issue.

xwalk_android_tests.gypi
```diff
@@ -248,12 +248,17 @@
       'dependencies': [
         'xwalk_app_runtime_java',
         'xwalk_runtime_client_test_utils_java',
+        'extensions/external_extension_sample.gyp:echo_extension',
```
halton: nit, order plz.

app/tools/android/make_apk.py
```diff
@@ -757,6 +776,13 @@ def main(argv):
           'Mac OS respectively. For example, '
           '--extensions=/path/to/extension1:/path/to/extension2.')
   group.add_option('--extensions', help=info)
+  info = ('The native external extension path. Put the architecture-specific '
+          'library file in the corresponding folder (armeabi-v7a/x86) under '
```
halton: There are more archs: x86-64, arm64 to be supported. I think you need verify whether the .so arch is matched with current app_template.

minggangw:
@halton PR updated, pls have a look, thanks!

app/tools/android/make_apk.py
```diff
@@ -95,6 +106,21 @@ def GetAndroidApiLevel(android_path):
...
+  # the length is 2 bytes.
+  header = library.read(20)
+  return struct.unpack_from('H', header, 18)[0]
+
+def CheckValidationOfExpectedLibraryArch(path, expected_arch):
+  extension_path_list = FindNativeExtensionFiles(path);
+  for extension_path in extension_path_list:
+    if not ConvertArchNameToArchValue(expected_arch) == \
+        GetNativeLibraryArchValueInELFHeader(extension_path):
+      print ('Invalid CPU arch of %s, %s is expected.' % (extension_path,
```
halton: nit, remove space after `print`

minggangw: Ok

halton: Other part LGTM.

halton:
LGTM, merging.

PR #3215, opened 2015-08-25
[Android] Fix the launch screen not working correctly

axinging:
The launch screen dialog has its own lifecycle. The main Activity's attributes
should not be directly applied to the launch screen dialog. So remove these here.
As to XWALK-4688, if you want to apply custom on xwalk_launch_screen, you need
to call window.screen.show(); in your javascript.

BUG=XWALK-4688, XWALK-4687, XWALK-4261

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkLaunchScreenManager.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkLaunchScreenManager.java
@@ -106,11 +106,6 @@ public class XWalkLaunchScreenManager
                 mLaunchScreenDialog = new Dialog(mLibContext,
                                                  android.R.style.Theme_Holo_Light_NoActionBar);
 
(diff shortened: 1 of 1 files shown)


axinging:
@lincsoon  @fujunwei  @halton

lincsoon:
The code LGTM.

But about this,
```
if you want to apply custom on xwalk_launch_screen, you need
to call window.screen.show(); in your javascript.
```
Do we need to give any instruction on the official website?

axinging:
It was here: https://crosswalk-project.org/documentation/manifest/launch_screen.html

lincsoon:
Thanks! Then LGTM

fujunwei:
LGTM.

rakuco:
does this need to be backported?

axinging:
This only happens crosswalk 15. So no need.

rakuco:
You've fixed this in Crosswalk 16.

rakuco:
So does this need to be backported or not?

axinging:
OK, Will do it soon.

PR #3216, opened 2015-08-25
[Android] Implement the API to clear cache for single file.

wuhengzhi:
This patch is to supply an API to clear partial cache.
User could transfer the url to ClearCacheForSingleFile() to clear the
cache for this url.
Add test case for ClearCacheForSingleFile().

BUG=XWALK-4609

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -511,6 +511,19 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
     }
 
     /**
+     * Clear the resource cache. Note that it only clear the cache for the specified
+     * url.
+     * @param url indicate which cache will be cleared.
+     * @since 6.0
+     */
+    @XWalkAPI
+    public void clearCacheForSingleFile(String url) {
+        if (mContent == null) return;
+        checkThreadSafety();
+        mContent.clearCacheForSingleFile(url);
+    }
+
+    /**
      * Indicate that a HTML element is occupying the whole screen.
      * @return true if any HTML element is occupying the whole screen.
      * @since 1.0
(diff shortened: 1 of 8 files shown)


wuhengzhi:
The trybot seems not to be triggered.

rakuco:
This looks like a problem on GitHub's side: according to its status page, it's under another DDoS attack, and in our case it's causing at least the pull request hooks not to be delivered.

wuhengzhi:
@fujunwei Yes, it should be documented.

wuhengzhi:
@lincsoon @halton @axinging PTAL, thanks.
@axinging ```clearCacheForSingleFile()``` should also be wrapped with ```XWalkView.post()```?

wuhengzhi:
ping @lincsoon @halton @axinging, thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -511,6 +511,19 @@ public void clearCache(boolean includeDiskFiles) {
     }
 
     /**
+     * Clear the resource cache. Note that it only clear the cache for the specified
+     * url.
+     * @param url indicate which cache will be cleared.
+     * @since 6.0
+     */
+    @XWalkAPI
+    public void clearCacheForSingleFile(String url) {
```
axinging: If there are two XWalkViews, and load the same url, Will all this url related cache being deleted?

wuhengzhi: The answer is Yes since the resurce was cached by url.

axinging:
LGTM

lincsoon:
LGTM

HimiGame:
usr play three website'ex: http://xxx.com/gamex.html',

i want remove one cache .so ,i use clearCacheForSingleFile, like this:

clearCacheForSingleFile('http://xxx.com/gamex.html'); ???

it's not work ~ how i do ,pls help ~ _ ~

btw:
how set cache mode in crosswalk ?

PR #3220, opened 2015-08-26
[Android] Use new interface of ResourceExtractor on M45

lincsoon:
After rebased to M45, the interface and implementation of
ResourceExtractor changes a log and also invalidates XWalkViewDelegate.
This patch is to restore the functionality with new interface.

BUG=XWALK-4755


lincsoon:
Need to roll the chromium version after https://github.com/crosswalk-project/chromium-crosswalk/pull/270 gets merged

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -37,7 +40,7 @@
     private static final String PRIVATE_DATA_DIRECTORY_SUFFIX = "xwalkcore";
     private static final String[] MANDATORY_PAKS = {
             "xwalk.pak",
-            "en-US.pak",
+            //"en-US.pak",
```
rakuco: Why was this commented out?

lincsoon: Sorry, I forgot to add the comment. Please check the note before MANDATORY_PAKS.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -145,13 +153,73 @@ private static void init(final Context context) {
...
+            int resListId = context.getResources().getIdentifier(
                 XWALK_RESOURCES_LIST_RES_NAME, "array", context.getClass().getPackage().getName());
-        if (resListResId == 0) {
-            resListResId = context.getResources().getIdentifier(
+            if (resListId == 0) {
+                resListId = context.getResources().getIdentifier(
                     XWALK_RESOURCES_LIST_RES_NAME, "array", context.getPackageName());
+            }
+
+            try {
```
rakuco: Not checking if `resListId` is still 0 here makes all the embedded shell test APKs crash immediately after being launched.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -35,9 +38,14 @@
     private static boolean sLibraryLoaded = false;
     private static boolean sRunningOnIA = true;
     private static final String PRIVATE_DATA_DIRECTORY_SUFFIX = "xwalkcore";
+    // In previous versions, MANDATORY_PAKS is actually not mandatory, it's just a list of
```
rakuco: Since we were basically looking for a file we never shipped, I think it's better to just remove it from the list and not add this comment.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
lincsoon: If `resListId` is 0, we will catch `NotFoundException` later and still make the APK crash. Since the resources in this list is mandatory, it's not allowed to go without anyone of them

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
lincsoon: I agree too. But since this file have been in this list for a long time, I think it's better to keep it for the moment and give some comments so that someone else could understand this change.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
rakuco: What I am saying is that it is not being caught. If that was the intention, something else still needs to be done. Just try launching one of the embedded shell APKs and see how they all crash instantly.
FWIW the old version of the code had this check.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
rakuco: Dead code usually rots very quickly. How about just mentioning what happened in the commit message?

rakuco:
Ping

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
lincsoon: Yes, the old code had this check. But if the check faild, it did nothing except logging the error. I don't think that was effective error-handling, because it won't help if the error really happened. In new code, I call `Assert.fail()` for this kind of situations. Because missing the resource file should never happen. And if it really happen, we can notice that right the first time of our develpment.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
lincsoon: Done!

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
rakuco: Then this commit needs more fixes. As I said, this assertion _is_ being hit whenever you launch one of the embedded shell APKs we build for testing, whereas this did not happen before. I can only merge this PR without this regression.

lincsoon:
@rakuco Added additional commit to roll the chromium version

rakuco:
ok, thanks. there's only the error handling vs crash part that needs to be solved before I merge this PR.

lincsoon:
OK. I'll fix the crash error of embedded shell APK ASAP.

lincsoon:
@rakuco Add additional commit to fix the crash of test APKs

lincsoon:
Sorry, although this patch fixed the crash error, but there are still some internal test failed. Please hold on.

rakuco:
Are the tests failing because of the commit, or are you referring to XWALK-4777? Several Android device tests are knowingly timing out in M45.

Also, most importantly, the last commit you added is fairly big, and doesn't seem to be M45-specific. If it fixes an actual bug that's always been present, I'd land it separately in master before the rebase so that it is noted very clearly in the commit history what it is doing (which is why I landed 8cfeae849 and fc6a098f in master, for example).

rakuco:
In case it helps, I've tested this series of patches here:
* XWalkCoreShell launches and works fine.
* XWalkCoreInternalShell launches fine but crashes when I type something in the address bar and touch Go. According to logcat it crashes on `src/ui/base/resource/data_pack.cc:193`.
* XWalkRuntimeClientEmbeddedShell behaves identically to XWalkCoreInternalShell.
* XWalkRuntimeClientShell launches and works fine.

rakuco:
I've rebased the next branch on top of a newer master, which makes the part of the patch dealing with the "animatable" crash unnecessary (it was an unrelated crash).

However, XWalkCoreShell, XWalkCoreInternalShell and XWalkRuntimeClientEmbeddedShell are crashing when I type something and press Go. Basically it's crashing when trying to load the "no internet" error message, so something is still broken with resource loading.

lincsoon:
@rakuco I realized that there are more improvement than fixing in the last commit, so I made a simple version that just fixes the crash error. In my test, both of XWalkCoreShall and XWalkCoreInternal works fine.

Please verify it.

rakuco:
Awesome, thanks. All the shell APKs seem to be behaving correctly now. I had not noticed that the shell/test APKs were organized differently from regular APKs with an embedded Crosswalk.

I tried out parts of the patch in M44 and it also seems to have worked. I'll try to polish it and land what I can in master first to make what we're doing clear.

rakuco:
I've managed to make most of the changes work in master, see #3241.
With that patch applied, the changes required in next are minimal: see my [next-with-3241](https://github.com/rakuco/crosswalk/commits/next-with-3241) branch.

What do you think of landing 3241, then rebasing next on top of master and doing something like what I did in my branch for next?

lincsoon:
@rakuco #3241 LGTM. I think it's a good choice for landing this change. And also thanks for your changes for some details!

rakuco:
Alright, we're finally there then! I've merged #3241 earlier today. I'm then going to land that commit in the branch I mentioned but setting you as the author, and then we can close this PR and focus on the failing tests.

rakuco:
I've just done that. Thank you so much for working on this!

PR #3221, opened 2015-08-27
[Android] Fixed XWalkPreference error before xwalk is ready

lincsoon:
Fixed the error that only last setValue works when calling setValue
multiple times before xwalk is ready.

BUG=XWALK-4868

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
@@ -13,6 +13,7 @@ import android.util.Log;
 
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
+import java.util.Arrays;
 import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


lincsoon:
This PR fixes the remote debugging issue

lincsoon:
@fujunwei @axinging @halton Please help to review

halton:
@lincsoon The code change to me is simple reduce code by introduce `sReservedActions` to substitute sReservedObjects, sReservedClass and sReservedMethods.

I can not understand which part is the key fixing for the issue "only last setValue works when calling setValue multiple times before xwalk is ready."

lincsoon:
The key is in line 51. I make a deep copy of arguments for each reserved methods. The second change is that the actions will be re-invoked exactly in the order that they are reserved. In previous implementation, the order is `create instance` -> `static class` -> `invoke method`

halton:
LGTM, will merge.

PR #3226, opened 2015-08-28
[Android] Fixed XWalkPreference error before xwalk is ready

lincsoon:
Fixed the error that only last setValue works when calling setValue
multiple times before xwalk is ready.

BUG=XWALK-4868
(cherry picked from commit ef202b5a6943a0e28f86074e07ede2ca2d239a0e)


lincsoon:
@rakuco Please help to merge

PR #3230, opened 2015-08-28
[Android] Fix the launch screen not work correctly

axinging:
The launch screen dialog has its own lifecycle. The main Activity's attributes
should not be directly apply to the launch screen dialog. So remove these here.
As to XWALK-4688, if you want to apply custom on xwalk_launch_screen, you need
to call window.screen.show(); in your javascript.

BUG=XWALK-4688, XWALK-4687, XWALK-4261

(cherry picked from commit b43c7cf68b73985130c2b3533fdda7dedfa77aff)


axinging:
@rakuco @lincsoon  @fujunwei  @halton

PR #3231, opened 2015-08-28
Initialize media query "display-mode" features properly

pozdnyakov:
The 'Runtime' class overrides 'GetDisplayMode' method (from the base
'content::WebContentsDelegate' class) so that media query "display-mode"
features are initialized properly.


pozdnyakov:
@kenchris ptal

PR #3235, opened 2015-08-31
Revert "[Android] Add method to config XWalkView render backend"

xzhan96:
This patch will case application crash when using XWalkActivity,
revert it temporarily for a better solution.

This reverts commit f8cb0458c9a59cc5658fd0907a9887ba1ecb9978.


xzhan96:
@axinging @halton @lincsoon

rakuco:
Our bots look fine, I thought the tests there would be crashing according to the description. Do you have more information about what's going on? Is a fix difficult?

xzhan96:
@rakuco yes, if the MainActivity of the test application extends  XWalkActivity instead of Activity, the application will crash while launch. 
I am still finding the root cause, there is a replacement by using Android meta-data, but we think the current solution is better, so what I mean is to find the fix for the crash.

rakuco:
ok, let's revert then

PR #3238, opened 2015-09-02
[Android] Implement XWalkView background color in manifest

axinging:
Add xwalk_view.background_color in manifest. This supports set XWalkView's
background color in manifest.json. This can also avoid white screen on start.
Set XWalkView's background color in manifest.json:
  "xwalk_view": {
    "background_color": "#ff01abcd"
  }
Note: Color is in Alpha/Red/Green/Blue order.
BUG=XWALK-4809,XWALK-4963

### DIFF
--- a/runtime/browser/android/xwalk_content.cc
+++ b/runtime/browser/android/xwalk_content.cc
@@ -19,6 +19,7 @@
 #include "base/path_service.h"
 #include "base/pickle.h"
 #include "base/prefs/pref_service.h"
+#include "base/strings/string_number_conversions.h"
 #include "components/autofill/content/browser/content_autofill_driver_factory.h"
 #include "components/autofill/core/browser/autofill_manager.h"
 #include "components/navigation_interception/intercept_navigation_delegate.h"
@@ -467,6 +468,22 @@ jboolean XWalkContent::SetManifest(JNIEnv* env,
     // No need to display launch screen, load the url directly.
     Java_XWalkContent_onGetUrlFromManifest(env, obj, url_buffer.obj());
   }
+  std::string view_background_color;
+  ManifestGetString(manifest,
+                    keys::kXWalkViewBackgroundColor,
+                    keys::kViewBackgroundColor,
+                    &view_background_color);
+
+  if (view_background_color.empty())
+    return true;
+  unsigned int view_background_color_int = 0;
+  if (!base::HexStringToUInt(view_background_color.substr(1),
+      &view_background_color_int)) {
+    LOG(ERROR) << "Background color format error! Valid background color"
+               "should be(Alpha Red Green Blue): #ff01abcd";
+    return false;
+  }
+  Java_XWalkContent_setBackgroundColor(env, obj, view_background_color_int);
   return true;
 }
 
--- a/application/common/application_manifest_constants.cc
+++ b/application/common/application_manifest_constants.cc
@@ -46,6 +46,12 @@ const char kLaunchScreenPortrait[] =
     "launch_screen.portrait";
 const char kLaunchScreenReadyWhen[] =
     "launch_screen.ready_when";
+const char kView[] = "view";
+const char kViewBackgroundColor[] = "view.background_color";
+
+// XWalk View Manifest extensions:
+const char kXWalkView[] = "xwalk_view";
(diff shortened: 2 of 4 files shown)


axinging:
@lincsoon  @fujunwei  @halton

axinging:
@lincsoon @fujunwei @halton

runtime/base/strings/string_number_conversions.cc
```diff
@@ -0,0 +1,238 @@
...
+#include <ctype.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <limits>
+
+namespace xwalk {
+
+namespace {
+
+
```
halton: nit, 1 empty line

runtime/base/strings/string_number_conversions.cc
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/runtime/base/strings/string_number_conversions.h"
```
halton: Why copying the code in `src/base/strings/string_number_conversions.cc` instead of re-use it?

runtime/browser/android/xwalk_content.cc
```diff
@@ -452,6 +453,23 @@ jboolean XWalkContent::SetManifest(JNIEnv* env,
...
+    if (!HexStringWithPrefixToInt(view_background_color, "#",
+        &view_background_color_int)) {
+      LOG(ERROR) << "Background color format error! Valid background color"
+                 "should be: #01abcd, 01abcd";
+      return false;
+    }
+    Java_XWalkContent_setBackgroundColor(env, obj, view_background_color_int);
+  }
+
   return true;
```
halton:
```
if (view_background_color.empty())
  return true;

int view_background_color_int = 0;
...
```

runtime/base/strings/string_number_conversions.cc
axinging: Most classes like BaseCharToDigit are defined in base/strings/string_number_conversions.cc instead of base/strings/string_number_conversions.h.
If re-use it, we have to include this  .cc.

axinging:
@lincsoon @fujunwei @halton

runtime/browser/android/xwalk_content.cc
```diff
@@ -467,6 +468,22 @@ jboolean XWalkContent::SetManifest(JNIEnv* env,
...
+  std::string view_background_color;
+  ManifestGetString(manifest,
+                    keys::kXWalkViewBackgroundColor,
+                    keys::kViewBackgroundColor,
+                    &view_background_color);
+
+  if (view_background_color.empty())
+    return true;
+  unsigned int view_background_color_int = 0;
+  if (!base::HexStringToUInt(view_background_color.substr(1),
```
lincsoon: Is it possible that the user didn't prefix the color value with `#`?

axinging: In the "xwalk_launch_screen", it use color with prefix "#". We do not want to re-invent a wheel.

axinging:
@fujunwei , but we still hope the use knows that when HexStringToUInt return false. This means something is wrong with the json file.

fujunwei:
I mean the featuring code after this section will can't be executed if the background color is empty.

fujunwei:
LGTM

runtime/browser/android/xwalk_content.cc
```diff
@@ -467,6 +468,22 @@ jboolean XWalkContent::SetManifest(JNIEnv* env,
...
     Java_XWalkContent_onGetUrlFromManifest(env, obj, url_buffer.obj());
   }
+  std::string view_background_color;
+  ManifestGetString(manifest,
+                    keys::kXWalkViewBackgroundColor,
+                    keys::kViewBackgroundColor,
+                    &view_background_color);
+
+  if (view_background_color.empty())
+    return true;
```
lincsoon: I agree with Junwei. If anyone else add another feature after here, the code won't be executed

lincsoon:
LGTM

PR #3239, opened 2015-09-02
Enable LZMA function at build time.

d0ngjun:
Besides the runtime build option provided in make_apk.py, we enable lzma
at build time also.

Due to the architecture issue, we need involve dummy library at
build time.

BUG=XWALK-4947

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -301,13 +301,7 @@ def Customize(options, app_info, manifest):
                options.xwalk_command_line, options.compressor)
 
 
-def CleanCompressedLibrary(library_path, arch):
-  useless = os.path.join(library_path, NATIVE_LIBRARY + '.' + arch)
-  if os.path.isfile(useless):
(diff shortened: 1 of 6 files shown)


d0ngjun:
@halton @heke123 @PeterWangIntel, please help review.

halton:
@PeterWangIntel is the trybot failure related?

halton:
@d0ngjun since the XWALK-4981 is resolved, could you re-push your change to trigger trybot.

halton:
LGTM, merging

suleymanccelik:
Can I add --enable-lzma command to xwalk-command-line file under asset directory in my android application instead of using make_apk.py?

heke123:
@suleymanccelik,   no,  the "--enable-lzma" and "-proguard" options are only for make_apk.py.
I guess you care about the final apk size? we have another version as "crosswalk-lite", in which we cut off some functions in order to reduce size. the final size is about half of the crosswalk version.

Issue #320, opened 2013-08-08
System Settings API doesn't work on Tizen 2.1 PR3


jinxian-intel:
BUG DETAILED DESCRIPTIONS
System Settings API doesn't work
Here is some failed cases and error message below:
Case name:
SystemSettingManager_getProperty_type_TypeMismatch.html
Error message:
assert_throws: TypeMismatchError should be thrown - given incorrect SystemSettingType - undefined function "function () {
                tizen.systemsetting.getProp..." threw object "TypeError: Cannot call method 'getProperty' of undefined" ("TypeError") expected object "[object Object]" ("TypeMismatchError")

Case name:
SystemSettingManager_getProperty_LOCK_SCREEN.html
Error message:
Cannot call method 'getProperty' of undefined(stack: TypeError: Cannot call method 'getProperty' of undefined
    at Object.&lt;anonymous&gt; (file:///opt/tct-systemsetting-tizen-tests/systemsetting/SystemSettingManager_getProperty_LOCK_SCREEN.html:56:41)
    at Object.Test.step (file:///opt/tct-systemsetting-tizen-tests/resources/testharness.js:1121:25)
    at file:///opt/tct-systemsetting-tizen-tests/systemsetting/SystemSettingManager_getProperty_LOCK_SCREEN.html:46:3)

Case name:
SystemSettingManager_setProperty_LOCK_SCREEN.html
Error message:
Cannot call method 'getProperty' of undefined(stack: TypeError: Cannot call method 'getProperty' of undefined
    at Object.&lt;anonymous&gt; (file:///opt/tct-systemsetting-tizen-tests/systemsetting/SystemSettingManager_setProperty_LOCK_SCREEN.html:95:25)
    at Object.Test.step (file:///opt/tct-systemsetting-tizen-tests/resources/testharness.js:1121:25)
    at file:///opt/tct-systemsetting-tizen-tests/systemsetting/SystemSettingManager_setProperty_LOCK_SCREEN.html:50:3)

Environment:
Tizen 2.1: Build 23

EXACT STEPS LEADING TO PROBLEM:
1) Install crosswalk https://github.com/otcshare/crosswalk/releases/download/1.28.1.0/crosswalk-1.28.1.0-0.i586.rpm.zip
2) Install tizen extension crosswalk: https://github.com/otcshare/tizen-extensions-crosswalk/releases/download/0.1/tizen-extensions-crosswalk-0.0.1-0.i586.rpm.zip
[...]

PR #3240, opened 2015-09-03
Fix build on Windows.

darktears:
And most likely on Mac as well.

https://github.com/crosswalk-project/crosswalk/pull/320 broke the
build by using classes not added to the build.

Fix the build by adding the new class from #320 to the build but it's
most likely that the Download feature doesn't yet work on Windows
because there are still code guarded in #ifdef in the original
commit. I've created XWALK-4969 to enable it which shouldn't require
much work.


darktears:
@XiaosongWei

mrunalk:
LGTM

PR #3241, opened 2015-09-04
[Android] Simplify the resource interception code in XWalkViewDelegate.

rakuco:
There are two related changes in this patch with the same goal of paving the way for the M45 rebase, where the upstream code handling the loading of Android resources has changed. Most of what we had to do there also applies to M44, so the patch is being landed in master in order to make it clear in the commit history what is being done and why.

1. Make the code easier to understand.
   * Set up the resource interception infrastructure in a separate
     function.
   * Also have a separate function for determining a resource ID after
     XWALK-3469 made the code a bit harder to follow.
   * Have variables with clear names such as `isSharedMode` and
     `isTestApk` to indicate our intents and what is being checked. We
     were previously relying on package name checks or ID comparisons
     that did not make any sense to anyone who was not already familiar
     with the code base.
   * Add more comments to the code to indicate what we are doing.
2. Functional changes.
   * Stop conditionally calling setResourceIntercepter(). The previous
     check basically skipped our test APKs. Since intercepting resources
     there does not cause any harm, do so to make the code easier to
     both write and understand.
   * Remove "en-US.pak" from the list of mandatory PAK files. It has
     never actually existed in our Android builds, and since we now use
     `MANDATORY_PAK_FILES` to determine which resources need to be
     intercepted, we cannot have a non-existent file there as we would
     end up trying to open it.

Most of the code was actually written by @lincsoon in pull request #3220, I only picked as much as I could into master.

Related to: XWALK-4755

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
@@ -9,6 +9,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.IOException;
 import java.lang.StringBuilder;
+import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


rakuco:
ping @lincsoon 
As I said in #3220, my idea was to land as much of the code there as possible into master to make the final _next_ patch smaller as well as make it clear that this change is important.
Please take a look.

lincsoon:
LGTM

PR #3242, opened 2015-09-04
Roll GYP forward so it supports compiling managed code on Windows.

darktears:
This is in preparation of landing the dotnet extensions for Crosswalk
on Windows. At some point GYP will naturally roll into Chromium
so we may get it around M47.

GYP CL : https://codereview.chromium.org/1331463002

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -50,6 +50,10 @@ solutions = [
       'src/v8':
         crosswalk_git + '/v8-crosswalk.git@' + v8_crosswalk_rev,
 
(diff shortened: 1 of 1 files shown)


darktears:
@rakuco

darktears:
This like some fixes were made since the last roll of GYP. I need to figure that out.

darktears:
So GYP fails on Android because two files are named the same. The fix was landed upstream a while back. It should be part of M45 so I won't push for this patch, I'll just wait M45.

PR #3247, opened 2015-09-09
[Android] Fix the error that the screen flashes at startup

lincsoon:
The moment when the XWalkViewBridge is added to the XWalkView, the
screen flashes black. The reason is when the SurfaceView appears
in the window the fist time, it requests the window's parameters
changing by calling IWindowSession.relayout(). But if the window
already has appropriate parameters, it will not refresh all the
window's stuff and the screen will not blink. So we add a 0px
SurfaceView at first. This will recreate the window before the
activity is shown on the screen, and when the actual XWalkViewBridge
is added, it will just continue to use the window with current
parameters. The temporary SurfaceView can be removed at last.

BUG=XWALK-4966

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -20,6 +20,7 @@ import android.provider.MediaStore;
 import android.util.AttributeSet;
 import android.util.Log;
 import android.view.KeyEvent;
+import android.view.SurfaceView;
 import android.view.ViewGroup;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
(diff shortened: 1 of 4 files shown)


lincsoon:
@axinging @fujunwei

lincsoon:
@axinging Added another commit. Please review

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -185,17 +186,29 @@ public void onActivityStateChange(Activity activity, int newState) {
...
     /**
      * Constructs a new XWalkView with a Context object.
      * @param context a Context object used to access application assets.
      * @since 6.0
      */
     @XWalkAPI(preWrapperLines = {
-                  "        super(${param1}, null);"},
+                  "        super(${param1}, null);",
+                  "        SurfaceView surfaceView = new SurfaceView(${param1});",
+                  "        surfaceView.setLayoutParams(new ViewGroup.LayoutParams(0, 0));",
```
axinging: No need to set the color of SurfaceView?

lincsoon: Because we don't know the actual background color, it's not appropriate to set the color to white here(the home page maybe red). The develop have to set the background color via styles.xml

axinging:
LGTM

PR #3248, opened 2015-09-09
[Android] Update the comment about shared mode

lincsoon:
Make it clear that the developer can use XWalkView with the old way. But
to take the advantage of shared mode, the XWalkActivity is necessary.


lincsoon:
@axinging @fujunwei

PR #3254, opened 2015-09-10
Port the update about shared mode from master

lincsoon:
2 commits are to fix the critical bug XWALK-4966. Another 2 commits are to update the API doc.


lincsoon:
@rakuco Please help to review and merge

PR #3256, opened 2015-09-10
Drop our custom upstream_revision.py script.

rakuco:
Starting with M45 and http://crrev.com/1197483002, it is possible to
pass the "--git-svn-go-deeper" option to Chromium's lastchange.py script
so that it goes through git history, finds the first git-svn commit
there and notes the associated SVN revision.

This is exactly what we were doing before manually with the
|blink_upstream_rev| variable in DEPS.xwalk.

Replace our hand-rolled solution with the upstream script we can now
use.


rakuco:
@darktears @kenchris

darktears:
YEAH LGTM

PR #3258, opened 2015-09-10
gyp: Use repack_action.gypi.

rakuco:
With M45, we can use upstream's newly-introduced repack_action.gypi to
simplify our work: it takes care of all the inputs/action/output
boilerplate.

Upstream CL: https://codereview.chromium.org/216013003


darktears:
LGTM.

PR #3260, opened 2015-09-11
Test case for Presentation API

RafuCater:
Override displayAvailableTest.html and add remote.html 
Change presentation_api.js  in app/android/runtime_client_shell/res/raw/


jondong:
@RafuCater in the commit log, use past tense instead of present tense; introduce a little bit more about the background of your patch: why you need to update the test case; the usage of the newly introduced file; why you need to replace the presentation_api.js, etc.

PR #3261, opened 2015-09-11
gyp: Make xwalk_runtime depend on xwalk_pak.

rakuco:
The runtime code actually needs xwalk.pak to be present to work. This is
evident when building a target like xwalk_extensions_browsertest from
scratch -- it depends on xwalk_runtime, but not xwalk_pak, and tests
InternalExtensionTest.InternalExtension and
XWalkExtensionsIFrameTest.CrossContextsReferenceShouldNotCrash were
failing when xwalk.pak was absent.

BUG=XWALK-4776

### DIFF
--- a/xwalk.gyp
+++ b/xwalk.gyp
@@ -70,6 +70,7 @@
         '../v8/tools/gyp/v8.gyp:v8',
         'generate_upstream_blink_version',
         'xwalk_application_lib',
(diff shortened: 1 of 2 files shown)


rakuco:
ping @darktears

darktears:
LGTM.

PR #3269, opened 2015-09-16
[Android] Refine the accept language from OS.

wuhengzhi:
The language's format from OS sometimes is like as "zh_CN", but the
adopted format by HTTP protocol is "zh-cn".
Replace "_" with "-" and change the alphabet to lower case.

BUG=XWALK-5045


wuhengzhi:
@axinging @lincsoon PTAL, thanks.

lincsoon:
LGTM

PR #3274, opened 2015-09-21
Fixed the bulk data transmission crash on Android

jondong:
This issue is introduced by patch e5d3b6c2, the solution only works
for Windows but crashes Android. Use defined(OS_WIN) macro to enable
it only for Windows and restore the old implementation on other
platforms.

BUG=XWALK-5029

### DIFF
--- a/extensions/common/xwalk_extension_server.cc
+++ b/extensions/common/xwalk_extension_server.cc
@@ -24,8 +24,7 @@ namespace extensions {
 const size_t kInlineMessageMaxSize = 256 * 1024;
 
 XWalkExtensionServer::XWalkExtensionServer()
-    : sender_(NULL),
-      renderer_process_pid_(0),
+    : channel_proxy_(NULL),
       permissions_delegate_(NULL) {}
 
 XWalkExtensionServer::~XWalkExtensionServer() {
@@ -51,10 +50,6 @@ bool XWalkExtensionServer::OnMessageReceived(const IPC::Message& message) {
   IPC_END_MESSAGE_MAP()
 
   return handled;
-}
-
-void XWalkExtensionServer::OnChannelConnected(int32 peer_pid) {
-  renderer_process_pid_ = peer_pid;
 }
 
 void XWalkExtensionServer::OnCreateInstance(int64_t instance_id,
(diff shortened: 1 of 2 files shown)


jondong:
@halton PTAL

jondong:
@darktears PTAL. I don't have the windows dev environment, would you please help me to verify? Thanks

darktears:
I can verify however I'm still not convinced the fix is correct. There is no need for special cases other platforms as it appears it is not working on Android only. I've tested the previous patch on Linux.

Also can you share mode details on what's is going on on Android? OpenWithExtraPrivileges fails?

jondong:
@darktears by my investigation OpenWithExtraPrivileges will call Open instead on Android, and the renderer process cannot be opened more than once, or we'll get invalid process handler. I kept the OpenWithExtraPrivileges call but moved it into OnChannelConnected instead of OnCreateInstance, which can avoid opening the process for multiple times. The code has been updated, PTAL.

darktears:
The last change is basically reverting https://github.com/crosswalk-project/crosswalk/commit/e5d3b6c2bc1251c979ab525a92d42e3341a1d25b. It's going to break Windows again. What about opening the process once in PostMessageToJSCallback and keep the handle for later usage?

jondong:
@darktears updated to address your suggestion, PTAL, thanks

halton:
@darktears could you prioritize your time to review this PR? This PR is demanded for Realsense work.

rakuco:
I think he's busy with [ViewSource|https://viewsourceconf.org] until the end of today.

darktears:
@halton will test the patch today.

darktears:
The patch breaks Windows again.

jondong:
thanks for helping test on Windows @darktears , would you please to elaborate is it break the windows build or test case on window?

darktears:
Sure I did not have time to debug further yesterday. But I have more details. 

So right now on Windows it fails to call if GiveReadOnlyToProcess(renderer_process_handle_, &handle) even the first time when  renderer_process_handle_ is getting set for the first time. The call to OpenWithExtraPrivileges works fine and GetPeerPID() returns the same PID as before. If I remove the check if (!renderer_process_handle_) and call the code inside the "if" everytime it works (which I know it doesn't work on Android) which shows to me that the Process object needs to be alive for the Handle to be valid. Here is the documentation for Process::Handle that we use : 

// Returns a handle for this process. There is no guarantee about when that
// handle becomes invalid because this object retains ownership.
ProcessHandle Handle() const;

Sounds to me that we can't rely on the Handle because we get it from Process but it gets destroyed. 

Would this works on Android? https://gist.github.com/darktears/049c35ffb05b7bfd8ff3

Now comes the question on why it's crashing on Android to open several times the Process?

jondong:
@darktears thanks for the elaboration, by my investigation actually we can open the Process for multiple times, the crash on Android is because we retrieve the wrong PID within OnChannelConnected, we'd borrow it from the the channel proxy which get passed when initialise. I bet it won't crash windows now ;)

darktears:
LGTM but there is a little style fix. I like that version better :).

jondong:
Thanks @darktears , updated.

PR #3276, opened 2015-09-21
[Android] PAC does not work with Crosswalk

marcuspridham:
Crosswalk does not work with the Proxy Auto-Config setting set.  There is some changes from https://codereview.chromium.org/742023003 that need to be applied to Crosswalk.  I have verified this change against a Nexus 7 (5.1.1) that has my corporate PAC set.  I also checked with a Nexus 5 (4.4.4)  that has a manual proxy set still works.

BUG=XWALK-5094

### DIFF
--- a/runtime/browser/runtime_url_request_context_getter.cc
+++ b/runtime/browser/runtime_url_request_context_getter.cc
@@ -46,6 +46,7 @@
 #include "xwalk/runtime/common/xwalk_switches.h"
 
 #if defined(OS_ANDROID)
+#include "net/proxy/proxy_config_service_android.h"
 #include "xwalk/runtime/browser/android/cookie_manager.h"
 #include "xwalk/runtime/browser/android/net/android_protocol_handler.h"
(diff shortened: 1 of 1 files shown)


rakuco:
The change looks sane to me. Ping @lincsoon, @halton

runtime/browser/runtime_url_request_context_getter.cc
```diff
@@ -51,6 +51,7 @@
 #include "xwalk/runtime/browser/android/net/url_constants.h"
 #include "xwalk/runtime/browser/android/net/xwalk_url_request_job_factory.h"
 #include "xwalk/runtime/browser/android/xwalk_request_interceptor.h"
+#include "net/proxy/proxy_config_service_android.h"
```
halton: please order, lgtm with this nit comment.

marcuspridham:
Fixed include order in latest commit. Thanks for the review.

halton:
Thanks for quick rework, will merge it after trybot get green.

PR #3278, opened 2015-09-22
[Android] Add silent download support in shared mode

XiaosongWei:
Shared mode APK can be configured to download the runtime APK in
background silently via a meta-data element "xwalk_silent_download"
in AndroidManifest.xml.

The downloaded runtime APK will be not installed as a normal application
instead, its core components (libxwalkcore.so, classes.dex, xwalk.pak
and icudtl.dat) will be extracted to application's private data
directory and the core library will be loaded as a plugin.

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
@@ -5,12 +5,17 @@
 package org.xwalk.core;
 
 import android.content.Context;
+import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.pm.Signature;
+import android.os.Build;
+import android.os.Environment;
 import android.util.Log;
+import dalvik.system.DexClassLoader;
 
+import java.io.File;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
@@ -27,7 +32,9 @@ class XWalkCoreWrapper {
     private static final String XWALK_APK_PACKAGE = "org.xwalk.core";
     private static final String WRAPPER_PACKAGE = "org.xwalk.core";
     private static final String BRIDGE_PACKAGE = "org.xwalk.core.internal";
+    private static final String PRIVATE_LIB_DIR = "xwalkcorelib";
     private static final String TAG = "XWalkLib";
+    private static final String META_XWALK_SILENT_DOWNLOAD = "xwalk_silent_download";
 
     private static XWalkCoreWrapper sProvisionalInstance;
     private static XWalkCoreWrapper sInstance;
@@ -150,14 +157,19 @@ class XWalkCoreWrapper {
         Log.d(TAG, "Attach xwalk core");
         sProvisionalInstance = new XWalkCoreWrapper(context, -1);
         if (!sProvisionalInstance.findEmbeddedCore()) {
-            int status = sProvisionalInstance.mCoreStatus;
+            if(sProvisionalInstance.isDownloadMode()) {
+                sProvisionalInstance.findDownloadedCore();
+            } else {
+                int status = sProvisionalInstance.mCoreStatus;
 
-            if (!sProvisionalInstance.findSharedCore()) {
-                if (sProvisionalInstance.mCoreStatus != XWalkLibraryInterface.STATUS_NOT_FOUND) {
-                    status = sProvisionalInstance.mCoreStatus;
+                if (!sProvisionalInstance.findSharedCore()) {
+                    if (sProvisionalInstance.mCoreStatus !=
+                            XWalkLibraryInterface.STATUS_NOT_FOUND) {
+                        status = sProvisionalInstance.mCoreStatus;
+                    }
+                    Log.d(TAG, "core status: " + status);
+                    return status;
                 }
-                Log.d(TAG, "core status: " + status);
-                return status;
             }
         }
         return sProvisionalInstance.mCoreStatus;
@@ -199,6 +211,21 @@ class XWalkCoreWrapper {
                 minApiVersion : mApiVersion;
         mCoreStatus = XWalkLibraryInterface.STATUS_NOT_FOUND;
         mWrapperContext = context;
+    }
+
+    private boolean isDownloadMode() {
+        try {
+            PackageManager packageManager = mWrapperContext.getPackageManager();
+            ApplicationInfo appInfo = packageManager.getApplicationInfo(
+                    mWrapperContext.getPackageName(), PackageManager.GET_META_DATA);
+            String download = appInfo.metaData.getString(META_XWALK_SILENT_DOWNLOAD);
+            if (download != null && download.equalsIgnoreCase("enable")) {
+                Log.i(TAG, "Running in silent download mode");
+                return true;
+            }
+        } catch (NameNotFoundException | NullPointerException e) {
+        }
+        return false;
     }
 
     private void initCoreBridge() {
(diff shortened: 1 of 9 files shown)


XiaosongWei:
@lincsoon The download mode is now combined with shared mode, please take a look.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -150,14 +157,19 @@ public static int attachXWalkCore(Context context) {
         Log.d(TAG, "Attach xwalk core");
         sProvisionalInstance = new XWalkCoreWrapper(context, -1);
         if (!sProvisionalInstance.findEmbeddedCore()) {
-            int status = sProvisionalInstance.mCoreStatus;
+            if(sProvisionalInstance.isDownloadMode()) {
```
lincsoon: There's no need to differentiate download mode. Check embedded mode -> download mode -> shared mode in sequence.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -201,6 +213,21 @@ private XWalkCoreWrapper(Context context, int minApiVersion) {
         mWrapperContext = context;
     }
 
+    private boolean isDownloadMode() {
+        try {
+            PackageManager packageManager = mWrapperContext.getPackageManager();
+            ApplicationInfo appInfo = packageManager.getApplicationInfo(
+                    mWrapperContext.getPackageName(), PackageManager.GET_META_DATA);
```
lincsoon: For shared mode, the developer doesn't need to specify it specially, just link to the shared library. For download mode, it should be the same, we don't want to force the developer to do extra settings.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -244,6 +271,24 @@ private boolean findSharedCore() {
...
+    private boolean findDownloadedCore() {
+        String libDir = mWrapperContext.getDir(PRIVATE_LIB_DIR, Context.MODE_PRIVATE).toString();
+        String dexPath = libDir + File.separator + "classes.dex";
+        String dexOutputPath = mWrapperContext.getDir("dex", 0).getAbsolutePath();
+        ClassLoader localClassLoader = ClassLoader.getSystemClassLoader();
+        mBridgeLoader = new DexClassLoader(dexPath, dexOutputPath, libDir, localClassLoader);
+
+        if (!checkCoreVersion() || !checkCoreArchitecture()) {
+            mBridgeLoader = null;
+            mCoreStatus = XWalkLibraryInterface.STATUS_INCOMPLETE_LIBRARY;
```
lincsoon: It's incorrect to set the status here. The error status has already been defined by `checkCoreVersion` or `checkCoreArchitecture`. This will overwrite the real status.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -244,6 +271,24 @@ private boolean findSharedCore() {
         return true;
     }
 
+    private boolean findDownloadedCore() {
+        String libDir = mWrapperContext.getDir(PRIVATE_LIB_DIR, Context.MODE_PRIVATE).toString();
+        String dexPath = libDir + File.separator + "classes.dex";
+        String dexOutputPath = mWrapperContext.getDir("dex", 0).getAbsolutePath();
```
lincsoon: Why don't use Context.MODE_PRIVATE explicitly here?

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
```diff
@@ -92,17 +150,36 @@
...
+    private static final String META_XWALK_APK_URL = "xwalk_apk_url";
+    private static final String META_XWALK_SILENT_DOWNLOAD = "xwalk_silent_download";
+
+    private static final String PRIVATE_LIB_DIR = "xwalkcorelib";
+
+    private static final String[] XWALK_LIB_RESOURCES = {
+        "libxwalkcore.so",
+        "icudtl.dat",
+        "xwalk.pak",
+        "classes.dex",
```
lincsoon: Where does this file come from?

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
```diff
@@ -148,42 +245,50 @@ public void run() {
...
 
-    /**
-     * Set the download URL of the Crosswalk runtime. By default, the updater will get the URL from
-     * the Android manifest.
-     *
-     * @param url The download URL.
-     */
-    public void setXWalkApkUrl(String url) {
-        mXWalkApkUrl = url;
+    private boolean isDownloadMode() {
```
lincsoon: Ditto, no need to add extra settings in the manifest

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
```diff
@@ -240,19 +371,75 @@ public void onDownloadCancelled() {
...
         public void onDownloadFailed(int status, int error) {
-            mDialogManager.dismissDialog();
-            mDialogManager.showDownloadError(status, error, mCancelCommand, mDownloadCommand);
+            mIsDownloading = false;
+            mUpdateListener.onXWalkUpdateFailed();
         }
+
+        @Override
+        public void onDownloadCompleted(Uri uri) {
+            if(!installLibraries(uri.getPath())) Assert.fail();
```
lincsoon: No need to copy the so file to private directory?

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
```diff
@@ -92,17 +150,36 @@
...
+    private static final String META_XWALK_SILENT_DOWNLOAD = "xwalk_silent_download";
+
+    private static final String PRIVATE_LIB_DIR = "xwalkcorelib";
+
+    private static final String[] XWALK_LIB_RESOURCES = {
+        "libxwalkcore.so",
+        "icudtl.dat",
+        "xwalk.pak",
+        "classes.dex",
+    };
```
lincsoon: How about this:

```
private static final String[] RESOURCES_TO_EXTRACT = {
    "libs/PLATFORM/libxwalkcore.so",
    "assets/icudtl.dat",
    "assets/xwalk.pak",
    "classes.dex",
};
```

We can specify the full paths here, and then extract the file name from this path, in order to avoid to check the file extensions later(not safe).

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
XiaosongWei: What if the check fails in embedded/download modes but success in shared mode and the user actually wanted to use their specified XWalkRuntimeLib.apk downloaded from their hosts?

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
XiaosongWei: Per the API doc:
Use 0 or MODE_PRIVATE for the default operation, MODE_WORLD_READABLE and MODE_WORLD_WRITEABLE to control permissions.

Yet, it would be good to use MODE_PRIVATE explicitly.

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
XiaosongWei: This is needed to differentiate the silent-download and explicit-download-installation since we need to make updateXWalkRuntime() works as expected in these two cases.

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
XiaosongWei: make sense. then only check libxwalkcore.so and replace PLATFORM with the real one.

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
```diff
@@ -148,42 +245,50 @@ public void run() {
...
-
-    /**
-     * Dismiss the dialog showing and waiting for user's input.
-     *
-     * @return Return false if no dialog is being displayed, true if dismissed the showing dialog.
-     */
-    public boolean dismissDialog() {
-        if (!mDialogManager.isShowingDialog()) return false;
-        mDialogManager.dismissDialog();
+        if (isDownloadMode()) {
```
lincsoon: In shared mode, the developer doesn't have to implement the callback method `onXWalkUpdateFailed()`, because there will be an alert dialog. But in download mode, the developer must handle the error. Which means that we can't guarantee that same code can apply to all modes. So it doesn't make sense to provide only one interface `updateXWalkRuntime`. We can provide another interface `updateXWalkRuntimeSilently` for the download mode

PR #3281, opened 2015-09-23
[Android] Implement expansion mode

r0b5t4:
Crosswalk expansion mode allows to load the runtime from an android
expansion file. This is a middle-ground between bundled mode and
shared mode, in that it lets app developers ship the entire app
from their own play store account, but does not count the runtime
size against the 50MB APK size limit.

This is the core functionality for expansion mode, and works for
crosswalk embedding apps that implement expansion file management
themselves. Tools support will be implemented in a follow-up patch.

How to use expansion mode:
1 - Create crosswalk embedding app
2 - Add AndroidManifest metadata named org.xwalk.expansionFile with
    value "main" or "patch" depending on which expansion file to use.
3 - When app starts, crosswalk is loaded from corresponding
    expansion file.
4 - App needs to be aware that expansion file gets ending ".apk"
    appended, so it can be used with DexClassLoader.
5 - App needs to ship assets/icudtl.dat, this will be fixed in the
    future.

BUG=XWALK-4726

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
@@ -151,7 +151,19 @@ class XWalkCoreWrapper {
         Log.d(TAG, "Attach xwalk core");
         sProvisionalInstance = new XWalkCoreWrapper(context, -1);
         if (!sProvisionalInstance.findEmbeddedCore()) {
-            sProvisionalInstance.findSharedCore();
+            int status = sProvisionalInstance.mCoreStatus;
+
+            if (sProvisionalInstance.findExpansionCore()) {
+                return sProvisionalInstance.mCoreStatus;
+            }
+
+            if (!sProvisionalInstance.findSharedCore()) {
+                if (sProvisionalInstance.mCoreStatus != XWalkLibraryInterface.STATUS_NOT_FOUND) {
+                    status = sProvisionalInstance.mCoreStatus;
+                }
+                Log.d(TAG, "core status: " + status);
+                return status;
+            }
         }
         return sProvisionalInstance.mCoreStatus;
     }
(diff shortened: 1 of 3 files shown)


r0b5t4:
@lincsoon @wuhengzhi

runtime/android/core/src/org/xwalk/core/XWalkExpansionContext.java
```diff
@@ -0,0 +1,312 @@
...
+     * Extract entry from the expansion file.
+     * @param expFile Expansion file
+     * @param entryName Name of entry
+     * @return true on success.
+     */
+    private boolean extractEntry(File expFile, String entryName) {
+
+        JarFile jar;
+        try {
+            jar = new JarFile(expFile);
```
XiaosongWei: Why not just use ZipFile to extract the entry as the expFile is actually a Zip file?

r0b5t4:
@XiaosongWei changed it.

@lincsoon @wuhengzhi @halton

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -96,8 +96,10 @@ public static void init(Context libContext, Context appContext) {
 
         try {
             if (libContext == null) {
-                init(appContext);
+                Log.d(TAG, "Initializing plain context");
+                init(libContext);
```
XiaosongWei: typo? it shall be appContext here. Also, please remove the debug messages and keep the change as clean as possible. Thanks.

runtime/android/core/src/org/xwalk/core/XWalkCoreWrapper.java
```diff
@@ -222,6 +234,25 @@ private boolean findEmbeddedCore() {
         return true;
     }
 
+    private boolean findExpansionCore() {
+        mBridgeContext = new XWalkExpansionContext(mWrapperContext);
```
XiaosongWei: Can you please double check if it works in Android older than JELLY_BEAN_MR1? Since you assign a non-null mBridgeContext, checkCoreArchitecture() -> XWalkViewDelegate.loadXWalkLibrary() will try to load a libxwalkcore.so from /data/data/" + mBridgeContext.getPackageName() + "/lib"  if Android version < JELLY_BEAN_MR1

r0b5t4:
Will rebase on top of silent download. Closing.

PR #3282, opened 2015-09-23
Implement .NET external extension system.

darktears:
This commit adds support to load external .NET extensions
for 3rd party developers.

You can specify the path of the extensions to load using
--external-extension-path=/path/to/extension

The extensions are expected to be a .DLL constructed with
Visual Studio as a C# Library (and thus multiarch). Alongside
with that library the 3rd party developer needs to include a .NET bridge
named myextension_bridge.dll given the .NET extension is called
myextension.dll. The dotnet bridge is shipped part of Crosswalk
and is called xwalk_dotnet_bridge and it's a simple copy/rename
for the developer.

This .NET library needs to contains two classes XWalkExtension and
XWalkExtensionInstance which are going to be invoked by Crosswalk
to load the extensions and call the various methods on each instances
to make communication between .NET and JS. .NET extensions are a bit
higher level than our C or C++ extensions because the .NET extension
system will automatically call the constructors in .NET when needed :
e.g. creating the XWalkExtension instance and then create the various
XWalkExtensionInstance instances. This is following inspiration of
what's inside tizen-extension-crosswalk common/ directory which is used
to abstract the low levelness to who write t-e-c extensions.

```
 _______       _______       ______________
| xwalk | <-> | bridge| <-> |.NET extension|
|_______|     |_______|     |______________|
```
The bridge is loaded dynamically like an external extension would
be loaded : it's a DLL and it contains a single exported symbol : XW_Initialize.

One bridge instance is used by each extension and it is responsible to load
the .NET extension, verify using .NET reflection that the entry points
(e.g. XWalkExtension constructor, XWalkInstance handleMessage method...)
are present and that the .NET extension can be used. It then handles the
communication back and forth with Crosswalk and the extension.

One implementation detail worth to note is the usage of the dynamic parameter
in XWalkExtensionInstance, this pointer is used to communicate back from
.NET to C++/XWalk and the fact is dynamic is to hide implementation
details to the 3rd party developer which should only care about calling
the right functions on it. If they don't exists, there will be a runtime
error showing up. It makes using the system very simple (nothing to
include, nothing to import).

Please note that we don't support all the features of the extension
system at the moment, for example you can't add multiple entry points
in JS, and you can't have access to the runtime variables inside the
extension. These features can be added in a following patch.

This commit also roll GYP forward so we can compile passing /clr as
a compiler paramenter. It will probably be not needed by the next
rebase.

Part of the feature work I've added a bunch of tests which
actually build .NET extensions and test the overall system (thus
the .csproj files checked in). There is a little script to invoke
msbuild so it builds the DLL in .NET (GYP doesn't know about that).

### DIFF
--- a/dotnet/xwalk_dotnet_bridge.cc
+++ b/dotnet/xwalk_dotnet_bridge.cc
@@ -1,0 +1,423 @@
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/dotnet/xwalk_dotnet_bridge.h"
+
+#include <Objbase.h>
+#include <vcclr.h>
+
+#include <assert.h>
+#include <iostream>
+
+using namespace System; // NOLINT
+using namespace System::IO; // NOLINT
+using namespace System::Reflection; // NOLINT
+using namespace System::Text; // NOLINT
+using namespace Runtime::InteropServices; // NOLINT
+
+namespace {
+
+xwalk::extensions::XWalkDotNetBridge* g_bridge = nullptr;
+XW_Extension g_xw_extension = 0;
+const XW_CoreInterface* g_core = nullptr;
+const XW_MessagingInterface* g_messaging = nullptr;
+const XW_Internal_SyncMessagingInterface* g_sync_messaging = nullptr;
+const XW_Internal_EntryPointsInterface* g_entry_points = nullptr;
+const XW_Internal_RuntimeInterface* g_runtime = nullptr;
+const XW_Internal_PermissionsInterface* g_permission = nullptr;
+
+typedef void* XWalkExtensionDotNet;
+typedef void* XWalkExtensionAssembly;
+
+static void MarshalString(String^ input, std::string* output) {
+  const char* chars = reinterpret_cast<const char*>(
+    Marshal::StringToHGlobalAnsi(input).ToPointer());
+  *output = chars;
+  Marshal::FreeHGlobal(IntPtr((void*)chars)); //NOLINT
+}
+
+bool InitializeInterfaces(XW_GetInterface get_interface) {
+  g_core = reinterpret_cast<const XW_CoreInterface*>(
+    get_interface(XW_CORE_INTERFACE));
+  if (!g_core) {
+    std::cerr << "Can't initialize extension: error getting Core interface.\n";
+    return false;
+  }
+
+  g_messaging = reinterpret_cast<const XW_MessagingInterface*>(
+      get_interface(XW_MESSAGING_INTERFACE));
+  if (!g_messaging) {
+    std::cerr <<
+        "Can't initialize extension: error getting Messaging interface.\n";
+    return false;
+  }
+
+  g_sync_messaging =
+      reinterpret_cast<const XW_Internal_SyncMessagingInterface*>(
+      get_interface(XW_INTERNAL_SYNC_MESSAGING_INTERFACE));
+  if (!g_sync_messaging) {
+    std::cerr <<
+        "Can't initialize extension: error getting SyncMessaging interface.\n";
+    return false;
+  }
+
+  g_entry_points = reinterpret_cast<const XW_Internal_EntryPointsInterface*>(
+      get_interface(XW_INTERNAL_ENTRY_POINTS_INTERFACE));
+  if (!g_entry_points) {
+    std::cerr << "NOTE: Entry points interface not available in this version "
+              << "of Crosswalk, ignoring entry point data for extensions.\n";
+  }
+
+  g_runtime = reinterpret_cast<const XW_Internal_RuntimeInterface*>(
+      get_interface(XW_INTERNAL_RUNTIME_INTERFACE));
+  if (!g_runtime) {
+    std::cerr << "NOTE: runtime interface not available in this version "
+              << "of Crosswalk, ignoring runtime variables for extensions.\n";
+  }
+
+  g_permission = reinterpret_cast<const XW_Internal_PermissionsInterface*>(
+      get_interface(XW_INTERNAL_PERMISSIONS_INTERFACE));
+  if (!g_permission) {
+    std::cerr << "NOTE: permission interface not available in this version "
+              << "of Crosswalk, ignoring permission for extensions.\n";
+  }
+
+  return true;
+}
+}  // anonymous namespace
+
+int32_t XW_Initialize(XW_Extension extension, XW_GetInterface get_interface) {
+  assert(extension);
+  g_xw_extension = extension;
+
+  if (!InitializeInterfaces(get_interface))
+    return XW_ERROR;
+
+  g_bridge = new xwalk::extensions::XWalkDotNetBridge();
+  if (!g_bridge->Initialize()) {
+    std::cerr << "Can't initialize .NET extension bridge \n";
+    return XW_ERROR;
+  }
+  using namespace xwalk::extensions; // NOLINT
+  g_core->RegisterShutdownCallback(
+      g_xw_extension, XWalkDotNetBridge::OnShutdown);
+  g_core->RegisterInstanceCallbacks(
+    g_xw_extension, XWalkDotNetBridge::OnInstanceCreated,
+    XWalkDotNetBridge::OnInstanceDestroyed);
+  g_messaging->Register(
+      g_xw_extension, XWalkDotNetBridge::HandleMessage);
+  g_sync_messaging->Register(
+      g_xw_extension, XWalkDotNetBridge::HandleSyncMessage);
+  return XW_OK;
+}
+
+namespace xwalk {
+namespace extensions {
+
+struct PrivateBridgeData {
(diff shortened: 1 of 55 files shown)


darktears:
@pozdnyakov  What about this?

darktears:
Of course right now this works for one extension in the path. I'm wondering how we could make it work : On way is to have the bridges alongside each extension with the same name for example : fooextension.dll with fooextension_bridge.dll (a copy of xwalk_dotnet_bridge.dll renamed) then barextension.dll with barextension_bridge.dll. Still make the creation of extensions simple and require to copy one file. Better suggestion?

extensions/common/xwalk_extension_server.cc
```diff
@@ -406,6 +406,8 @@ std::vector<std::string> RegisterExternalExtensionsInDirectory(
     // load multiple times.
     (*runtime_variables)["extension_path"] =
         new base::StringValue(extension_path.AsUTF8Unsafe());
+    (*runtime_variables)["extension_dir"] =
```
pozdnyakov: can we get it from "extension_path" when needed?

extensions/common/xwalk_external_instance.h
```diff
@@ -30,6 +30,8 @@ class XWalkExternalInstance : public XWalkExtensionInstance {
                         XW_Instance xw_instance);
   ~XWalkExternalInstance() override;
 
+  void* InstanceData() const { return instance_data_; }
```
pozdnyakov: GetInstanceData also I'd introduce "using InstanceData = void*"

extensions/public/win/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,402 @@
+// Copyright (c) 2015 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/extensions/public/win/xwalk_dotnet_bridge.h"
```
pozdnyakov: I'd put it into "xwalk/dotnet_extensions" or "xwalk/dotnet'

extensions/public/win/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,402 @@
...
+};
+
+XWalkDotNetBridge::XWalkDotNetBridge()
+  : library_path_(""),
+    initialized_(false),
+    extension_dotnet_(0) {}
+
+void XWalkDotNetBridge::OnShutdown(XW_Extension) {
+  delete g_bridge;
+  g_bridge = NULL;
```
pozdnyakov: nit: nullptr

pozdnyakov:
I like this approach better than the previous one.

darktears:
@pozdnyakov : Implemented what we discussed. Also added a test to verify multiple extensions in the same directory works correctly.

@rakuco : what about the python code?

darktears:
In any case please do not merge. I'll do it.

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,398 @@
...
+  }
+private:
+  XWalkDotNetBridge* bridge_;
+  XW_Instance instance_;
+};
+
+XWalkDotNetBridge::XWalkDotNetBridge()
+  : library_path_(""),
+    initialized_(false),
+    extension_dotnet_(nullptr) {}
```
pozdnyakov: extension_assembly_(nullptr)

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,398 @@
...
+    void* extension_instance_dotnet =
+        static_cast<void*>(extension_instance_object_ptr);
+    callback->setNativeInstance(native_instance);
+    return extension_instance_dotnet;
+  }
+  catch (Exception^ e) {
+    std::string message;
+    MarshalString(e->ToString(), &message);
+    std::cerr << "Error creating extension instance"
+              << library_path_.c_str() << "': " << message;
```
pozdnyakov: std::err << std::string should work fine, why c_str()?

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,398 @@
...
+    MarshalString(message, &message_to_js);
+    bridge_->SetSyncReply(instance_, message_to_js);
+  }
+private:
+  XWalkDotNetBridge* bridge_;
+  XW_Instance instance_;
+};
+
+XWalkDotNetBridge::XWalkDotNetBridge()
+  : library_path_(""),
```
pozdnyakov: std::string does not require explicit initialization here

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,398 @@
...
+      g_xw_extension, XWalkDotNetBridge::HandleSyncMessage);
+  return XW_OK;
+}
+
+namespace xwalk {
+namespace extensions {
+
+public ref class Bridge : public Object {
+public:
+  Bridge(XWalkDotNetBridge* bridge)
```
pozdnyakov: explicit

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,398 @@
...
+    : bridge_(bridge),
+    instance_(0) {}
```
pozdnyakov: instance_(nullptr)

dotnet/xwalk_dotnet_bridge.h
```diff
@@ -0,0 +1,45 @@
...
+  bool Initialize();
+  void* CreateInstance(XW_Instance native_instance);
+  void PostMessageToInstance(XW_Instance instance, const std::string& message);
+  void SetSyncReply(XW_Instance instance, const std::string& message);
+
+ private:
+  std::string library_path_;
+  bool initialized_;
+  void* extension_dotnet_;
+  void* extension_assembly_;
```
pozdnyakov: how about struct PrivateData { gcroot<Object^> extension_object; ... extension_assembly; } // in .cc file
and here in class declaration:
PrivateData* p_data_;

darktears: Ok

dotnet/xwalk_dotnet_bridge.cc
darktears: XW_Instance is defined as an int32_t so I can't do that.

dotnet/xwalk_dotnet_bridge.cc
darktears: ok

dotnet/xwalk_dotnet_bridge.cc
darktears: Indeed.

dotnet/xwalk_dotnet_bridge.h
```diff
@@ -0,0 +1,44 @@
...
+  static void HandleMessage(XW_Instance xw_instance, const char* msg);
+  static void HandleSyncMessage(XW_Instance xw_instance, const char* msg);
+
+  bool Initialize();
+  void* CreateInstance(XW_Instance native_instance);
+  void PostMessageToInstance(XW_Instance instance, const std::string& message);
+  void SetSyncReply(XW_Instance instance, const std::string& message);
+
+ private:
+   PrivateBridgeData* private_data_;
```
pozdnyakov: nit: class PrivateBridgeData* private_data_;

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,414 @@
...
+
+struct PrivateBridgeData {
+  PrivateBridgeData()
+    : initialized_(false),
+      extension_dotnet_(nullptr),
+      extension_assembly_(nullptr) {
+  }
+  std::string library_path_;
+  bool initialized_;
+  void* extension_dotnet_;
```
pozdnyakov: better to use explicit types, no need void*

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,414 @@
...
+  void* extension_assembly_;
```
pozdnyakov: ditto

dotnet/xwalk_dotnet_bridge.cc
```diff
@@ -0,0 +1,414 @@
...
+        static_cast<gcroot<Object^>*>(private_data_->extension_dotnet_);
+    delete extension_object_ptr;
+    private_data_->extension_dotnet_ = nullptr;
+  }
+  if (private_data_->extension_assembly_) {
+    gcroot<Assembly^> *extension_assembly_ptr =
+        static_cast<gcroot<Assembly^>*>(private_data_->extension_assembly_);
+    delete extension_assembly_ptr;
+    private_data_->extension_assembly_ = nullptr;
+  }
```
pozdnyakov: delete private_data_

darktears: oops :)

pozdnyakov:
lgtm with comments

PR #3283, opened 2015-09-24
[Android] Experimental silent download mode

lincsoon:
This is an experimental featurea that provides developers with a way to
download the Crosswalk runtime in background silently without
interrupting the end-user.

To take a trial, you have to remove libxwalkcore.so, icudtal.dat,
xwalk.pak from xwalk_core_library, and put these files in a zip file.
Then please host this zip file, and specify the url with a meta-data
element named "xwalk_lib_url" inside the application tag in the Android
manifest.

You also must implement the interface XWalkUpdate.XWalkSilentUpdateListener.


lincsoon:
@axinging @fujunwei @XiaosongWei

XiaosongWei:
LGTM

PR #3284, opened 2015-09-25
M46 gardening: adapt to https://codereview.chromium.org/1272113002

mrunalk:
@seanlong can you please advise if we need to add standard url scheme after above change? As I see constants.cc already has "app" Application Scheme already defined. Not sure about the difference between standard and savable scheme.


rakuco:
Please note that @-mentions in the pull request message do not add the people to the PR; also, I don't think Xiang Long is working on Crosswalk anymore.

mrunalk:
@rakuco but I get notified when someone @-mentions my name in the PR :) Anyways thanks for the heads up about Xiang Long. I will ask either @halton or someone appropriate to review this patch.

halton:
This PR is obvious lgtm.

PR #3286, opened 2015-09-25
Fix occasional crash.

PeterWangIntel:
Empty the Java-bounding functions of C++, once devtools is disabled to
avoid incident invoke from Java.

BUG=XWALK-5051


PeterWangIntel:
@d0ngjun @halton 
this patch is to resolve crash caused by disabling devtools. It's verified on real device.

halton:
LGTM, merging.

PeterWangIntel:
thx

PR #3288, opened 2015-09-28
[Android] Backport patch to encode the data by base64

fujunwei:
The data type of "text/html" can be parsed, so don't need be escaped before
loading.

BUG=XWALK-3571

(cherry picked from commit 776dcd72649f71c92755524f32ed4681449250c8)


fujunwei:
@lincsoon @halton  PTAL, thanks.

PR #3289, opened 2015-09-28
Revert "[Android] Experimental silent download mode"

lincsoon:
This reverts commit dbab22de26da9649be62c400a8d00249729dab8b.


halton:
@lincsoon any reason to revert it? Seems I missed something.

lincsoon:
The details of silent download mode are still under discussion. To avoid misunderstanding that we've released this feature, I revert this commit and we'll implement this feature in a standalone branch.

PR #3290, opened 2015-09-28
Cleanup some unused files.

darktears:
Nobody is using them.


r0b5t4:
lgtm

joone:
Tizen may use this files? We need to check that.

darktears:
@joone : We have bots for that. Tizen Common is green :)

PR #3291, opened 2015-09-30
[Android] Host android support class in app/android/app_support direc…

fujunwei:
…tory

Those classes need to be compiled in app_side because of depending on
Android Support library.

### DIFF
--- a/build/android/merge_jars.py
+++ b/build/android/merge_jars.py
@@ -7,6 +7,7 @@
 
 import optparse
 import os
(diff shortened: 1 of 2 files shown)


fujunwei:
The commit depend on https://github.com/fujunwei/chromium-crosswalk/commit/63c781aa1b9c6782809bbc81c419634eed380421

fujunwei:
@rakuco PTAL, thanks.

PR #3294, opened 2015-10-07
Implement --enable-inspector command line.

darktears:
By default there is no inspector support (right click) but
when passing --enable-inspector it is possible to launch the
inspector.

We need that flag so that crosswalk-app-tools can use it to
enable/disable inspector (especially when you actually make
a release of your .msi or application, you don't want the
inspector to be enabled).

BUG=XWALK-4858


darktears:
@pozdnyakov @r0b5t4 some review meat for you guys

r0b5t4:
lgtm

pozdnyakov:
lgtm

PR #3296, opened 2015-10-08
Roll new history after cleaning up our upstream forks.

darktears:


darktears:
@rakuco PTAL

rakuco:
le lgtm :fr:

PR #3300, opened 2015-10-10
Remove patch_skia.py and its related patch.

rakuco:
They were only necessary before moving to M45. The hook that applied it
to the tree was already removed in 9eff35de ("Roll Chromium
45.0.2454.85"), but the script and patch were left behind.


rakuco:
ping @darktears

pozdnyakov:
rs=me

PR #3301, opened 2015-10-12
Fix right click empty menu when inspector is disabled, also

darktears:
fixed leak of the menu itself if you right click several times.


darktears:
@pozdnyakov PTAL

pozdnyakov:
lgtm, sorry for delay

PR #3302, opened 2015-10-13
[Android] Fix the issue about manifest.json was always used.

wuhengzhi:
This patch is to fix manifest.json was always used issue.
In application's folder, if manifest.json was existed, it will be
always used even if it was not specified in command line.
e.g. if manifest.json was existed and app was packaged by "make_apk.py
--package=org.xwalk.test --name=test --app-root=path-to-app
--app-local-path=index.html", this way won't work, manifest.json will be
used.

BUG=XWALK-5131

### DIFF
--- a/app/tools/android/customize.py
+++ b/app/tools/android/customize.py
@@ -323,7 +323,7 @@ def SetVariable(file_path, string_line, variable, value):
   shutil.move(temp_file_path, file_path)
 
 
-def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on):
+def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on, manifest):
   name = app_info.android_name
   package = app_info.package
   app_dir = GetBuildDir(name)
@@ -331,12 +331,14 @@ def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on):
   dest_activity = os.path.join(app_pkg_dir, name + 'Activity.java')
   ReplaceString(dest_activity, 'org.xwalk.app.template', package)
   ReplaceString(dest_activity, 'AppTemplate', name)
-  manifest_file = os.path.join(app_dir, 'assets', 'www', 'manifest.json')
-  if os.path.isfile(manifest_file):
+  if manifest:
+    manifest_name = os.path.basename(manifest.input_path)
+    manifest_in_asset = 'file:///android_asset/www/' + manifest_name
+    load_from_manifest = 'loadAppFromManifest("' + manifest_in_asset + '")'
     ReplaceString(
         dest_activity,
         'loadAppFromUrl("file:///android_asset/www/index.html")',
-        'loadAppFromManifest("file:///android_asset/www/manifest.json")')
+        load_from_manifest)
   else:
     if app_url:
       if re.search(r'^http(|s)', app_url):
@@ -589,7 +591,7 @@ def CustomizeAll(app_info, description, icon_dict, permissions, app_url,
   try:
     Prepare(app_info, compressor)
     CustomizeXML(app_info, description, icon_dict, manifest, permissions)
-    CustomizeJava(app_info, app_url, app_local_path, keep_screen_on)
+    CustomizeJava(app_info, app_url, app_local_path, keep_screen_on, manifest)
     CustomizeExtensions(app_info, extensions)
     GenerateCommandLineFile(app_info, xwalk_command_line)
   except SystemExit as ec:
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -908,16 +908,12 @@ def main(argv):
     if not os.path.isfile(options.manifest):
       print('Error: The manifest file does not exist.')
       sys.exit(8)
-
-  if options.app_root and not options.manifest:
-    manifest_path = os.path.join(options.app_root, 'manifest.json')
-    if os.path.exists(manifest_path):
-      print('Using manifest.json distributed with the application.')
-      options.manifest = manifest_path
-
-  app_info = AppInfo()
-  manifest = None
-  if not options.manifest:
+    try:
+      manifest = ParseManifest(options)
+    except SystemExit as ec:
+      return ec.code
+  else:
+    manifest = None
     # The checks here are really convoluted, but at the moment make_apk
     # misbehaves any of the following conditions is true.
     if options.app_url:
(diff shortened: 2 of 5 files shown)


rakuco:
This confusing behavior was introduced in e65612ece27df264225649e6fb7e307774e63cef.
To me, it makes more sense to remove this check altogether and stop second-guessing the user: either --manifest is specified and we use it, or the user must specify the --app-* options.

Doing so also allows you to simplify the code around the check, so it looks like this:

```python
xpk_temp_dir = ''
# xpk block

if options.manifest:
  options.manifest = os.path.abs(options.manifest)
  if not os.path.isfile(options.manifest):
    # yadda, yadda
  try:  # this is ugly, but can be changed separately
    manifest = ParseManifest(options)
  except SystemExit as ec:
    return ec.code
else:
  manifest = None
  # The checks here are really convoluted [...]

if not options.name:
  # ...

try:
  app_info = AppInfo()
  MakeApk(...)
# ...
```

wuhengzhi:
I will refine it, thanks.

app/tools/android/make_apk.py
```diff
@@ -903,21 +903,17 @@ def main(argv):
     CleanDir(xpk_temp_dir)
     ParseXPK(options, xpk_temp_dir)
 
+  manifest = None
```
rakuco: You don't need this here; Python's scoping works so that you can move this line to the `else` clause below.

rakuco:
lgtm!

rakuco:
btw, one thing that worries me a little is that nothing had to be changed in the make apk tests, which means that part was not being checked. did you verify both code paths? is it possible to add a test for it, or adapt an existing one?

wuhengzhi:
Sorry for forgetting the test case, will add it.

wuhengzhi:
@rakuco Test case is done, ```customize.py``` has also been modified to fix this issue if the ```start_url``` is not same as the current html files.

app/tools/android/customize.py
```diff
@@ -331,8 +331,7 @@ def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on):
   dest_activity = os.path.join(app_pkg_dir, name + 'Activity.java')
   ReplaceString(dest_activity, 'org.xwalk.app.template', package)
   ReplaceString(dest_activity, 'AppTemplate', name)
-  manifest_file = os.path.join(app_dir, 'assets', 'www', 'manifest.json')
-  if os.path.isfile(manifest_file):
+  if manifest:
```
rakuco: If the file is not called `manifest.json`, isn't the `ReplaceString()` call below going to fail (ie. it'll generate a load call for a file that is not present)?

app/tools/android/make_apk_test.py
```diff
@@ -781,6 +781,50 @@ def testManifestWithError(self):
     self.assertTrue(
         out.find('\'Telephony\' related API is not supported') != -1)
 
+  def testManifestWithHtmlFile(self):
+    manifest_path = os.path.join('test_data', 'manifest',
+                                 'unite', 'manifest.json')
```
rakuco: I don't understand why this new directory is called "unite". What is it supposed to mean? Can you come up with a more descriptive name, like "manifest_and_app_root", or "cli_parser_priority"?

app/tools/android/customize.py
wuhengzhi: Yes, it will be failed, inconsiderate.

app/tools/android/customize.py
```diff
@@ -331,12 +331,18 @@ def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on):
...
   ReplaceString(dest_activity, 'org.xwalk.app.template', package)
   ReplaceString(dest_activity, 'AppTemplate', name)
-  manifest_file = os.path.join(app_dir, 'assets', 'www', 'manifest.json')
-  if os.path.isfile(manifest_file):
-    ReplaceString(
-        dest_activity,
-        'loadAppFromUrl("file:///android_asset/www/index.html")',
-        'loadAppFromManifest("file:///android_asset/www/manifest.json")')
+  if manifest:
+    manifest_name = manifest.input_path.split(os.path.sep)[-1]
```
rakuco: You can use `os.path.basename()` here.

app/tools/android/customize.py
```diff
@@ -331,12 +331,18 @@ def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on):
...
+    manifest_file = os.path.join(app_dir, 'assets', 'www', manifest_name)
+    if os.path.isfile(manifest_file):
```
rakuco: I think it is safe to assume that if the file does not exist we should have failed earlier. Let's just consider it is present if `manifest` is set.

app/tools/android/customize.py
```diff
@@ -331,12 +331,18 @@ def CustomizeJava(app_info, app_url, app_local_path, keep_screen_on):
...
+      manifest_in_asset = 'file:///android_asset/www/' + manifest_name
```
rakuco: Nonetheless, have you checked if the manifest ends up in `www/` if it is in a sub-directory (ie. does `--manifest=foo/bar/my_manifest.json` cause the file to be in `www/my_manifest.json` or `www/foo/bar/my_manifest.json`)?

app/tools/android/customize.py
wuhengzhi: Do you mean the check at https://github.com/crosswalk-project/crosswalk/blob/master/app/tools/android/make_apk.py#L908? It checks the specified ```manifest```.
Here we re-check it under ```assets/www```.

app/tools/android/customize.py
wuhengzhi: The specified ```foo/bar/my_manifest.json``` will be copied to ```www/my_manifest.json```

app/tools/android/customize.py
wuhengzhi: Will update it, thanks.

app/tools/android/customize.py
rakuco: Yeah, but can't you assume this code is always called after everything has been copied to `assets/www` (meaning that if copying has failed the entire program should have failed and exited)? I'd like to avoid having duplicated checks everywhere like this one.

wuhengzhi:
@rakuco PTAL again, thanks.

rakuco:
lgtm!

wuhengzhi:
@rakuco This patch should be merged to crosswalk-15/16, what do you think?

rakuco:
I don't think the problem it fixes is severe enough for it to make it into Crosswalk 15; as for Crosswalk 16, I defer the decision to @baleboy

baleboy:
Crosswalk 16 is still open for fixes, so I'm fine to have this one in

wuhengzhi:
I will cherry pick to Crosswalk-16.

PR #3310, opened 2015-10-19
Fix issue in coverity : Uninitialized pointer field in

darktears:
XWalkExternalExtension.

Make sure the constructor gives a NULL value.


mrunalk:
LGTM

PR #3313, opened 2015-10-20
[Android][64-bit] Fix launching app crash.

shawngao5:
Use correct method to get native library path.

BUG=XWALK-5219

### DIFF
--- a/app/android/runtime_client/src/org/xwalk/app/runtime/extension/XWalkRuntimeExtensionManager.java
+++ b/app/android/runtime_client/src/org/xwalk/app/runtime/extension/XWalkRuntimeExtensionManager.java
@@ -7,6 +7,8 @@ package org.xwalk.app.runtime.extension;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager.NameNotFoundException;
 import android.content.res.AssetManager;
 import android.content.res.Resources;
 import android.content.res.Resources.NotFoundException;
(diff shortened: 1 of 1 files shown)


app/android/runtime_client/src/org/xwalk/app/runtime/extension/XWalkRuntimeExtensionManager.java
```diff
@@ -152,10 +154,15 @@ private void loadExternalExtensions() {
...
-        String path = "/data/data/" + mActivity.getPackageName() + "/lib";
-        File libraryDir = new File(path);
-        if (libraryDir.listFiles().length > 0)
-            mNativeExtensionLoader.registerNativeExtensionsInPath(path);
+        String path = "";
+        try {
+            ApplicationInfo appInfo =
+                    mContext.getPackageManager()
+                    .getApplicationInfo(mActivity.getPackageName(), 0);
+            path = new File(appInfo.nativeLibraryDir).getParent();
```
halton: Can not understand why use parent of library dir?

app/android/runtime_client/src/org/xwalk/app/runtime/extension/XWalkRuntimeExtensionManager.java
```diff
@@ -152,10 +154,15 @@ private void loadExternalExtensions() {
...
+        } catch (final NameNotFoundException e) {
+        }
+        mNativeExtensionLoader.registerNativeExtensionsInPath(path);
```
halton: Why to remove previous checking `if (libraryDir.listFiles().length > 0)`?

shawngao5: This method is not current. "nativeLibraryDir" is more straightforward, which be introduced  since 2.3.x.

halton:
lgtm, will merge after trybot green

PR #3314, opened 2015-10-20
Fix flakiness when building the .NET extensions tests

darktears:
On a machine where ninja creates a lot of parrallel tasks we were
running in some build errors :
- The .NET extensions tests depends on the .NET bridge but it was
not listed as an explicit dependency.
- In the multiple extensions tests (where we have several extensions in
the same directory with multiple bridges) we were copying multiple times
the bridge and renaming it according to each extension in the directory.
We can't do these tasks in parallel because we copy then rename the bridge for
extension 1 and then in the same time we copy and rename the bridge
for extension 2. Turns out that the bridge file may be accessed by
both actions running in parrallel. We make it sequential : we copy
the bridge for extension 1, then rename it, then copy the bridge for
extension 2 and then rename it.

### DIFF
--- a/extensions/dotnet_extension_sample.gyp
+++ b/extensions/dotnet_extension_sample.gyp
@@ -644,10 +644,10 @@
       ],
     },
     {
-      'target_name': 'create_dir_with_multiple_extensions',
+      'target_name': 'copy_echo_extension2_bridge',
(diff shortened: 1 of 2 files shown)


r0b5t4:
lgtm

PR #3317, opened 2015-10-22
[Android] Add build target for XWalkRuntimeLibLzma.apk

XiaosongWei:
This runtime APK is used in silent download mode and the size is
significantly reduced by using LZMA compression to compress the
key components of the runtime APK.

We put the compressed ingredients to assets and then package them as
an APK, this is intented to utilize the APK integrity/signature
check for security concern.

This is a partial implementation for XWALK-5214

### DIFF
--- a/build/android/lzma_compress.py
+++ b/build/android/lzma_compress.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2015 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+# pylint: disable=F0401
+
+import optparse
+import os
+import shutil
+import sys
+import subprocess
+
+GYP_ANDROID_DIR = os.path.join(os.path.dirname(__file__),
+                               os.pardir, os.pardir, os.pardir,
+                               'build',
+                               'android',
+                               'gyp')
+sys.path.append(GYP_ANDROID_DIR)
+
+from util import build_utils
+
+
+def DoCompress(dest_path, sources):
+  build_utils.DeleteDirectory(dest_path)
+  build_utils.MakeDirectory(dest_path)
+
+  for source in sources:
+    shutil.copy(source, dest_path)
+    file_to_compress = os.path.join(dest_path, os.path.basename(source))
+    subprocess.check_call(['lzma', '-f', file_to_compress])
+
+
+def DoShowOutputNames(dest_path, sources):
+  for source in sources:
+    print('%s.lzma' % os.path.join(dest_path, os.path.basename(source)))
+
+
+def main():
+  parser = optparse.OptionParser()
+  parser.add_option('--dest-path',
+                    help='Destination directory for compressed files')
+  parser.add_option('--mode', choices=('compress', 'show-output-names'),
+                    help='Whether to compress the files or show their '
+                    'compressed names')
+  parser.add_option('--sources', help='The list of files to be compressed')
+
+  options, _ = parser.parse_args(sys.argv)
+  sources = build_utils.ParseGypList(options.sources)
+
+  if options.mode == 'compress':
+    return DoCompress(options.dest_path, sources)
+  else:
+    return DoShowOutputNames(options.dest_path, sources)
+
+
+if __name__ == '__main__':
+  sys.exit(main())
--- a/xwalk_android.gypi
+++ b/xwalk_android.gypi
@@ -213,6 +213,23 @@
       'includes': ['../build/jni_generator.gypi'],
     },
     {
+      'target_name': 'xwalk_runtime_lib_lzma_apk',
+      'type': 'none',
+      'dependencies': [
+        'xwalk_runtime_lib_apk'
+      ],
+      'variables': {
+        'apk_name': 'XWalkRuntimeLibLzma',
+        'java_in_dir': 'runtime/android/runtime_lib',
+        'resource_dir': 'runtime/android/runtime_lib/res',
+        'asset_location': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/assets',
+        'app_manifest_version_name': '<(xwalk_version)',
+        'app_manifest_version_code': '<(xwalk_version_code)',
+        'is_test_apk': 1,
+      },
+      'includes': ['../build/java_apk.gypi'],
+    },
+    {
       'target_name': 'xwalk_runtime_lib_apk',
       'type': 'none',
       'dependencies': [
@@ -252,6 +269,47 @@
         'app_manifest_version_code': '<(xwalk_version_code)',
       },
       'includes': ['../build/java_apk.gypi'],
+      'actions': [
+        {
+          'action_name': 'runtime_lib_lzma',
+          'message': 'Compressing runtime APK assets with LZMA',
+          'variables': {
+            # We have to use three separate variables because libxwalkcore.so
+            # in the <(stripped_libraries_dir) is not registered as an output
+            # in the build system, so we cannot use it as an input for the
+            # action. Instead, we use the stamp file produced by
+            # strip_native_libraries.gypi and pass the library as an input only
+            # to the lzma_compress.py script.
+            'base_inputs': [
+              '<(dex_path)',
+              '<(PRODUCT_DIR)/xwalk_runtime_lib/assets/xwalk.pak',
+              '<(PRODUCT_DIR)/xwalk_runtime_lib/assets/icudtl.dat',
+            ],
+            'build_system_inputs': [
+              '<@(base_inputs)',
+              '<(strip_stamp)',
+            ],
+            'lzma_compress_inputs': [
+              '<@(base_inputs)',
+              '<(stripped_libraries_dir)/libxwalkcore.so',
+            ],
+            'assets_dir': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/assets',
+          },
+          'inputs': [
+            '<@(build_system_inputs)',
+          ],
+          'outputs': [
+            "<!@(['python', 'build/android/lzma_compress.py', '--mode=show-output-names', \
+                  '--sources=<(lzma_compress_inputs)', '--dest-path=<(assets_dir)'])",
+          ],
+          'action': [
+            'python', 'build/android/lzma_compress.py',
+            '--mode=compress',
+            '--dest-path=<(assets_dir)',
+            '--sources=<(lzma_compress_inputs)',
+          ],
+        },
+      ],
     },
     {
       'target_name': 'xwalk_runtime_lib_apk_pak',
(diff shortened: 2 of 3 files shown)


XiaosongWei:
@lincsoon @halton Please take a look, Thanks!
@rakuco We need to distribute this APK through 01.org download.

XiaosongWei:
@axinging

xwalk_android.gypi
```diff
@@ -213,6 +213,101 @@
...
+        'xwalk_runtime_lib_apk',
+      ],
+      'variables': {
+        'assets_dir': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/assets/',
+        'libxwalkcore_so': '<(PRODUCT_DIR)/xwalk_runtime_lib_apk/stripped_libraries/libxwalkcore.so',
+        'classes_dex': '<(PRODUCT_DIR)/xwalk_runtime_lib_apk/classes.dex',
+        'icudtl_dat': '<(PRODUCT_DIR)/xwalk_runtime_lib/assets/icudtl.dat',
+        'xwalk_pak':'<(PRODUCT_DIR)/xwalk_runtime_lib/assets/xwalk.pak',
+      },
+      'actions': [
```
rakuco: What if instead of having a series of hardcoded actions like this you just have one that takes a list of arguments (files)? The same Python script could also be used in a mode that just outputs the file names with ".lzma" appended to them, sort of like `src/build/grit_action.gypi`.

build/android/lzma_compress.py
```diff
@@ -0,0 +1,37 @@
...
+  option_parser.add_option('-f', dest='filename',
+                           help='The filename to be comprossed from source to target',
+                           type='string')
+
+  options, _ = option_parser.parse_args(argv)
+
+  if not os.path.exists(options.target):
+    os.makedirs(options.target)
+
+  print ('Copying file %s from %s' %(os.path.basename(options.filename), os.path.dirname(options.filename)))
```
rakuco: Minor nitpick: you need a space after "%".

rakuco: And there shouldn't be a space after `print`.

build/android/lzma_compress.py
```diff
@@ -0,0 +1,37 @@
...
+  shutil.copy(options.filename, options.target)
+
+  file_to_compress = os.path.join(options.target, os.path.basename(options.filename))
+  print ('Compressing file %s to %s' %(os.path.basename(options.filename), options.target))
+  subprocess.check_call(['lzma', '-f', file_to_compress])
+  print ('Compress %s done' %(os.path.basename(options.filename)))
+
+if __name__ == '__main__':
+  sys.exit(main(sys.argv))
```
rakuco: There's no need to pass `sys.argv`, `main` can access it directly.

XiaosongWei:
@rakuco Please take a look again. thanks.

xwalk_android.gypi
```diff
@@ -213,6 +213,59 @@
...
+        'assets_dir': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/assets/',
+      },
+      'actions': [
+        {
+          'action_name': 'lzma_compress',
+          'message': 'compress ingredients for runtime APK',
+          'inputs': [
+            'build/android/lzma_compress.py',
+          ],
+          'outputs': [
```
rakuco: I still think that instead of hardcoding the list here you could add a mode to the new Python script to output those names. This would also allow you to do more in `files_to_compress`, like maybe putting the V8 snapshot files there in the future if/when we start using them.

XiaosongWei: Those names will be automatically generated, will use a timestamp instead in the action outputs.

rakuco: I might be wrong, but if you do this you risk making `ninja -t clean` fail and if any other target depends on the files generated here they might not work since those files won't be known to gyp. In general, it is better to register all files that your action ends up creating.

build/android/lzma_compress.py
```diff
@@ -0,0 +1,52 @@
...
+GYP_ANDROID_DIR = os.path.join(os.path.dirname(__file__),
+                               os.pardir, os.pardir, os.pardir,
+                               'build',
+                               'android',
+                               'gyp')
+sys.path.append(GYP_ANDROID_DIR)
+
+from util import build_utils
+
+def Touch(path):
```
rakuco: That's exactly what `build_utils.Touch` does.

xwalk_android.gypi
XiaosongWei: I tried ```ninja -t clean``` and it works. It's fine to just put a stamp in the action output, please see target ```xwalk_core_reflection_layer_java_gen``` in this file. Actually, there's no any target directly depends on the files generated here, the action here is just used to prepare assets for XWalkRuntimeLibLzma APK generation.

rakuco: > I tried ninja -t clean and it works.

All files in `out/Release/xwalk_runtime_lib_lzma/assets/` remain there after the call to `-t clean`, the only file that gets removed is the stamp.

> Actually, there's no any target directly depends on the files generated here, the action here is just used to prepare assets for XWalkRuntimeLibLzma APK generation.

I understand that, yet time and again we've been bitten by this: a year from now someone copy-pastes this block elsewhere to implement a new feature, or this doesn't work as expected with GN, or someone uses this target in a new one and things work unpredictably. I really don't see any downside to properly specifying all your outputs.

xwalk_android.gypi
```diff
@@ -213,6 +213,58 @@
...
+          '<(PRODUCT_DIR)/xwalk_runtime_lib/assets/xwalk.pak',
+        ],
+        'assets_dir': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/assets/',
+        'stamp': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/gen.stamp',
+      },
+      'actions': [
+        {
+          'action_name': 'lzma_compress',
+          'message': 'compress ingredients for runtime APK',
+          'inputs': [
```
rakuco: You need `<@(files_to_compress)` here, otherwise you'll only run this script when the script itself has changed.

xwalk_android.gypi
```diff
@@ -213,6 +213,58 @@
...
+          '<(PRODUCT_DIR)/xwalk_runtime_lib_apk/stripped_libraries/libxwalkcore.so',
+          '<(PRODUCT_DIR)/xwalk_runtime_lib_apk/classes.dex',
+          '<(PRODUCT_DIR)/xwalk_runtime_lib/assets/icudtl.dat',
+          '<(PRODUCT_DIR)/xwalk_runtime_lib/assets/xwalk.pak',
+        ],
+        'assets_dir': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/assets/',
+        'stamp': '<(PRODUCT_DIR)/xwalk_runtime_lib_lzma/gen.stamp',
+      },
+      'actions': [
+        {
```
rakuco: In addition to this action, you should probably look at `src/build/rmdir_and_stamp.py` and how it is used, for example, in `src/third_party/mojo/mojom_bindings_generator.gypi`. The idea is that you wipe out `assets_dir` before working on the files you copy to it, otherwise you risk having stale files there in the future that can break incremental builds or generate wrong APKs.

build/android/lzma_compress.py
```diff
@@ -0,0 +1,46 @@
...
+
+def main():
+  parser = optparse.OptionParser()
+  parser.add_option('--dest-path', help='Destination directory for compressed files')
+  parser.add_option('--sources', help='The list of files to be comprossed')
+  parser.add_option('--stamp', help='The file to touch on success.')
+
+  options, _ = parser.parse_args(sys.argv)
+
+  if not os.path.exists(options.dest_path):
```
rakuco: You could use `build_utils.MakeDirectory()` and avoid having to check for the path.

xwalk_android.gypi
XiaosongWei: I used to put it here, but it will run into problem:
```ninja: error: 'xwalk_runtime_lib_apk/stripped_libraries/libxwalkcore.so', needed by 'xwalk_runtime_lib_lzma/gen.stamp', missing and no known rule to make it```

Two input files (libxwalkcore.so and classes.dex) are generated in building xwalk_runtime_lib_apk target and ninja don't know rule to make them.

File: xwalk_android.gypi (hunk 4 above)
Comment (XiaosongWei) [inline xwalk_android.gypi]: I could add generated files to outputs, but it would be hardcoded. The purpose of using stamp only is to avoid that hard code list.

File: xwalk_android.gypi (hunk 7 above)
Comment (XiaosongWei) [inline xwalk_android.gypi]: will wipe out the dir in the python script instead.

File: xwalk_android.gypi (hunk 4 above)
Comment (rakuco) [inline xwalk_android.gypi]: That's why I suggested adding a new mode to `lzma_compress.py` that just printed its list of inputs with ".lzma" appended.

Comment (XiaosongWei) [inline xwalk_android.gypi]: I am a little bit confused by the mode mentioned here. Could you please elaborate? Those output file name will be automatically with ".lzma" appended when running the python script ```lzma_compress.py```. The action outputs is needed only for setup the ninja build dependency, please see below.

```
 build xwalk_runtime_lib_lzma/gen.stamp $
    xwalk_runtime_lib_lzma/assets//libxwalkcore.so.lzma $
    xwalk_runtime_lib_lzma/assets//classes.dex.lzma $
    xwalk_runtime_lib_lzma/assets//icudtl.dat.lzma $
    xwalk_runtime_lib_lzma/assets//xwalk.apk.lzma: $
xwalk_runtime_lib_lzma_compress_target_lzma_compress_6e0c00b19991794c1889fe40f5e2db2d $
    ../../xwalk/build/android/lzma_compress.py || $
    obj/xwalk/xwalk_runtime_lib_apk.actions_rules_copies.stamp
```

XiaosongWei: @rakuco ping...

rakuco: Unfortunately my time is finite :-) I'm working on some other Android stuff, but will try to get back to this pull request as soon as I can.

halton:
@heke123 you should be included in this review cause it affect lite branch somehow.

rakuco:
OK, I finally had time to get back to this. See [my branch 3317-the-right-way](https://github.com/rakuco/crosswalk/commits/3317-the-right-way): it implements the suggestions I've mentioned in this review on top of your existing commit. There's no hardcoding of outputs anymore, and the action input list is now correct.

XiaosongWei:
@rakuco Great, your implementation looks better. Would you help to submit a PR combining the full commits? I wanted to update this PR with your commit but found that it's still WIP. Thanks!

rakuco:
feel free to squash them together and update this pull request. I only marked my commit WIP because it was a proof of concept.

XiaosongWei:
@rakuco Thanks so much! Updated.

rakuco:
merging. please remember to file a ticket if you need additional APKs to be shipped in download.01.org.

PR #3319, opened 2015-10-23
Fix blurry rendering on high DPI screens.

darktears:
We need to make sure to mark the process as "DPI aware".

BUG=XWALK-5170


darktears:
@pozdnyakov PTAL

@baleboy enjoy!

pozdnyakov:
lgtm

PR #3320, opened 2015-10-26
[Android] Fix the display with overlayed multiple XWalkViews.

wuhengzhi:
When multiple XWalkViews were set to overlay,
XWalkView.setVisibility(View.INVISIBLE) does not work, that is current
XWalkView was always showing. This patch is to fix this issue.
For the usage, please refer to the
sample:MultiXWalkViewOverlayActivity.java.

BUG=XWALK-5064


wuhengzhi:
@lincsoon @fujunwei  @axinging PTAL, thanks.

axinging:
LGTM

PR #3322, opened 2015-10-27
To disable Accessibility support

PeterWangIntel:
This feature is about supporting disabled people, such as reading screen.
Since it's pretty independent in code, so it's harmless for the stability of code to disable it.

Reduced size = 72K

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = '357ca5c5aa727768d803aff71cb2bb6d7da01faf'
+chromium_crosswalk_rev = 'a073a6737aad6673d170c427c9477c8c36ba83ec'
 v8_crosswalk_rev = '255b844a25c36e84145bab3751adde7c0e8f5dd6'
 ozone_wayland_rev = '1cde4077b26fb5cbb2661b465b496550a587df55'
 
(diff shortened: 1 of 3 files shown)


PeterWangIntel:
@heke123 @halton

r0b5t4:
"So, It's harmless to disable it."
It is not that simple. For people who are disabled, it is quite important indeed. Additionally, in some countries this functionality is required to be eligible for contracts from official agencies (magistrate/government etc). Needs to be considered carefully, on non-technical reasons.

baleboy:
Agree that for those who need it, this is very important. Also is it really worth removing it for a 72K gain?

heke123:
Actually what we did is adding a flag to control it, instead of just remove it.  So for users need this function, we just open the flag, then everything is OK.  and for those who don't want this and care about the binary size, we can delete this function.

heke123:
I agree that those non-tech reasons are really very important, so we need to consider when to use the "disable flag" carefully.

PeterWangIntel:
@r0b5t4 @baleboy ,
Seems, there is misunderstanding. For CrosswalkLite, we're providing series of patch to 'disable' some features, but all of them are optional. The App developer can choose 'enable' or 'disable' the features for the special scenario of their App.

r0b5t4:
@PeterWangIntel thanks for the clarification. I do agree that small size is important, and for some applications like some games the feature might not be useful.

baleboy:
Same here, sorry for misunderstanding

PeterWangIntel:
@r0b5t4 @baleboy ,
No problem. My code comments should be improved.

heke123:
LGTM

PeterWangIntel:
@halton 
Please help to merge, thx.

PeterWangIntel:
@halton 
Please help to merge. Thx.

PR #3325, opened 2015-10-28
[Android]Fix the issue that 'File chooser' does not assume device lan…

xzhan96:
…guage settings

As the default ChooserActivity will get default title from
internal resource, which includes all languages' string, there
is no need to set the chooser intent title.

BUG=XWALK-5178


xzhan96:
@lincsoon @axinging @halton @wuhengzhi

wuhengzhi:
LGTM, thanks.

rakuco:
does this need to be backported?

xzhan96:
@rakuco do you mean revert this patch? no need, even customer has another issue, this patch fix the "File chooser" title language issue indeed.

xzhan96:
@rakuco I think so. As there still another language issue of "File chooser", I will backport the patches after fixing another one.

PR #3328, opened 2015-10-28
Fixed the fail to build issue on Android.

jondong:
In runtime/browser/devtools/xwalk_devtools_frontend.h/.cc, the class
xwalk::NativeAppWindowDesktop is needed, but the source files of this
class's implementation at:
runtime/browser/ui/native_app_window_desktop.h/.cc are excluded within
Android build. As XWalkDevToolsFrontend is not needed for Android
platform for now, we'd better exclude its source files for Android
build as well.

### DIFF
--- a/xwalk.gyp
+++ b/xwalk.gyp
@@ -322,6 +322,8 @@
             'experimental/native_file_system/virtual_root_provider_android.cc',
           ],
           'sources!':[
+            'runtime/browser/devtools/xwalk_devtools_frontend.cc',
+            'runtime/browser/devtools/xwalk_devtools_frontend.h',
             'runtime/browser/runtime_ui_delegate_desktop.cc',
             'runtime/browser/runtime_ui_delegate_desktop.h',
             'runtime/browser/ui/desktop/download_views.cc',


jondong:
@darktears @halton PTAL.

rakuco:
Under what circumstances does the build fail and what does the error message say (these should have been mentioned in the commit/PR message, please remember to do that in the future)? We've been producing Android canaries without any issues.

jondong:
@rakuco Thanks for your reminder, I think I can open a bug for this, and post the build log onto it.

rakuco:
If the error message is not very big, I tend to put it in the commit message itself so people do not need to go to JIRA to find out what something was fixing in the future. See, for example, b42e0238486f08e12727873a2b2e579f11240048 or d079f0ca280bf8debc0d21c65efe980caa31b639.

jondong:
Updated per above comments, @rakuco @darktears PTAL again, thanks.

rakuco:
Thanks. I normally keep the commit and PR in sync; the latter is what most people will look at, and it still does not contain all information.

Anyway, I'm still curious about what's different in your setup: I built Android successfully yesterday, all our bots are green and our canaries are being built. Is this a debug build? Are you using different flags?

jondong:
yes that is a debug build, and I'll try to find out why when get back to office tomorrow, thanks :)

rakuco:
I was able to reproduce it with a Debug build indeed, thanks.

@jondong: personally I'd mention that this fixes the Debug build.
@darktears: PTAL and check if we can't just rename those files in a way that they are not picked up by the Android build.

jondong:
commit log updated. PTAL again.

darktears:
LGTM.

rakuco:
@jondong: does this also affect crosswalk-16? If so, can you backport this fix?

rakuco:
@jondong @halton: reping

jondong:
@rakuco sorry for the late response, yes this patch should also be cherry-picked to crosswalk-16, do I need to submit another PR to do this?

rakuco:
yes, please. also remember to use `git cherry-pick -x` (the `-x` is especially important) to cherry-pick your commit.

PR #3330, opened 2015-10-29
[Android]Fix the issue that 'File chooser' does not assume device lan…

xzhan96:
…guage settings

As the default ChooserActivity will get default title from
internal resource, which includes all languages' string, there
is no need to set the chooser intent title.

BUG=XWALK-5178

(cherry picked from commit 8d00de878ebe25c868f7afb28d8d7526c56f5be6)


xzhan96:
@rakuco please help to review and merge,thanks

rakuco:
@baleboy: do we need this in Crosswalk 15?

baleboy:
16 is enough

PR #3333, opened 2015-10-29
[Android] Add download mode support in XWalkActivityDelegate

XiaosongWei:
Insert a meta-data named "xwalk_enable_download_mode" to AndroidManifest
to enable download mode.

Related to XWALK-5121

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
@@ -5,16 +5,21 @@
 package org.xwalk.core;
 
 import android.app.Activity;
+import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.PackageManager.NameNotFoundException;
 import android.util.Log;
 import android.view.Window;
 
 import org.xwalk.core.XWalkLibraryLoader.ActivateListener;
 import org.xwalk.core.XWalkLibraryLoader.DecompressListener;
+import org.xwalk.core.XWalkUpdater.XWalkBackgroundUpdateListener;
 import org.xwalk.core.XWalkUpdater.XWalkUpdateListener;
 
 public class XWalkActivityDelegate
-            implements DecompressListener, ActivateListener, XWalkUpdateListener {
+            implements DecompressListener, ActivateListener {
     private static final String TAG = "XWalkActivity";
+    private static final String META_XWALK_ENABLE_DOWNLOAD_MODE = "xwalk_enable_download_mode";
 
     private Activity mActivity;
     private XWalkDialogManager mDialogManager;
@@ -26,16 +31,54 @@ public class XWalkActivityDelegate
     private boolean mIsXWalkReady;
     private boolean mBackgroundDecorated;
     private boolean mWillDecompress;
+    private final boolean mIsDownloadMode;
 
     public XWalkActivityDelegate(Activity activity,
             Runnable cancelCommand, Runnable completeCommand) {
         mActivity = activity;
         mCancelCommand = cancelCommand;
         mCompleteCommand = completeCommand;
+        mIsDownloadMode = isDownloadModeEnabled();
 
-        mDialogManager = new XWalkDialogManager(mActivity);
-        mXWalkUpdater = new XWalkUpdater(this, mActivity, mDialogManager);
+        if (mIsDownloadMode) {
+            mXWalkUpdater = new XWalkUpdater(
+                new XWalkBackgroundUpdateListener() {
+                    @Override
+                    public void onXWalkUpdateStarted() {
+                    }
 
+                    @Override
+                    public void onXWalkUpdateProgress(int percentage) {
+                    }
+
+                    @Override
+                    public void onXWalkUpdateCancelled() {
+                        mCancelCommand.run();
+                    }
+
+                    @Override
+                    public void onXWalkUpdateFailed() {
+                        mCancelCommand.run();
+                    }
+
+                    @Override
+                    public void onXWalkUpdateCompleted() {
+                        XWalkLibraryLoader.startActivate(XWalkActivityDelegate.this, mActivity);
+                    }
+                },
+                mActivity);
+        } else {
+            mDialogManager = new XWalkDialogManager(mActivity);
+            mXWalkUpdater = new XWalkUpdater(
+                new XWalkUpdateListener() {
+                    @Override
+                    public void onXWalkUpdateCancelled() {
+                        mCancelCommand.run();
+                    }
+                },
+                mActivity,
+                mDialogManager);
+        }
         XWalkLibraryLoader.prepareToInit(mActivity);
     }
 
@@ -45,6 +88,10 @@ public class XWalkActivityDelegate
 
     public boolean isSharedMode() {
         return mIsXWalkReady && XWalkLibraryLoader.isSharedLibrary();
+    }
+
+    public boolean isDownloadMode() {
+        return mIsDownloadMode;
     }
 
     public void setXWalkApkUrl(String url) {
@@ -103,7 +150,7 @@ public class XWalkActivityDelegate
     public void onActivateFailed() {
         mIsInitializing = false;
 
-        if (mXWalkUpdater.updateXWalkRuntime()) {
+        if (mXWalkUpdater.updateXWalkRuntime() && !mIsDownloadMode) {
             // Set the background to screen_background_dark temporarily if the default background
             // is null in order to avoid the visual artifacts around the alert dialog
             Window window = mActivity.getWindow();
@@ -117,7 +164,9 @@ public class XWalkActivityDelegate
 
     @Override
     public void onActivateCompleted() {
-        if (mDialogManager.isShowingDialog()) mDialogManager.dismissDialog();
+        if (mDialogManager != null && mDialogManager.isShowingDialog()) {
+            mDialogManager.dismissDialog();
+        }
 
         if (mBackgroundDecorated) {
             Log.d(TAG, "Recover the background");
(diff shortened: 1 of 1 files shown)


XiaosongWei:
@lincsoon @halton Please take a look, Thanks!

halton:
I will leave the code review to @lincsoon 

Do we have JIRA number related? If so, please add in git log, related to or BUG=

XiaosongWei:
@halton No, this is a sub feature implementation of the download mode feature. Should we create JIRAs for all sub features of a big feature?

halton:
If you have download mode JIRA, you can mark this PR as Related to `XWALK-xxx`, that is what I meant.

XiaosongWei:
ping @lincsoon

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
```diff
@@ -128,4 +188,15 @@ protected void onResume() {
         super.onResume();
         mActivityDelegate.onResume();
     }
+
+    private boolean isDownloadMode() {
```
lincsoon: Please keep `XWalkActivity` as simple as possible and move all specific process to `XWalkActivityDelegate`, because Cordova plugin wants to share same functions with `XWalkActivity` through `XWalkActivityDelegate`.

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -10,10 +10,40 @@
...
 import org.xwalk.core.XWalkLibraryLoader.DecompressListener;
-import org.xwalk.core.XWalkUpdater.XWalkUpdateListener;
 
 public class XWalkActivityDelegate
-            implements DecompressListener, ActivateListener, XWalkUpdateListener {
+            implements DecompressListener, ActivateListener {
+    /**
+     * Interface used to listen to XWalk background update events
+     */
+    public interface XWalkUpdateListener {
```
lincsoon: I can't see the necessity of this interface, most of it aren't being used by other classes, and it could confuse other developers by the same name with `XWalkUpdater`

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -66,29 +144,33 @@ public void onResume() {
 
     @Override
     public void onDecompressStarted() {
-        mDialogManager.showDecompressProgress(new Runnable() {
```
lincsoon: If the library is not compressed, `onDecompressStarted()` won't be triggered, `DecompressTask` will jump to `onDecompressComplete` directly. So no need to modify this method because of download mode.

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -66,29 +144,33 @@ public void onResume() {
...
+                    Log.d(TAG, "Cancel by XWalkActivity");
+                    XWalkLibraryLoader.cancelDecompress();
+                }
+            });
+        }
         mWillDecompress = true;
     }
 
     @Override
     public void onDecompressCancelled() {
```
lincsoon: Same as `onDecompressStarted()`

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -66,29 +144,33 @@ public void onResume() {
...
+        if (mUpdateListener == null) {
+            mDialogManager.dismissDialog();
+            mCancelCommand.run();
+        }
     }
 
     @Override
     public void onDecompressCompleted() {
         if (mWillDecompress) {
-            mDialogManager.dismissDialog();
```
lincsoon: It's enough to check `mWillDecompress`, so no need to modify this method

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -104,34 +186,39 @@ public void onActivateFailed() {
...
+                mActivity.getWindow().setBackgroundDrawable(null);
+                mBackgroundDecorated = false;
+            }
 
-    @Override
-    public void onXWalkUpdateCancelled() {
-        mCancelCommand.run();
+            mCompleteCommand.run();
+        } else {
+            mUpdateListener.onXWalkUpdateCompleted();
```
lincsoon: This sequence is confusing.
`XWalkUpdater.onXWalkUpdateCompleted -> onActivateCompleted -> XWalkActivityDelegate.onXWalkUpdateCompleted`

lincsoon:
A few more questions:
1. Should we provide a interface to let user cancel the background update in XWalkActivity?
2. It's okay that download mode and other modes share same complete command. But in other modes, if the update failed, the user will get UI promption. How to deal with this situration in download mode?
3. XWalkActivity is a one-stop solution that doesn't need more interaction with the developers, and it clearly conflict the principle of download mode. So is it necessary to let XWalkActivity support download mode?

runtime/android/core/src/org/xwalk/core/XWalkActivity.java
XiaosongWei: Do you mean Cordova plugin depends on XWalkActivityDelegate directly?

lincsoon: Yes. Please take a look at https://github.com/crosswalk-project/cordova-plugin-crosswalk-webview/blob/master/platforms/android/src/org/crosswalk/engine/XWalkWebViewEngine.java

It maybe helpful for future integration into Cordova plugin

XiaosongWei:
> 1. Should we provide a interface to let user cancel the background update in XWalkActivity?

Won't provide such a API. for more fine grain control, developer should use XWalkInitializer/XWalkUpdate instead.

> 2. It's okay that download mode and other modes share same complete command. But in other modes, if the update failed, the user will get UI promption. How to deal with this situration in download mode?

could be showing a toast and then quit by default.

> 3. XWalkActivity is a one-stop solution that doesn't need more interaction with the developers, and it clearly conflict the principle of download mode. So is it necessary to let XWalkActivity support download mode?

The primary purpose of making XWalkActivity support download mode is to make Web App packaging with make_apk.py support download mode.

lincsoon:
I suspect whether the `onXWalkUpdateProgress` interface in `XWalkActivity` is necessary, as you said, the developers can use `XWalkInitializer` if they want more control.

On the other hand, we need to provide the interface for handling failed error. Because it's background download, we must keep all silent instead of showing something in foreground.

lincsoon:
I think cancel command and complete command are necessary for shared mode, and failed command and complete command are necessary for download mode. There's no need to add extra listener to XWalkActivityDelegate

XiaosongWei:
Moved all changes to XWalkActivityDelegate to support download mode, for update failed case, showing a Toast to notify "Crosswal Runtime Update failed" and then call cancel command to quit.

XiaosongWei:
@lincsoon

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -26,16 +32,57 @@
...
+                    public void onXWalkUpdateCancelled() {
+                        mCancelCommand.run();
+                    }
+
+                    @Override
+                    public void onXWalkUpdateFailed() {
+                        Toast.makeText(mActivity,
+                                "Crosswalk Runtime Update Failed!",
+                                Toast.LENGTH_SHORT).show();
+                        mCancelCommand.run();
```
lincsoon: As mentioned before, showing toast breaks the basic principle "silent" of silent download mode. And we can't just finish the activity when download failed. In shared mode, there's a dialog interacting with the end-user, and all other interactions are interrupted. The cancel command will only be called when the end-user selected to cancel. But in download mode, all things are done in background and the end-user may be input something in foreground. We can't just send a notification and then close the activity without approval. The best solution is letting the developer to handle this kind of situation.

XiaosongWei:
@lincsoon

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -104,25 +151,33 @@ public void onActivateFailed() {
...
 
         if (mXWalkUpdater.updateXWalkRuntime()) {
-            // Set the background to screen_background_dark temporarily if the default background
-            // is null in order to avoid the visual artifacts around the alert dialog
-            Window window = mActivity.getWindow();
-            if (window != null && window.getDecorView().getBackground() == null) {
-                Log.d(TAG, "Set the background to screen_background_dark");
-                window.setBackgroundDrawableResource(android.R.drawable.screen_background_dark);
-                mBackgroundDecorated = true;
+            if (!mIsDownloadMode) {
```
lincsoon: I think `if (mXWalkUpdater.updateXWalkRuntime() && !mIsDownloadMode)` is enough. It reduces changed code

runtime/android/core/src/org/xwalk/core/XWalkActivityDelegate.java
```diff
@@ -104,25 +151,33 @@ public void onActivateFailed() {
...
-
-        if (mBackgroundDecorated) {
-            Log.d(TAG, "Recover the background");
-            mActivity.getWindow().setBackgroundDrawable(null);
-            mBackgroundDecorated = false;
+        if (!mIsDownloadMode) {
+            if (mDialogManager.isShowingDialog())
+                mDialogManager.dismissDialog();
+
+            if (mBackgroundDecorated) {
```
lincsoon: Just `if (mDialogManager != null && mDialogManager.isShowingDialog())` is okay, keep other code unchanged. Because `mBackgroundDecorated` is always false, we don't need to check `mIsDownloadMode` here.

lincsoon:
LGTM

PR #3334, opened 2015-10-29
[Android] Add download mode support in web app packaging

XiaosongWei:
Insert meta-data in AndroidManifest to make web app running in
silent download mode.


XiaosongWei:
@lincsoon @halton Please take a look, thanks!

halton:
LGTM, will merge.

Just let u know, make_apk suppose to be replace by app-tools soon, all new features added to make_apk should make sure @r0b5t4 aware.  My suggestion is to report a app-tools JIRA. @r0b5t4 how do you think.

r0b5t4:
@halton agreed, it would be good if we could have a patch for app-tools also.

PR #3335, opened 2015-10-29
[Android] Add integrity/signature check in download mode

XiaosongWei:
Signature check is enabled by default and it requires that the crosswalk runtime APK must be signed with the same key used in application APK signing. Developer can disable the signature check by inserting a meta-data in AndroidManifest.

Related to XWALK-5121

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
@@ -10,6 +10,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.pm.ApplicationInfo;
+import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
 import android.net.Uri;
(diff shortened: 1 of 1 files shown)


XiaosongWei:
@lincsoon @rakuco @baleboy  @halton Please take a look, Thanks!
@terriko Please help to review from security point of view, Thanks!

runtime/android/core/src/org/xwalk/core/XWalkUpdater.java
```diff
@@ -216,6 +217,11 @@
     private static final String XWALK_CORE_EXTRACTED_DIR = "extracted_xwalkcore";
     private static final String ARCH_QUERY_STRING = "?arch=";
 
+    // Hash for Chromium debug key
+    private static final String XWALK_APK_HASH_ALGORITHM = "SHA-256";
+    private static final String XWALK_APK_HASH_CODE =
+            "32a2fc74d731105859e5a85df16d95f102d85b22099b8064c5d8915c61dad1e0";
```
lincsoon: Please move these signature-related constants to XWalkAppVersion.template

lincsoon:
@baleboy The content of `XWalkCoreVerifier` comes from the shared mode and has already been reviewed by the security team before. Please help to contact some one in security team to review the part about silent download mode.

XiaosongWei:
@baleboy @terriko New signature check implementation: now it requires that the crosswalk runtime APK must be signed with the same key used in application APK signing. Please take a look, Thanks!

terriko:
Are you checking the signatures or the hashes of the apk?  The file should have both a signature and a set of hashes for each file and it looks like you're checking hashes but not the signature.  (It may not matter for this case, I just want to be clear.)

We may be able to bytecompare things directly rather than doing the digest, but I'm not sure if that's a performance improvement or not, although it would avoid hash collisions.

XiaosongWei:
@terriko It's checking the signatures of the APKs. Yes, it's possible to do bytecompare directly and signature.equals() is doing bytecompare internally. 

```
appPackageInfo.signatures[i].equals(runtimePackageInfo.signatures[i])
```

Now I changed it to use bytecompare, please take a look again. Thanks!

terriko:
lgtm!

XiaosongWei:
@lincsoon @halton Please help to merge if no more comments, thanks!

PR #3336, opened 2015-10-30
[Android] Unable to obtain image taken from Camera on Android M

fujunwei:
Android M return an empty Intent rather than a null intent to onActivityResult
from camera activity; updated the function to check for an empty Intent
if the intent is not null.

BUG=XWALK-5076

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -668,7 +668,9 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
 
             // Check that the response is a good one
             if(Activity.RESULT_OK == resultCode) {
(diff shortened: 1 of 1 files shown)


fujunwei:
@lincsoon @axinging PTAL, thanks.

rakuco:
Looks sane. The change seems to come from [here](https://android.googlesource.com/platform/packages/apps/Camera2/+/24be7cc6d138129b4087ef28f114701de54aba3c) (specifically, `StateSavingPicture::onEnter()`).
FWIW, https://codereview.chromium.org/1035003003 fixes the same issue in a slightly different way, but I don't think the differences matter.

rakuco:
If this gets in, we'll probably need it in both crosswalk-16 and crosswalk-15 (cc @baleboy)

baleboy:
ok

lincsoon:
LGTM

PR #3342, opened 2015-11-03
Xwalk_4997 to enable the tv remote control

redColorJ:
BUG=XWALK-4997


redColorJ:
retry to pull request, for the error is about the net status as below:
error: The requested URL returned error: 406 while accessing https://chromium.googlesource.com/chromium/tools/depot_tools.git/info/refs
fatal: HTTP request failed
error: Could not fetch origin

PR #3343, opened 2015-11-03
[Android] Fix the issue about manifest.json was always used.

wuhengzhi:
This patch is to fix manifest.json was always used issue.
In application's folder, if manifest.json was existed, it will be
always used even if it was not specified in command line.
e.g. if manifest.json was existed and app was packaged by "make_apk.py
--package=org.xwalk.test --name=test --app-root=path-to-app
--app-local-path=index.html", this way won't work, manifest.json will be
used.
Refine the code that user can either specify --manifest or must specify
the --app-* options.

BUG=XWALK-5131

(cherry picked from commit 3ba459cef9d4a40a381f40e752e8c54e7b450ca8)

### DIFF
--- a/app/tools/android/customize.py
+++ b/app/tools/android/customize.py
@@ -323,7 +323,7 @@ def SetVariable(file_path, string_line, variable, value):
   shutil.move(temp_file_path, file_path)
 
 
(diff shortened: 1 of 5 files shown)


wuhengzhi:
@rakuco @baleboy Please review and merge, thanks.

PR #3344, opened 2015-11-04
[Android] Public API XWalkUIClient.onJsAlert, XWalkUIClient.onJsConfi…

shaoboyan:
Public API XWalkUIClient.onJsAlert, XWalkUICLient.onJsConfirm and
XWalkUIClient.onJsPropmt for client to customize these dialogs.
Implement test cases for XWalkUIClient.onJsAlert, XWalkUIClient.onJsConfirm,
and XWalkUIClient.onJsPrompt.

BUG=XWALK-4917

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
@@ -381,8 +381,20 @@ public class XWalkUIClientInternal {
     @XWalkAPI
     public void onPageLoadStopped(XWalkViewInternal view, String url, LoadStatusInternal status) {
     }
-
-    private boolean onJsAlert(XWalkViewInternal view, String url, String message,
+    
+    /**
+     * Tell the client to display an alert dialog to the user.
+     * WARN: Please DO NOT override this API and onJavascriptModalDialog API in the 
+     *       same subclass to avoid unexpected behavior. 
+     * @param view the owner XWalkViewInternal instance.
+     * @param url the url of the web page which wants to show this dialog.
+     * @param message the message to be shown.
+     * @param result the callback to handle the result from caller.
+     * @return Whether the client will handle the alert dialog.
+     * @since 6.0
+     */
+    @XWalkAPI
+    public boolean onJsAlert(XWalkViewInternal view, String url, String message,
             XWalkJavascriptResultInternal result) {
         final XWalkJavascriptResultInternal fResult = result;
         AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mContext);
@@ -408,7 +420,19 @@ public class XWalkUIClientInternal {
         return false;
     }
 
-    private boolean onJsConfirm(XWalkViewInternal view, String url, String message,
+    /**
+     * Tell the client to display a confirm dialog to the user.
+     * WARN: Please DO NOT override this API and onJavascriptModalDialog API in the 
+     *       same subclass to avoid unexpected behavior.
+     * @param view the owner XWalkViewInternal instance.
+     * @param url the url of the web page which wants to show this dialog.
+     * @param message the message to be shown.
+     * @param result the callback to handle the result from caller.
+     * @return Whether the client will handle the confirm dialog.
+     * @since 6.0
+     */
+    @XWalkAPI
+    public boolean onJsConfirm(XWalkViewInternal view, String url, String message,
             XWalkJavascriptResultInternal result) {
         final XWalkJavascriptResultInternal fResult = result;
         AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mContext);
@@ -443,8 +467,21 @@ public class XWalkUIClientInternal {
         mDialog.show();
         return false;
     }
-
-    private boolean onJsPrompt(XWalkViewInternal view, String url, String message,
+    
+    /**
+     * Tell the client to display a prompt dialog to the user.
+     * WARN: Please DO NOT override this API and onJavascriptModalDialog API in the 
+     *       same subclass to avoid unexpected behavior.
+     * @param view the owner XWalkViewInternal instance.
+     * @param url the url of the web page which wants to show this dialog.
+     * @param message the message to be shown.
+     * @param defaultValue the default value string. Only valid for Prompt dialog.
+     * @param result the callback to handle the result from caller.
+     * @return Whether the client will handle the prompt dialog.
+     * @since 6.0
+     */
+    @XWalkAPI
+    public boolean onJsPrompt(XWalkViewInternal view, String url, String message,
             String defaultValue, XWalkJavascriptResultInternal result) {
         final XWalkJavascriptResultInternal fResult = result;
         AlertDialog.Builder dialogBuilder = new AlertDialog.Builder(mContext);
--- a/test/android/core/javatests/src/org/xwalk/core/xwview/test/OnJavascriptModalDialogTest.java
+++ b/test/android/core/javatests/src/org/xwalk/core/xwview/test/OnJavascriptModalDialogTest.java
@@ -9,15 +9,22 @@ import android.test.suitebuilder.annotation.SmallTest;
 import org.chromium.base.test.util.Feature;
 
 /**
- * Test suite for OnJavascriptModalDialog().
+ * Test suite for OnJavascriptModalDialog(), OnJsAlert(), OnJsConfirm(), OnJsPrompt().
  */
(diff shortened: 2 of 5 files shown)


shaoboyan:
@fujunwei @lincsoon  PTAL. thx

lincsoon:
LGTM

lincsoon:
LGTM

fujunwei:
LGTM

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
```diff
@@ -381,8 +381,20 @@ public void onPageLoadStarted(XWalkViewInternal view, String url) {
...
+    
+    /**
+     * Tell the client to display an alert dialog to the user.
+     * WARN: Please DO NOT override this API and onJavascriptModalDialog API in the 
+     *       same subclass to avoid unexpected behavior. 
+     * @param view the owner XWalkViewInternal instance.
+     * @param url the url of the web page which wants to show this dialog.
+     * @param message the message to be shown.
+     * @param result the callback to handle the result from caller.
+     * @return boolean Whether the client will handle the confirm dialog.
```
lincsoon: Sorry, I missed one thing. There is no need to write the type of return value. `@return Wether ...` is okay.

test/android/core/javatests/src/org/xwalk/core/xwview/test/XWalkViewTestBase.java
```diff
@@ -83,15 +83,59 @@ public void onScaleChanged(XWalkView view, float oldScale, float newScale) {
         public void onRequestFocus(XWalkView view) {
             mInnerContentsClient.onRequestFocus();
         }
-
+        
+
```
lincsoon: Please recover this place

test/android/core/javatests/src/org/xwalk/core/xwview/test/OnJavascriptModalDialogTest.java
```diff
@@ -9,16 +9,24 @@
...
     @Override
     public void setUp() throws Exception {
         super.setUp();
+    
         mOnJavascriptModalDialogHelper = mTestHelperBridge.getOnJavascriptModalDialogHelper();
+        mOnJsAlertHelper = mTestHelperBridge.getOnJsAlertHelper();
+        mOnJsConfirmHelper = mTestHelperBridge.getOnJsConfirmHelper();
+        mOnJsPromptHelper = mTestHelperBridge.getOnJsPromptHelper();
     }
+
```
lincsoon: extra line break

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
shaoboyan: Done

test/android/core/javatests/src/org/xwalk/core/xwview/test/OnJavascriptModalDialogTest.java
shaoboyan: done

test/android/core/javatests/src/org/xwalk/core/xwview/test/XWalkViewTestBase.java
shaoboyan: Done

shaoboyan:
@lincsoon PTAL

lincsoon:
LGTM

@shaoboyan  Well done for your first pull request for embedding API!

rakuco:
@lincsoon @fujunwei @shaoboyan this change has broken cordova-plugin-crosswalk-webview to me at least:

```
/tmp/ow4n3jj95k12hhb/CIRC/platforms/android/src/org/crosswalk/engine/XWalkCordovaUiClient.java:71: error: onJsAlert(XWalkView,String,String,XWalkJavascriptResult) in XWalkCordovaUiCl
ient cannot override onJsAlert(XWalkView,String,String,XWalkJavascriptResult) in XWalkUIClient
    private boolean onJsAlert(XWalkView view, String url, String message,
                    ^
  attempting to assign weaker access privileges; was public
/tmp/ow4n3jj95k12hhb/CIRC/platforms/android/src/org/crosswalk/engine/XWalkCordovaUiClient.java:89: error: onJsConfirm(XWalkView,String,String,XWalkJavascriptResult) in XWalkCordovaUi
Client cannot override onJsConfirm(XWalkView,String,String,XWalkJavascriptResult) in XWalkUIClient
    private boolean onJsConfirm(XWalkView view, String url, String message,
                    ^
  attempting to assign weaker access privileges; was public
/tmp/ow4n3jj95k12hhb/CIRC/platforms/android/src/org/crosswalk/engine/XWalkCordovaUiClient.java:112: error: onJsPrompt(XWalkView,String,String,String,XWalkJavascriptResult) in XWalkCo
rdovaUiClient cannot override onJsPrompt(XWalkView,String,String,String,XWalkJavascriptResult) in XWalkUIClient
    private boolean onJsPrompt(XWalkView view, String origin, String message, String defaultValue,
                    ^
  attempting to assign weaker access privileges; was public
```

Perhaps there should be a default implementation so that existing code does not break?

rakuco:
@lincsoon @shaoboyan: sorry for coming back to this, but the more I look at the patch the less I understand the need for it.

* The JIRA ticket has zero context, which is something that *always* bothers me. How is anyone supposed to figure out what "JS interface" and 3 lines of API declarations should mean?
* Everything achieved by overriding `onJsAlert()` and friends could also be done with `onJavascriptModalDialog`.
* The warning in the API documentation for the new public functions is very likely not to be followed by users unless they pay a lot of attention to the newly-added documentation.
* `XWalkViewTestBase.java` does exactly what the aforementioned documentation tells users *not* to do.
* It causes ugly patches like crosswalk-project/cordova-plugin-crosswalk-webview#64 when backwards compatibility needs to be maintained (even though it's also possible to fix this specific issue on the other side instead of Crosswalk).

lincsoon:
@rakuco These three APIs are modeled on the methods with the same name in WebChromeClient. `onJavascriptModalDialog` does the same thing with these new APIs but is not intuitive for developer.

shaoboyan:
@rakuco  thx for your review and questions. Here are some explanation.


1. About necessary of public these three API. 

In my opinion, if the function is not our differenciation, keep the same API design  ,which chromium webview have, is important. This will lower developer's learning costs. 

We could assume a situation based on a premise, that developers want to use crosswalk webview  in the same way as chromium webview on common functions (like onJsAlert which both webview offer the same behavior). 

If someone want to override onJsAlert but find there is no onJsAlert could be overridden on crosswalk, he may feel frustration. And after some efforts, he'll find an unknown API named onJavascriptModalDialog may works.  What's more, if he just want to customize 'alert', he may introduce bugs if he didn't invoke onJsPrompt and onJsConfirm in the overriden onJavascriptModalDialog. 

It is not a good experience. And the API onJavascriptModalDialog seems like something could be used properly only by veterans. But in fact, this function is not so critical. 

That's why I think public these three APIs are necessary. In common and not important area, we need to keep the same with upstream.


2. About warning. 

I'd like to emphasize that the problem will appear only when onJavascriptModalDialog and onJsAlert (or other similar APIs) are overridden in the same subclass. The situation seems not so common.

 Let's assume one possible situation, for a developer who is not familiar with crosswalk but use a library which based on crosswalk. The library has overridden onJavascriptModalDialog and has its own logic to replace onJsAlert (so the library do not invoke super.onJsAlert). 

Now the developer has its own reason to customize "alert" behavior, and he try onJsAlert, yes, it passed compile. But, when the app run, he may confused why it not works? (Remember that the library has overridden onJavascriptModalDialog and not invoked super.onJsAlert). 

How to solve this situation? one of the way is to deprecate onJavascriptModalDialog but it will caused compatibility problem. so we take the way to add warning and after read it, customer may locate where the problem is. (in this case, developer will know onJavascriptModalDialog API through warning, and choose to learn and override onJavascriptModalDialog correctly).


3. About backwards compatibility. 
since the situation that onJavascriptModalDialog and onJsAlert (or other similar APIs) are overridden in the same subclass are not common as I mentioned above. The backwards compatibility suppose to be no problem. 


4. About test case. 
Yes, that's embarrassed. For I'd like to limit modification scale, I'm in a situation which onJavascriptModalDialog and onJsAlert (or other similar APIs) are overriden in the same subclass. And for not misleading developers, I add comments to describe what I've done.


Thx for your reading and sorry for my poor English.

PR #3345, opened 2015-11-04
[Android] Respect the system setting for password echoing.

fujunwei:
BUG=XWALK-4298


rakuco:
lgtm

rakuco:
@baleboy: I guess this should also be in crosswalk-16?

baleboy:
yes

PR #3350, opened 2015-11-04
[Android] Only merge JARs which we have built ourselves.

rakuco:
Add a new parameter to merge_jars.py, --build-dir, and exclude JAR files
which lie outside this directory from merging.

The immediate effect is that android-support-v13.jar will not be part of
xwalk_core_library.jar anymore. The dependency on
org.apache.http.legacy.jar was already dropped with chromium-crosswalk
commit 5909ede ("[Android] Stop using org.apache.http classes in
AccessibilityInjector").

We should not bundle android-support-v13.jar in xwalk_core_library.jar,
as it can cause build problems for users of the embedding API that also
pull the same package themselves or via another dependency. Gradle, for
example, produces the following error message:

```
UNEXPECTED TOP-LEVEL EXCEPTION:
com.android.dex.DexException: Multiple dex files define Landroid/support/annotation/AnimRes;
	at com.android.dx.merge.DexMerger.readSortableTypes(DexMerger.java:596)
	at com.android.dx.merge.DexMerger.getSortedTypes(DexMerger.java:554)
	at com.android.dx.merge.DexMerger.mergeClassDefs(DexMerger.java:535)
	at com.android.dx.merge.DexMerger.mergeDexes(DexMerger.java:171)
	at com.android.dx.merge.DexMerger.merge(DexMerger.java:189)
	at com.android.dx.command.dexer.Main.mergeLibraryDexBuffers(Main.java:454)
	at com.android.dx.command.dexer.Main.runMonoDex(Main.java:303)
	at com.android.dx.command.dexer.Main.run(Main.java:246)
	at com.android.dx.command.dexer.Main.main(Main.java:215)
	at com.android.dx.command.Main.main(Main.java:106)
```

BackgroundSyncLauncherService in content/ is currently the only class
part of Crosswalk's build that depends on android-support-v13.jar
because it uses android.support.v4 API (which is bundled in the
support-v13 JAR).

Users of the embedding API must now depend on at least
android-support-v4.jar themselves. It appears to be done by default for
new applications in both Eclipse and Android Studio. We may need to
update our documentation and make the dependency explicit in Maven as
well as cordova-plugin-crosswalk-webview.

Related to: XWALK-5092

### DIFF
--- a/build/android/merge_jars.py
+++ b/build/android/merge_jars.py
@@ -5,6 +5,11 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+"""
+Given a list of JAR files passed via --jars, produced one single JAR file with
+all their contents merged. JAR files outside --build-dir are ignored.
+"""
+
 import optparse
 import os
 import sys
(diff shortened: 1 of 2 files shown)


rakuco:
ping @lincsoon @fujunwei @halton @r0b5t4

fujunwei:
LGTM.

halton:
lgtm and merging.

fujunwei:
Should it be in Crosswalk -16?

halton:
@heke123 please consider port this commit into lite branch.

rakuco:
Yes, I intend to backport all the necessary commits once we are sure everything works.

PR #3357, opened 2015-11-05
Revert "Wifi Direct Experimental API initial contribution"

rakuco:
This reverts commit ef2dad6c381e35eb5424c8d789a6d2b60704a54b.

Reason: breaks `org.xwalk.runtime.client.embedded.test.PauseResumeTest#testPauseAndResume`.

```
E/AndroidRuntime( 9436): FATAL EXCEPTION: main
E/AndroidRuntime( 9436): java.lang.RuntimeException: Unable to resume activity {org.xwalk.runtime.client.embedded.shell/org.xwalk.runtime.client.embedded.shell.XWalkRuntimeClientEmbeddedShellActivity}: java.lang.NullPointerException
E/AndroidRuntime( 9436):        at android.app.ActivityThread.performResumeActivity(ActivityThread.java:2974)
E/AndroidRuntime( 9436):        at android.app.ActivityThread.handleResumeActivity(ActivityThread.java:3003)
E/AndroidRuntime( 9436):        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1408)
E/AndroidRuntime( 9436):        at android.os.Handler.dispatchMessage(Handler.java:108)
E/AndroidRuntime( 9436):        at android.os.Looper.loop(Looper.java:152)
E/AndroidRuntime( 9436):        at android.app.ActivityThread.main(ActivityThread.java:5691)
E/AndroidRuntime( 9436):        at java.lang.reflect.Method.invokeNative(Native Method)
E/AndroidRuntime( 9436):        at java.lang.reflect.Method.invoke(Method.java:511)
E/AndroidRuntime( 9436):        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:793)
E/AndroidRuntime( 9436):        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:560)
E/AndroidRuntime( 9436):        at dalvik.system.NativeStart.main(Native Method)
E/AndroidRuntime( 9436): Caused by: java.lang.NullPointerException
E/AndroidRuntime( 9436):        at android.app.ActivityManagerProxy.registerReceiver(ActivityManagerNative.java:2150)
E/AndroidRuntime( 9436):        at android.app.ContextImpl.registerReceiverInternal(ContextImpl.java:1517)
E/AndroidRuntime( 9436):        at android.app.ContextImpl.registerReceiver(ContextImpl.java:1485)
[...]

PR #3312, opened 2015-10-20
[Android] Update shouldInterceptLoadRequest to Android L style

xzhan96:
As current API misses lots of context, The older
API:shouldInterceptRequest(WebView view, String url)
was deprecated in Android API level 21, and was replaced by
shouldInterceptRequest(WebView view, WebResourceRequest request),
so update XWalk's corresponding API.

However, this new API can be used in all Android version.

BUG=XWALK-3934

PR #3358, opened 2015-11-05
Revert "[Android] Update shouldInterceptLoadRequest to Android L style"

rakuco:
This reverts commit db7728df0b1324ccc450c2d403376140afc1f8a7.

Reason: broke almost all tests because it uses `android.util.ArrayMap` (API Level 19)
whereas we need to support API levels >= 14:

```
W/System.err( 5990): java.lang.NoClassDefFoundError: android.util.ArrayMap
W/System.err( 5990):    at org.xwalk.core.internal.XWalkContentsIoThreadClient.shouldInterceptRequest(XWalkContentsIoThreadClient.java:53)
W/System.err( 5990):    at dalvik.system.NativeStart.run(Native Method)
F/chromium( 5990): [FATAL:jni_android.cc(249)] Check failed: false. Please include Java exception stack in crash report
```

Related to: XWALK-3934


rakuco:
This reverts #3312 so that we can get the tree back to green together with #3357.
cc @xzhan96, @halton.

PR #3359, opened 2015-11-05
Wifi Direct Experimental API initial contribution second try + bugfix + permission fix

astojilj:
…n" + fix

This reverts the revert #2c526ae Revert "Wifi Direct Experimental... " and fixes
 the underlying revert issues: NullPointerException in .WifiDirect.onActivityStateChange
To make the review of this, the only change compared to original is this:
     public void onActivityStateChange(Activity activity, int newState) {
+        if (mReceiver == null) {
+            return;
+        }
         switch (newState) {
             case ActivityState.RESUMED:
                 mActivity = activity;

### DIFF
--- a/app/android/runtime_client_embedded_shell/AndroidManifest.xml
+++ b/app/android/runtime_client_embedded_shell/AndroidManifest.xml
@@ -31,7 +31,9 @@
   <uses-sdk android:minSdkVersion="14" android:targetSdkVersion="21" />
   <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
   <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
+  <uses-permission android:name="android.permission.CHANGE_NETWORK_STATE" />
   <uses-permission android:name="android.permission.ACCESS_WIFI_STATE" />
+  <uses-permission android:name="android.permission.CHANGE_WIFI_STATE" />
   <uses-permission android:name="android.permission.CAMERA"/>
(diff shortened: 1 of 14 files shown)


astojilj:
Re revert with fix for the NPE causing the revert.#3357

@jondong, I'm submitting full patch for re-review since @r0b5t4  raised concern about permissions in default template here: #3357

CC @halton @baleboy @rakuco

halton:
I'm confused, this PR is revert of revert? Any new change compared with PR #3357?

astojilj:
Hello.
@halton, It is the same code as the SHA already in the tree, with
difference in marked 3 lines, of course with different SHA here.

halton:
Then I'd suggest to remove "Rerevert for Revert" in log message. And please modified three lines seems a bug fixing, normally we submit following PR to fix instead of "revert and amend and submit".

astojilj:
@halton @rakuco 
Hope this way is better - revert or revert in first commit (amend that it is already reviewed) and fixes are in second commit. First commit breaks Crosswalk Android x86-64 test, let's see about second.

@jondong please help on reviewing second commit: astojilj/crosswalk@bb18f19. Thanks

astojilj:
@rakuco Maybe you could also help with review of  astojilj/crosswalk@ d838008  - fixes the reason for revert #3357. thanks

astojilj:
@halton, @jondong  Is this looking better now? Split CLs show the original CL issue with Android tests and and the the fix.
Please advise if needed to do something else. Thanks a lot.

rakuco:
thanks!

PR #3366, opened 2015-11-06
[Windows] Support of JavaScript dialogs.

darktears:
Add a JavaScript dialog class which can be implemented in
future platforms like Linux or Mac, where one would just
provide the _platform to put the implementation.

Dialogs probably can be made nicer looking like in Chrome.

BUG=XWALK-5594


darktears:
@pozdnyakov some review love...

runtime/browser/runtime.cc
```diff
@@ -224,7 +225,10 @@ void Runtime::DidNavigateMainFramePostCommit(
 
 content::JavaScriptDialogManager* Runtime::GetJavaScriptDialogManager(
     content::WebContents* web_contents) {
-  return NULL;
+  if (!javascript_dialog_manager_.get()) {
```
pozdnyakov: get() is not needed

runtime/browser/runtime_javascript_dialog.h
```diff
@@ -0,0 +1,45 @@
...
+      const content::JavaScriptDialogManager::DialogClosedCallback& callback);
+  ~RuntimeJavaScriptDialog();
+
+  void Cancel();
+
+ private:
+  RuntimeJavaScriptDialogManager* manager_;
+  RuntimeJavaScriptDialogManager::DialogClosedCallback callback_;
+
+#if defined(OS_WIN)
```
pozdnyakov: better make a hierarchy, RuntimeJavaScriptDialogWin, and a factory method

pozdnyakov:
lgtm

PR #3368, opened 2015-11-09
[Android] Enable touch event intercept through onTouchListener

XiaosongWei:
Enable touch event intercept through onTouchListener on XWalkView by
installing the listener to ContentView.

BUG=XWALK-3888

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -24,6 +24,7 @@ import android.view.inputmethod.InputConnection;
 import android.view.MotionEvent;
 import android.view.SurfaceView;
 import android.view.ViewGroup;
+import android.view.View.OnTouchListener;
 import android.webkit.ValueCallback;
 import android.webkit.WebResourceResponse;
 import android.widget.FrameLayout;
@@ -657,6 +658,10 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
 
     public boolean onTouchEvent(MotionEvent event) {
         return mContentViewCore.onTouchEvent(event);
+    }
+
+    public void setOnTouchListener(OnTouchListener l) {
+        mContentView.setOnTouchListener(l);
     }
 
     //--------------------------------------------------------------------------------------------
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -22,6 +22,7 @@ import android.util.Log;
 import android.view.KeyEvent;
 import android.view.SurfaceView;
 import android.view.ViewGroup;
+import android.view.View.OnTouchListener;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
 import android.view.MotionEvent;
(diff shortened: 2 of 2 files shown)


XiaosongWei:
@lincsoon @halton @axinging @fujunwei

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1312,4 +1314,12 @@ public void onScrollChangedDelegate(int l, int t, int oldl, int oldt) {
               preWrapperLines = {"onFocusChanged(gainFocus, direction, previouslyFocusedRect);"})
     public void onFocusChangedDelegate(boolean gainFocus, int direction, Rect previouslyFocusedRect) {
     }
+
+    // Override XWalkView.setOnTouchListener to install the listener to ContentView
+    // therefore touch event intercept through onTouchListener is available on XWalkView.
+    @Override
+    @XWalkAPI
+    public void setOnTouchListener(OnTouchListener l) {
```
lincsoon: Please add description to this interface and use a readable name for the argument

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -130,6 +131,10 @@ public XWalkContent(Context context, AttributeSet attrs, XWalkViewInternal xwVie
         XWalkPreferencesInternal.load(this);
     }
 
+    public View getContentView() {
```
lincsoon: Why cast to `View`?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
axinging: Google Android has such a name style too.  
Now that this API is from Android View, it was there since android started.  Should we add @since x.0? @lincsoon

XiaosongWei: Xing is right. I just copy and paste it from Android SDK doc.
```public void setOnTouchListener (View.OnTouchListener l)```.
Since it's a Android View API and it is always there but just not working, adding @Since xx may cause confusion, and it's not exposed in XWalkView API doc as no javadoc  tag is added here.

axinging:
LGTM

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
XiaosongWei: Just because I want to use it like a View only and then hide  the ContentView from XWalkViewInternal.

lincsoon: It's strange that getting a `View` object but the function name is `getContentView`. In this case, I think it's better to call `mContent.setOnTouchListener()`, and move the real implementation into `XWalkContent`

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
lincsoon: Actually it does appear in java doc as long as you add `@XWalkAPI` onto it. How ever, if you remove this mark, the reflection layer won't support this interface. I agree that it's not necessary add `@since` is it's not an embedding API. You can add comment like `See android.view.View.OnTouchListener` instead of leaving the description blank.

PR #3370, opened 2015-11-09
Keep single instance of ApplicationDataCache

pozdnyakov:
Rationals:
- The ApplicationDataCache is supposed to be shared.
- This avoids crashes on exit if there are WebConents instances apart from Application (like it happens during presentation sessions)


pozdnyakov:
@kenchris @darktears

PR #3371, opened 2015-11-10

axinging:
This is supported to work on Android >= Lollipop (5.0). Otherwise it will always show a gray background in the task switcher.

BUG=XWALK-4305

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
@@ -86,6 +86,7 @@ public class XWalkUIClientInternal {
     public void onDidChangeThemeColor(XWalkViewInternal view, int color) {
         if (view == null || view.getActivity() == null) return;
         ApiCompatibilityUtils.setStatusBarColor(view.getActivity().getWindow(),color);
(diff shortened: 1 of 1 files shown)


axinging:
@fujunwei  @halton  @lincsoon

lincsoon:
LGTM

rakuco:
> For Android version older than 5.0, this doesn't work.

Do you mean "older" or "newer"? The title says this change makes it work on Android >= 5.0, but the contents then say this is supposed to make it work on versions older than 5.0 (< 5.0).

axinging:
Sorry for this confusion, older means  >5.0. I used a time since android begins. So android 2.2 or 4.0 is much older than android 5.0

fujunwei:
LGTM

rakuco:
> Sorry for this confusion, older means >5.0. I used a time since android begins. So android 2.2 or 4.0 is much older than android 5.0

2.2 is older than 5.0, and it is also < 5.0. Sorry for being picky here, but otherwise this will get very confusing in the future if someone checks this pull request or the commit message.

If I understood the intent of your commit correctly, you need to update your commit and pull request messages to say something like "this is supported on Android >= Lollipop (5.0), where it would otherwise always show a gray background in the task switcher". This avoids confusion and also explains why you are doing what you are doing.

axinging:
Thanks @rakuco, updated.

rakuco:
Thanks! Please backport it to 16.

PR #3372, opened 2015-11-10

axinging:
For Android version older than 5.0, this doesn't work.

BUG=XWALK-4305


darktears:
@baleboy do we really want to backport? The patch is simple but it's a stable version thus only security fixes and important/critical fixes.

PR #3373, opened 2015-11-10

axinging:
For Android version older than 5.0, this doesn't work.

BUG=XWALK-4305


darktears:
We do not release any version 14. See https://crosswalk-project.org/documentation/downloads.html

rakuco:
not only that, but it looks like you are not following the mailing list: https://lists.crosswalk-project.org/pipermail/crosswalk-dev/2015-October/003136.html

PR #3375, opened 2015-11-10

axinging:
This is supported to work on Android >= Lollipop (5.0). Otherwise it will always show a gray background in the task switcher.

BUG=XWALK-4305


axinging:
@rakuco

PR #3376, opened 2015-11-10
[Android] Add support to Map type parameter

xzhan96:
Currently, reflection does not support Collection type parameter
or return value, such case happens in some API upgrade.

BUG=XWALK-5631

### DIFF
--- a/tools/reflection_generator/java_method.py
+++ b/tools/reflection_generator/java_method.py
@@ -10,6 +10,10 @@ from collections import OrderedDict
 from string import Template
 
 def ConvertClassExpressionToClassType(class_name):
+  """ Turn "Map<String, String>" to Map.class. """
+  generic_re = re.compile('[a-zA-z0-9]+(\<[a-zA-Z0-9]+,\s[a-zA-z0-9]+\>)')
+  if re.match(generic_re, class_name):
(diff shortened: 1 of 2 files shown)


xzhan96:
@axinging @lincsoon @XiaosongWei @wuhengzhi

xzhan96:
@halton please help to review, thanks.

lincsoon:
LGTM

xzhan96:
@wuhengzhi 
1. because python need difference separator, so Map<String, String> has different regular rule HashMap<String>
2. Generic type includes Collection type, and other format, so I describe it as above.

tools/reflection_generator/java_method.py
```diff
@@ -10,7 +10,11 @@
 from string import Template
 
 def ConvertClassExpressionToClassType(class_name):
+  """ Turn "Map<String, String>" to Map.class. """
+  generic_re = re.compile('[a-zA-z0-9]+(\<[a-zA-Z0-9]+,\s[a-zA-z0-9]+\>)')
   """ Turn "final HashMap<String>" to HashMap.class. """
```
XiaosongWei: This comment should go with the original statement. others LGTM

xzhan96:
@lincsoon could you help to merge ?

PR #3377, opened 2015-11-10
[Android] Put the call to onPageFinished in message loop

lincsoon:
Originally, onPageFinished is invoked directly. In contrast,
onPageStarted is invoked through message loop. So the order of these two
methods maybe reversed.

BUG=XWALK-4849

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClient.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClient.java
@@ -66,7 +66,7 @@ abstract class XWalkContentsClient extends ContentViewClient {
 
         @Override
         public void didStopLoading(String url) {
-            onPageFinished(url);
+            mCallbackHelper.postOnPageFinished(url);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


lincsoon:
@axinging @fujunwei @XiaosongWei @xzhan96 @wuhengzhi

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClient.java
```diff
@@ -66,7 +66,7 @@ public void didChangeThemeColor(int color) {
 
         @Override
         public void didStopLoading(String url) {
-            onPageFinished(url);
```
axinging: Does the following call happens at the same thread? If true, this fix is ok.
XWalkWebContentsObserver::didStopLoading
InterceptNavigationDelegateImpl::shouldIgnoreNavigation

lincsoon: Both of them run in the UI thread

wuhengzhi:
LGTM, thanks.

axinging:
LGTM

fujunwei:
LGTM

PR #3380, opened 2015-11-10
Keep single instance of ApplicationDataCache

pozdnyakov:
Rationals:

- The ApplicationDataCache is supposed to be shared.
- This avoids crashes on exit if there are WebConents instances apart from
Application (like it happens during presentation sessions)

### DIFF
--- a/application/browser/application_protocols.cc
+++ b/application/browser/application_protocols.cc
@@ -163,8 +163,8 @@ class URLRequestApplicationJob : public net::URLRequestFileJob {
 };
 
 // This class is a thread-safe cache of active application's data.
-// This class is used by ApplicationProtocolHandler as it lives on IO thread
-// and hence cannot access ApplicationService directly.
+// This class is used by ApplicationProtocolHandler which lives on
+// IO thread and hence cannot access ApplicationService directly.
 class ApplicationDataCache : public ApplicationService::Observer {
  public:
   scoped_refptr<ApplicationData> GetApplicationData(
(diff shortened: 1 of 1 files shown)


pozdnyakov:
@darktears ptal

rakuco:
Isn't it problematic that `cache_` won't be deleted?

pozdnyakov:
> Isn't it problematic that cache_ won't be deleted?
Its lifetime is the same as the process lifetime, so it's fine.

rakuco:
It's possibly a good idea to add a comment to `ApplicationDataCache` explaining that it's supposed to be `static` and people shouldn't assume its destructor will be run (or add things there that need to be destroyed).

pozdnyakov:
>It's possibly a good idea to add a comment to ApplicationDataCache explaining that it's supposed to be static and people shouldn't assume its destructor will be run (or add things there that need to be destroyed).

done

rakuco:
hmm, I was actually thinking of something more verbose :-) for example, a comment on top of `ApplicationDataCache` itself (like Chromium documents some of its classes) saying that the class is leaked and shouldn't have a destructor or things that need to be destroyed. plus s/leacked/leaked/.

perhaps you could even add a destructor that asserts so that we're always sure it's used statically?

pozdnyakov:
@rakuco I've slightly changed the interface PTAL

application/browser/application_protocols.cc
```diff
@@ -189,21 +189,39 @@ class ApplicationDataCache : public ApplicationService::Observer {
...
     : public net::URLRequestJobFactory::ProtocolHandler {
  public:
   explicit ApplicationProtocolHandler(ApplicationService* service) {
-    DCHECK(service);
-    DCHECK(BrowserThread::CurrentlyOn(BrowserThread::UI));
-    // ApplicationProtocolHandler lives longer than ApplicationService,
-    // so we do not need to remove cache_ from ApplicationService
-    // observers list.
-    service->AddObserver(&cache_);
+      ApplicationDataCache::CreateIfNeeded(service);
```
rakuco: I think the indentation is wrong here.

rakuco:
yes! this is much better :cool: 
lgtm with the indentation nit

PR #3384, opened 2015-11-11
[Android] Fix the issue of cordova 4.x app launch crash.

wuhengzhi:
Provide a separate function for SurfaceView to set visibility.
Keep XWalkView.setVisibility() as an embedding API, the base class's
setVisibility will be called even if xwalk is not ready.
Mark "setSurfaceViewVisibility()" as reserved function to initial later.
In XWalkViewInternal.java, "setVisibility" is just an auxiliary function
to generate XWalkView.setVisibility().

BUG=XWALK-5700

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -25,6 +25,7 @@ import android.view.ViewGroup;
 import android.view.inputmethod.EditorInfo;
 import android.view.inputmethod.InputConnection;
 import android.view.MotionEvent;
+import android.view.View;
 import android.webkit.ValueCallback;
 import android.widget.FrameLayout;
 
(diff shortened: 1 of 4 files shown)


rakuco:
ping @lincsoon

wuhengzhi:
In this PR, provide a separate function for SurfaceView, but I think we should expose {{getSurfaceView()}} to let user to set visibility. As far as I know, two customers has this requirement.

wuhengzhi:
@lincsoon @axinging @fujunwei PTAL, thanks.

fujunwei:
LGTM

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1088,10 +1089,21 @@ public void clearFormData() {
...
-        super.setVisibility(visibility);
+    }
+
+    /**
+     * Set the enabled state of SurfaceView.
+     * @param visibility One of VISIBLE, INVISIBLE, or GONE.
+     * @since 6.0
+     */
+    @XWalkAPI(reservable = true)
+    public void setSurfaceViewVisibility(int visibility) {
```
axinging: setVisibility works well with TextureView?

wuhengzhi: Yes.

axinging:
Please update the API documents.

LGTM.

wuhengzhi:
Just close this PR temporally, will open it when updated the code for M46 rebase.

rakuco:
Can you explain how this and the rebase are related? We do not even have a schedule for when M46 will land, and we cannot stay with a P1 release blocker in the tree indefinitely. Either this change lands after review, or the original pull request gets reverted.

wuhengzhi:
I am fixing the error for XWALK-5241, need to update the ```next``` code, but local code based on my github, not crosswalk-project/crosswalk.
I need to update the master code in my github, but this PR is based on master code.
I have updated the ```next``` code, now this PR can be reopend.

rakuco:
that's why you should use separate branches for each pull request...

wuhengzhi:
Yes, I did it each time before, but this PR is fault.

PR #3389, opened 2015-11-12
[Android] Fix test error of evaluating javascript

lincsoon:
From 14ea977ff2e300fd131d3ca8ff6f5311f5fd1df8 in Chromium, there is a
checkpoint before executing JavaScript. The URLs that
ExecuteJavaScript can be invoked on are restricted to chrome-controlled
URLs.

This change blocks all xwalk test with evaluating JavaScript. This patch
is to add additional interface for evaluating javascript in test mode.

Related bug: XWALK-5241

### DIFF
--- a/runtime/browser/xwalk_browser_main_parts_android.cc
+++ b/runtime/browser/xwalk_browser_main_parts_android.cc
@@ -18,6 +18,7 @@
 #include "content/public/browser/android/compositor.h"
 #include "content/public/browser/browser_thread.h"
 #include "content/public/browser/cookie_store_factory.h"
+#include "content/public/browser/render_frame_host.h"
 #include "content/public/common/content_switches.h"
 #include "content/public/common/result_codes.h"
 #include "net/android/network_change_notifier_factory_android.h"
@@ -187,6 +188,12 @@ void XWalkBrowserMainPartsAndroid::PreMainMessageLoopRun() {
   xwalk_runner_->PreMainMessageLoopRun();
 
   extension_service_ = xwalk_runner_->extension_service();
+
+  // Due to http://code.google.com/p/chromium/issues/detail?id=507809,
+  // it's not possible to inject javascript into the main world by default.
+  // So lift this limitation here to enable XWalkView.evaluateJavaScript
+  // to work.
+  content::RenderFrameHost::AllowInjectingJavaScriptForAndroidWebView();
 
   // Prepare the cookie store.
   base::FilePath user_data_dir;


lincsoon:
@rakuco @halton Please help to review

runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/presentation/XWalkPresentationContent.java
```diff
@@ -56,7 +56,7 @@ public void onPageLoadStarted(XWalkViewInternal view, String url) {
                     String script = "navigator.presentation.session = new navigator.presentation.PresentationSession(";
                     script += mPresentationId;
                     script += ");";
-                    view.evaluateJavascript(script, null);
+                    view.evaluateJavascriptForTests(script, null);
```
halton: Presentation API is not for testing, I do not understand why it is needed here.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -315,6 +315,21 @@ public void handleJavaScriptResult(String jsonResult) {
...
+        if (fCallback != null) {
+            coreCallback = new JavaScriptCallback() {
+                @Override
+                public void handleJavaScriptResult(String jsonResult) {
+                    fCallback.onReceiveValue(jsonResult);
+                }
+            };
+        }
+        mContentViewCore.getWebContents().evaluateJavaScriptForTests(script, coreCallback);
+    }
```
halton: Above added code seems totally dup with evaluateJavascript except two final variables. Can we re-use code somehow?

And please add comment to mention this new API for testing only, and why it resolve the problem.

rakuco:
you need to be more verbose in your commit/PR messages. there is no context explaining what Chromium change caused this and how this code is basically a copy of the original Chromium commit. I had to do the digging myself to figure out this is https://codereview.chromium.org/1123783002 and https://code.google.com/p/chromium/issues/detail?id=507809. please include as much information as possible in your commit message, as we try to keep most of it in the squashed commit when rolling a new Chromium milestone.

rakuco:
with that said:
* Do we not need to call `content::RenderFrameHost::AllowInjectingJavaScriptForAndroidWebView()`?
* Aren't we adding a new public API call to XWalkView? That sounds bad, as nothing prevents a user from calling it.

lincsoon:
@rakuco Thanks for your investigation! The point is from now on, there is `CanExecuteJavaScript()` in `render_frame_host_impl.cc`. We have several ways to let unit test pass this check:

1. Add `content::RenderFrameHost::AllowInjectingJavaScriptForAndroidWebView()` to `xwalk_browser_main_parts_android.cc` or somewhere else.
2. As this PR.

For the first method, the problem is that `CanExecuteJavaScript()` will always return true. This is equivalent to disabling this check.

The reason of second method is that since the infrastructure is providing two ways(normal and for test) to execute JavaScript, we expose these interfaces to external and let the user to choose which way to use.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
lincsoon: Because there are only a few duplicated code and only two methods are sharing them, I think there is no need to add a private member method specially.

I added comment in `XWalkViewInternal`.

halton:
LGTM, I'll wait for @rakuco 's lgtm before merging.

rakuco:
Thanks for being more verbose with the commit/PR messages above. What I don't get is the part where you say this is only for "test mode". Isn't this API going to be exported and public?

lincsoon:
Sorry for everyone who has helped to review this PR, I read the commit from Chromium again and changed to use method 1 above, because:

1. There are still a few concern about exposing new API for tests, and we still don't have explicit reasons to do this except for unit test. Since we can't change the interface easily once we published it, it's safe to keep current change internally and get this issue fixed ASAP. We can discuss whether to add a new API later if necessary.

2. Keep the same implementation with Android WebView(in aw_browser_main_parts.cc:98). Since default WebView is always allowed to evaluate JavaScript, I think there won't be security issues.

rakuco:
there's absolutely no need to apologize, I'm really glad to see you've put real effort into this investigation and wasn't afraid to go for a different solution :-)

runtime/browser/xwalk_browser_main_parts_android.cc
```diff
@@ -188,6 +189,8 @@ void XWalkBrowserMainPartsAndroid::PreMainMessageLoopRun() {
 
   extension_service_ = xwalk_runner_->extension_service();
 
+  content::RenderFrameHost::AllowInjectingJavaScriptForAndroidWebView();
```
rakuco: Just one request before merging: can you please add a comment here mentioning crbug.com/507809 and saying we need to call this because we provide `evaluateJavaScript` like the Android WebView?

lincsoon:
@rakuco Done! Please help to merge

PR #3391, opened 2015-11-12
[Lite] Add default values for new introduced flags

heke123:
Without default values, the first exec of gyp_xwalk triggered by
gclient sync would fail.

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -7,6 +7,19 @@
     'use_webui_file_picker%': 0,
     'disable_bundled_extensions%': 0,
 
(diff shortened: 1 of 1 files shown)


heke123:
@PeterWangIntel  @halton

heke123:
some previously added flags exist already in src/build/common.gypi.  Let's add our new introduced flags in xwalk/build/common.gypi, where the macros are defined.

PeterWangIntel:
LGTM, as face to face review.
Thank you very much.

PR #3393, opened 2015-11-12
[Windows] Presentation session security policy

pozdnyakov:
Presentation session should take into account the CSP rules applied for the parent application.


pozdnyakov:
@darktears @rakuco

PR #3395, opened 2015-11-12
Remove tools/build.

rakuco:
The files FILES.cfg and TESTS present in different subdirectories were
only used in the very beginning of the project when internal Buildbot
builders used to store the build results.

They have not been used for many years, so garbage-collect the entire
directory to avoid confusion.


darktears:
LGTM>

PR #3397, opened 2015-11-13
[Android] Generate the Maven POM files and depend on support-v4.

rakuco:
This patch set contains two commits. The first one adds all the required
plumbing for generating the Maven POM files we have for xwalk_core_library and
xwalk_shared_library as part of our build.

Quoting the commit message:

> Currently, the Maven POM XML files part of the Crosswalk Maven repository
> in download.01.org are generated by Maven itself when we call
> `mvn deploy:deploy-file` with `-DgeneratePom=true`.
>
> While it works for most cases, it is not a flexible solution: if one needs
> to include more information in the POM file, such as a dependency
> declaration, the POM file cannot be automatically generated.
>
> This patch adds template POM files for both xwalk_core_library and
> xwalk_shared_library that are preprocessed during the build and turn into
> POM files that can be used with Maven's `install-file` and
> `deploy-file` commands.
>
> By default, the artifact IDs used in the POM files are
> `xwalk_core_library_canary` and `xwalk_shared_library_canary`, but their
> values can be changed with the `xwalk_core_library_artifact_id` and
> `xwalk_shared_library_artifact_id` gyp variables.
>
> The new POM files will not be automatically used when building new
> Crosswalk releases just yet; the release scripts need to be modified first,
> but that will be done shortly.

The second commit leverages this infrastructure to add a dependency on the
support-v4 Android support library to the xwalk_core_library artifact, as we
have stopped bundling external JARs since #3350.

Related to: XWALK-5092

### DIFF
--- a/build/android/maven_pom.gypi
+++ b/build/android/maven_pom.gypi
@@ -1,0 +1,52 @@
+# Copyright (c) 2015 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file is meant to be included into a target to provide a rule to generate
+# a Maven POM XML file that can later be used in Maven repositories.
+# It takes a template POM and substitutes the values @ARTIFACT_ID@ and
+# @ARTIFACT_VERSION@ with the variables provided to this file.
+#
+# Usage:
+# {
+#   'target_name': 'my_artifact_pom_gen',
+#   'type': 'none',
+#   'variables': {
+#     'pom_input': 'path/to/my_artifact.pom.xml.in',
+#     'pom_output': '<(PRODUCT_DIR)/my_artifact.pom.xml',
+#     'artifact_id': 'foo_bar_canary',
+#     'artifact_version': '1.2.3.4',
+#   }
+#   'includes': ['path/to/this/gypi/file'],
+# }
+#
+# Required variables:
+#  pom_input - Path to the POM file that will be processed.
+#  pom_output - Path to the generated POM file.
+#  artifact_id - Value for the <artifactId> field in the POM file.
+#  artifact_version - Value for the <version> field in the POM file.
+
+{
+  'actions': [
+    {
+      'action_name': 'generate_pom_file',
+      'message': 'Generating <(pom_output)',
+      'variables': {
+        'version_py': '<(DEPTH)/build/util/version.py',
+      },
+      'inputs': [
+        '<(version_py)',
+        '<(pom_input)',
+      ],
+      'outputs': [
+        '<(pom_output)',
+      ],
+      'action': [
+        'python', '<(version_py)', '-i', '<(pom_input)',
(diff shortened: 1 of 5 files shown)


rakuco:
ping @halton @fujunwei @lincsoon @r0b5t4 @baleboy

rakuco:
@fujunwei: for the billionth time: **be specific**. You did not answer my first question, nor did you say which app you are building, which IDE you are using, how the app is being built or how you checked `classes.dex`, so I had to waste my time trying to reproduce from scratch what you had already done. This is not helpful or productive **at all**.

Here's what I did:
* Set up a local Maven repository
  * `cd src/out/Release`
  * `mvn install:install-file -Dfile=xwalk_core_library.aar -DpomFile=xwalk_core_library.pom.xml`
  * This creates `~/.m2/repository/org/xwalk/xwalk_core_library_canary` with all the artifacts there, including the POM file with the dependency on support-v4.
* Used it in a new project
  * Fired up Android Studio 1.4.1
  * Created a new project with the Empty Activity template, targeting devices >= Android 4.0.
  * Removed the dependency on `com.android.support:appcompat-v7:23.1.1` from `app/build.gradle`. Consequently:
    * Removed the `<item>` tags from the theme in `app/src/main/res/values/styles.xml`, as well as the `parent` attribute of the `<style>` tag.
    * Made `MainActivity` inherit from `Activity` instead of `android.support.v7.app.AppCompatActivity`.
  * Started building only a x86 APK:
    * Added a `productFlavors` block to the `app/build.gradle`:
    ```
    productFlavors {
      x86 {
        ndk {
        abiFilters "x86", ""
        }
      }
    }
    ```
    * Added `android.useDeprecatedNdk=true` to the top-level `gradle.properties`.
  * At this point, if I build the APKs with `gradlew assemblex86`, unpack `app/build/outputs/apk/app-x86-debug.apk` and run `dexdump` on the unpacked `classes.dex`, I can see that the only 9 classes listed are from my own project.
  * Started using Crosswalk:
    * Added `mavenLocal()` to the repositories in the top-level `build.gradle`.
    * Added a `compile` dependency on `org.xwalk:xwalk_core_library_canary:17+` to `app/build.gradle`.
    * At this point, if I run `dexdump` again, I can see all classes in `android/support/annotation`, `android/support/v4`, `javax/annotation`, `org/chromium`, `org/xwalk/core` and `SevenZip` in the unpacked `classes.dex`.
    * This is still true if I actually put a `XWalkView` in the app and use it to load some page.

If I remove the dependency from the POM in my local repository and rebuild, I can see that the `android/support/*` classes are gone from `dexdump`'s output.

So again, you need to tell me in a reproducible way what you did differently to reach a different result so I can understand if there really is a problem or not, otherwise there's nothing I can do.

fujunwei:
It works finely when i try it again.

fujunwei:
LGTM

baleboy:
@rakuco merge?

PR #3398, opened 2015-11-13
Revert presentation api

pozdnyakov:
Revert Presentation API  commits as it breaks components build on Win: browser tests do not link


darktears:
LGTM. Sorry.

PR #3399, opened 2015-11-16
[Android] Fix the failed test case for database and domstorage.

wuhengzhi:
Follow the upstream's change, convert the test to isolate. Please
refer to https://codereview.chromium.org/933593003.
Test case usage:
  ./build/android/test_runner.py instrumentation --release \
  --test-apk=XWalkCoreInternalTest --isolate-file-path \
  xwalk/xwalk_view_internal_test_apk.isolate

BUG=XWALK-5241


wuhengzhi:
Have verified on Nexus 5, I will borrow a ZTE Geek to test tomorrow.

rakuco:
Thanks for working on this.

Looking at the upstream CL you've referenced, this change seems to be fixing the symptoms without fully explaining or going after the root cause. That Chromium CL was landed in February, and has thus been present in Crosswalk since version 14. Therefore, it's something else that is causing the non-isolate case to fail, and we need to find out what it is, otherwise we cannot be sure that this patch is doing the right thing or just masking a deeper issue.

wuhengzhi:
After deep into the test script, this PR could be closed.

rakuco:
Feel free to close it then.

PR #3400, opened 2015-11-16
[Win] Presentation API implementation

pozdnyakov:
This patch contains the implementation of the W3C Presentation
API for Crosswalk on Windows.

With the current chromium-crosswalk revision (which is based on M45
upstream version) the Presentation API bindings correspond to
http://www.w3.org/TR/2015/WD-presentation-api-20150701.

Missing API features: there is no bindings for incoming presentaion
sessions and hence message exchange is also not available.

Missing implementation features (TBD in the following patches):
- CSP rules for presentation URL
- UI for presentation window selection

### DIFF
--- a/runtime/browser/xwalk_content_browser_client.cc
+++ b/runtime/browser/xwalk_content_browser_client.cc
@@ -15,6 +15,7 @@
 #include "content/public/browser/browser_ppapi_host.h"
 #include "content/public/browser/child_process_data.h"
 #include "content/public/browser/client_certificate_delegate.h"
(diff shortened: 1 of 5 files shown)


pozdnyakov:
@darktears the reason for https://github.com/crosswalk-project/crosswalk/pull/3398 was the forgotten 'DisplayInfoManager::self_' static member declaration, for some reason VS did not complain me about that. Fixed now. Thanks

darktears:
I suspected that. LGTM again.

PR #3403, opened 2015-11-17
[Android] Put the call to onPageFinished in message loop

lincsoon:
Originally, onPageFinished is invoked directly. In contrast,
onPageStarted is invoked through message loop. So the order of these two
methods maybe reversed.

BUG=XWALK-4849
(cherry picked from commit c989414b8cbe9a3001e3d106195c9e8cf81bfff0)


lincsoon:
@rakuco Please help to merge

PR #3405, opened 2015-11-19
[Android] Enable scrollTo/By in XWalkView

XiaosongWei:
Delegate scrollTo and scrollBy methods to ContentView as the
actual scroll is done in native

BUG=XWALK-3112

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -664,6 +664,14 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
         mContentView.setOnTouchListener(l);
     }
 
+    public void scrollTo(int x, int y) {
+        mContentView.scrollTo(x, y);
+    }
+
+    public void scrollBy(int x, int y) {
+        mContentView.scrollBy(x, y);
+    }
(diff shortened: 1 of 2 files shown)


XiaosongWei:
@lincsoon @axinging @fujunwei Please take a look!

lincsoon:
These APIs are not embedding API but inherited from View, so I think there's no need to add Crosswalk-style comments. But are we going to leave the description of all this kind of APIs blank? Any better idea?

lincsoon:
This is what java doc looks like.

```
boolean	saveState(android.os.Bundle outState)
Save current internal state of this XWalkView.

void	scrollBy(int x, int y) 

void	scrollTo(int x, int y) 

void	setAcceptLanguages(java.lang.String acceptLanguages)
Set the accept languages of XWalkView.
```

fujunwei:
But There are nothing in the documentation.
Maybe Those API don't list on the documentation.

fujunwei:
Customer can use those API as View.

XiaosongWei:
What I saw in the API document is:

>
scrollTo
>
public void scrollTo(int x,
            int y)
Overrides: scrollTo in class android.view.View
>
scrollBy
>
public void scrollBy(int x,
            int y)
Overrides: scrollBy in class android.view.View

There's a statement saying that the method is overrided from View's

PR #3406, opened 2015-11-19
[Android] XWalkView should support edit mode in developer tool

xzhan96:
A View is usually in edit mode when displayed within a
developer tool, so isInEditMode() should be used inside
the corresponding XWalkView constructor.

BUG=XWALK-3568

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -229,6 +229,8 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
         initXWalkContent(mContext, null);
     }
 
+    // A View is usually in edit mode when displayed within a developer tool, like Android Studio.
+    // So isInEditMode() should be used inside the corresponding XWalkView constructor.
     /**
      * Constructor for inflating via XML.
      * @param context  a Context object used to access application assets.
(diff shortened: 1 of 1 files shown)


xzhan96:
@axinging @lincsoon @wuhengzhi @XiaosongWei @fujunwei 
please help to review, thanks

axinging:
LGTM

wuhengzhi:
Why other two constructors do not need ```isInEditMode()```?

lincsoon:
```
Subclasses should check the return value of this method to provide different behaviors if their normal behavior might interfere with the host environment
```
Could you explain why adding a surfaceView would interfere with this?

xzhan96:
@wuhengzhi because edit the layout in Android studio will only use this constructor. I don't think it's need to add the other two.

xzhan96:
@lincsoon I have verified, add isInEditMode check before or after adding a surfaceView has the same effect. So which you think is better? and I am not sure if add comments in preWrapperLines, if yes, I will add.
From official explanation, the point is that must check isInEditMode before the class spawns a thread in its constructor, or the drawing code relies on device-specific features, etc.
The cause that not put this check in the XWalkViewInternal (below "super(context, attrs);" ) is that reflection can't parsing it.

XiaosongWei:
LGTM

xzhan96:
@lincsoon please help review and merge, thanks

PR #3408, opened 2015-11-19
DEPS.xwalk: Re-roll v8-crosswalk.

rakuco:
Follow-up to 9d93b7d ("Roll Chromium to 45.0.2454.101, the last M45
stable release").

The original v8-crosswalk pushed with 45.0.2454.101 was faulty and
missing several commits (of our own, not upstream ones )which were
present when we were in .93.

v8-crosswalk's master branch has been re-pushed, and we need to use the
new hash.

Related to: XWALK-5762


darktears:
LGTM

PR #3409, opened 2015-11-19
crosswalk-16: Roll Chromium to 45.0.2454.101, the last M45 stable release.

rakuco:
The most notable change in this roll is that this upstream release was
made after Blink was merged into the Chromium git repository. It also
contains some security fixes, as described in [the release announcement](http://googlechromereleases.blogspot.com/2015/09/stable-channel-update_24.html).

We have inherited this change, and the blink-crosswalk repository is now
deprecated. People `gclient sync`'ing to this commit must pay attention to
gclient's output: the existing blink-crosswalk checkout in
`src/third_party/WebKit` will be moved automatically to a separate
location, and needs to be manually removed.

It is also likely that the `UPSTREAM.blink` file will contain the wrong
value the first time gclient sync is called, so people relying on it
should call gclient sync twice. Release builds are not affected because
they are made from scratch.

Related to: XWALK-5139
BUG=XWALK-5762

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,9 +17,8 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = '39190af3f7fab51baceaeb77365e3f3a2757661b'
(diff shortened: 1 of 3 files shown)


darktears:
LGTM.

PR #3413, opened 2015-11-20
[Android] Move extension registration to file system thread

lincsoon:
The registration of xwalk extension includes file system operations,
while being executed during loading URL which is under UI thread.
In debug mode, this will violate the thread retriction and cause the
app crash. So move this procedure to the thread that dedicates to
file system operation.

BUG=XWALK-5754


lincsoon:
@halton @DonnaWuDongxia Please help to review

PR #3414, opened 2015-11-20
[Android] Fix UnsupportedOperationException caused by setLayerType

XiaosongWei:
setLayerType may be called before XWalk is ready and then UnsupportedOperationException is thrown. As we only support hardware acceleration in XWalkView, so setLayerType should be noop.

BUG=XWALK-5021, BUG=XWALK-5047

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -795,17 +795,14 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
         mContent.setBackgroundColor(color);
     }
 
-    /**
-     * override setLayerType
-     */
+    // We can't let XWalkView's setLayerType call to this via reflection as this method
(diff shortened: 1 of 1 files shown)


XiaosongWei:
@lincsoon @axinging @fujunwei @wuhengzhi @xzhan96 Please take a look, thanks!

axinging:
LGTM

fujunwei:
LGTM

halton:
Question: My guess is the `disableReflectMethod = true` will remove this API from embedding layer, so what is the impact of user?

XiaosongWei:
@halton ```disableReflectMethod = true``` ensure that the API won't call back to internal layer and be implemented with specific impl. The change here is to make ```XWalkView.setLayerType()``` return directly since only ```LAYER_TYPE_HARDWARE``` is allowed and it's set by default.

halton:
@XiaosongWei so we should mark this API as deprecated somehow? And remove it someday, correct?

XiaosongWei:
@halton no. it's Android View API, we can't remove it. Just need to make it noop.

halton:
Thanks for explanation, lgtm and merge.

PR #3424, opened 2015-11-25
[Windows] Fix detection of available presentation displays

pozdnyakov:
This patch does the following:
1) Immediately notifies screen listener about available displays
2) Changes the behaviour so that available displays are considered to be
the displays that may be used for starting *new* presentations (as per
http://www.w3.org/TR/presentation-api/#dfn-monitor-the-list-of-available-presentation-displays)


pozdnyakov:
@rakuco @darktears PTAL

rakuco:
rs=me

PR #3425, opened 2015-11-25
[Android] Fix crash of incompatible app

lincsoon:
In reflection layer, for those callbacks from library side, if there's
no corresponding method in app side, the library will throw an exception
then the app crashes. This happens particularly when app's version is
older than library's. By this patch, it will try to invoke default
implementation in library side instead of throwing exception.

Related to XWALK-5796

### DIFF
--- a/tools/reflection_generator/java_method.py
+++ b/tools/reflection_generator/java_method.py
@@ -526,21 +526,35 @@ class Method(object):
 ${METHOD_DECLARE_NAME}.invoke(${PARAMS_PASSING}));
     }
 """)
-    else :
+    elif self._is_abstract:
       template = Template("""\
     public ${RETURN_TYPE} ${NAME}(${PARAMS}) {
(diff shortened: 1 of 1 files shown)


lincsoon:
@axinging @fujunwei @wuhengzhi @xzhan96 @XiaosongWei Please help to review

wuhengzhi:
In comment, ```if there's corresponding method``` -> ```if there's no corresponding method```.

fujunwei:
LGTM

tools/reflection_generator/java_method.py
```diff
@@ -549,7 +563,7 @@ def GenerateBridgeWrapperMethod(self):
       if typed_param.generic_type != 'ValueCallback':
         continue
       if typed_param.contains_internal_class:
-        generic_type_declare += 'final %s %sFinal = %s;\n        ' % (
+        generic_type_declare += 'final %s %sFinal = %s;\n            ' % (
```
xzhan96: is this change a typo?

lincsoon: No, just set more indent

PR #3426, opened 2015-11-25
[Android] Fix the mismatch issue.

wuhengzhi:
In newer crosswalk, the parameter nums for loadXWalkLibrary() is
increased, this leads to older crosswalk can not invoke the right
function to load library in shared mode.
Keep the single parameter function to preserve backward compatibility.

Related to XWALK-5796

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
@@ -112,6 +112,11 @@ class XWalkViewDelegate {
         }
     }
 
+    // Keep this function to preserve backward compatibility.
+    public static boolean loadXWalkLibrary(Context context) {
+        return loadXWalkLibrary(context, null);
+    }
+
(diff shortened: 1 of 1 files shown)


wuhengzhi:
@axinging @fujunwei @wuhengzhi @xzhan96 @XiaosongWei PTAL, thanks.

darktears:
This bug leads to major issues on the store. We should have a test.

fujunwei:
Maybe we need a approach to make API compatibility

axinging:
LGTM
@fujunwei  Agree. We should run compatibility test.

xzhan96:
LGTM
For "compatibility test", I think when upgrade a API, please keep the old API and also keep the old test cases instead of changing them.

lincsoon:
@wuhengzhi Please also cherry pick this patch to Crosswalk-16

darktears:
And please add tests that are run part of the test suite. Thanks.

wuhengzhi:
@darktears Yes, I am considering how to add test case. Thanks.

PR #3427, opened 2015-11-25
[Android] Change the library load logic for embedded application.

wuhengzhi:
For embedded application, it loads the library from "app-lib" at first,
if that failed, then loads library from "/data/data/". This will also be
failed when the architecture is inconsistent and lead to crash.
So architecture should be checked before library loaded from "/data/data/".

BUG=XWALK-5804


wuhengzhi:
@axinging @fujunwei @wuhengzhi @xzhan96 @XiaosongWei PTAL, thanks.

wuhengzhi:
This PR should be back ported to Crosswalk-16.

darktears:
Can you rephrase the commit there is some mistakes in english. Also it would be nice to add a test.

xzhan96:
yeah, I also think it better to rephrase the commit log.

wuhengzhi:
Please don't merge this PR.

PR #3428, opened 2015-11-25
crosswalk-16: Backport commits creating POM files in Crosswalk.

rakuco:
Like #3397, this patch set contains two commits. The first one adds all
the required plumbing for generating the Maven POM files we have for
`xwalk_core_library` and `xwalk_shared_library` as part of our build.
The second commit leverages this infrastructure to add a dependency on
the support-v4 Android support library to the xwalk_core_library
artifact, as we have stopped bundling external JARs since #3350.

Related to: XWALK-5092


pozdnyakov:
rs=me

kenchris:
rs=me

PR #3431, opened 2015-11-26
[Android] Fix crash of incompatible app

lincsoon:
In reflection layer, for those callbacks from library side, if there's
no corresponding method in app side, the library will throw an exception
then the app crashes. This happens particularly when app's version is
older than library's. By this patch, it will try to invoke default
implementation in library side instead of throwing exception.

BUG=XWALK-5796
(cherry picked from commit 557f14842ed792746ce1b32daed926519a25d771)


lincsoon:
@rakuco Please help to merge

baleboy:
before merging to the stable branch I would like to see the test results on beta/16

lincsoon:
I see!

PR #3434, opened 2015-11-26
[Android] Ignore the second loading when library was loaded by houdini.

wuhengzhi:
For embedded arm application was installed on x86 device, the library
under folder "/data/app-lib" will be loaded successfully by houdini, for
stability reason, we don't use houdini now, so the variable
"sLibraryLoaded" was not set to "true", then try to load library from
specified url by System.load(). Load library twice will lead to crash
in houdini.
If library was loaded successfully by houdini then ignore the second
loading.

BUG=XWALK-5804


wuhengzhi:
@axinging @fujunwei @lincsoon @xzhan96 @XiaosongWei PTAL, thanks.

XiaosongWei:
LGTM

xzhan96:
LGTM

PR #3387, opened 2015-11-12
[Android] Enable runtime auto-update in download mode

XiaosongWei:
Crosswalk Runtime auto-update is enabled by default in download mode, it requires that the build version of xwalk_shared_library which developer used to bundle with the application has to match the build version of runtime APK, otherwise it will trigger an update to download a new crosswalk runtime from the server.

BUG=XWALK-5681

PR #3440, opened 2015-11-30
[Android] Add build version to XWalkAppVersion/XWalkCoreVersion

XiaosongWei:
Add build version string to both XWalkAppVersion and XWalkCoreVersion.
This would be used to do auto update detection in download mode.

Related to XWALK-5681


XiaosongWei:
@halton

halton:
Since this commit has get lgtm during review on PR #3387 , I'd like to merge this right after trybot green.

PR #3443, opened 2015-12-02
DEPS.xwalk: Re-roll chromium-crosswalk.

rakuco:
No new commits, but a few backports had to have their commit messages
edited. The original crosswalk-16/45.0.2454.101 branch in
chromium-crosswalk has been preserved as
crosswalk-16/45.0.2454.101-with-wrong-cr-commit-position-markers.

After the Blink repository was merged into the Chromium one upstream,
`src/build/util/lastchange.py` looks for the latest upstream Blink commit
that we use in the devtools URL by going through the commit messages and
finding the first with the "Cr-Commit-Position" marker.

This heuristic works most of the time (100% of the time for upstream),
but it breaks for us if we have backported commits with that marker in
the commit message, as the script will find one of those commits that do
not actually exist upstream. Consequently, remote debugging breaks
because the devtools URL we use does not exist.

Crosswalk 17 is fine, because none of our commits on top of Chromium
contain the "Cr-Commit-Position" marker.

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = '45111a77fed0c90c45a1dcc23fdaa0a1c7def478'
+chromium_crosswalk_rev = '9f33a4e0eb2096ee27b6b4407874e2feaee86e2a'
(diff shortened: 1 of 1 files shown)


rakuco:
@darktears @baleboy et al
Remote debugging has been broken on Crosswalk 16 for 2 weeks and nobody noticed :-(

darktears:
I keep saying that these tests QA runs should be part of the tree but.....

rakuco:
For completeness: this is XWALK-5817.

PR #3444, opened 2015-12-03
[Android] IPC message should be setup when subframe is created

xzhan96:
Currently, intercepting http requests with shouldInterceptLoadRequest
is unreliable in combination with iframes, the root cause is that no
ipc message sent from renderer to browser process when subframe is
created, then the XWalkContentsIoThreadClient will not be created.

BUG=XWALK-5621

### DIFF
--- a/runtime/browser/xwalk_render_message_filter.cc
+++ b/runtime/browser/xwalk_render_message_filter.cc
@@ -4,6 +4,11 @@
 
 #include "xwalk/runtime/browser/xwalk_render_message_filter.h"
 
+#if defined(OS_ANDROID)
+#include "xwalk/runtime/browser/android/xwalk_contents_io_thread_client.h"
+#include "xwalk/runtime/common/android/xwalk_render_view_messages.h"
+#endif
+
 #include "xwalk/runtime/common/xwalk_common_messages.h"
 #include "xwalk/runtime/browser/runtime_platform_util.h"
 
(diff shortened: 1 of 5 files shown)


xzhan96:
@lincsoon @axinging @XiaosongWei @fujunwei @wuhengzhi 
Please help to review, thanks.

fujunwei:
LGTM

xzhan96:
@lincsoon @axinging @XiaosongWei  @wuhengzhi 
Please help to review, thanks.

wuhengzhi:
I think we should lift test cases from https://github.com/crosswalk-project/chromium-crosswalk/blob/master/android_webview/javatests/src/org/chromium/android_webview/test/AwContentsClientShouldInterceptRequestTest.java

xzhan96:
@wuhengzhi Current codebase already have the same test case in ShouldInterceptLoadRequestTest.java

runtime/browser/xwalk_render_message_filter.cc
```diff
@@ -13,11 +18,21 @@ XWalkRenderMessageFilter::XWalkRenderMessageFilter()
     : BrowserMessageFilter(ViewMsgStart) {
 }
 
+#if defined(OS_ANDROID)
+XWalkRenderMessageFilter::XWalkRenderMessageFilter(int process_id)
+    : BrowserMessageFilter(AndroidWebViewMsgStart),
```
axinging: Use this name can avoid change of ipc. So, seems ok.

runtime/renderer/xwalk_content_renderer_client.cc
```diff
@@ -31,6 +32,7 @@
 
 #if defined(OS_ANDROID)
 #include "components/cdm/renderer/android_key_systems.h"
+#include "xwalk/runtime/common/android/xwalk_render_view_messages.h"
```
axinging: Better put this below #include "xwalk/runtime/browser/android/net/url_constants.h"

runtime/browser/xwalk_render_message_filter.cc
xzhan96: so, what's the point?

runtime/renderer/xwalk_content_renderer_client.cc
xzhan96: right, thank

axinging:
LGTM

halton:
@pozdnyakov  could you review since you're the OWNER of C/C++ part.

wuhengzhi:
LGTM, thanks.

xzhan96:
@pozdnyakov could you help review? thanks a lot.

xzhan96:
@pozdnyakov could you help to review? thanks.

pozdnyakov:
rs=me as a fixer for XWALK-5621
however I wonder is it Android-specific problem and couldn't it be converted to a cross-platform solution?

xzhan96:
@pozdnyakov 
Per my understanding, it's should be a Android specific problem, chrome and contentshell don't have such event "SubFrameCreated", while Android WebView has.

xzhan96:
@halton can this patch to merge?

xzhan96:
@pozdnyakov could you explain what's mean of "rs=me"?
And do you think this patch can be merged?

halton:
I am not right owner to do the merging, @pozdnyakov，could u do that?

xzhan96:
@pozdnyakov could you help to do that?

PR #3447, opened 2015-12-08
[Android] Ignore the second loading when library was loaded by houdini.

wuhengzhi:
For embedded arm application was installed on x86 device, the library
under folder "/data/app-lib" will be loaded successfully by houdini, for
stability reason, we don't use houdini now, so the variable
"sLibraryLoaded" was not set to "true", then try to load library from
specified url by System.load(). Load library twice will lead to crash
in houdini.
If library was loaded successfully by houdini then ignore the second
loading.

BUG=XWALK-5804

(cherry picked from commit f366b1a9ebddcc6716b03cebbd4edfdd4b8cd97f)


wuhengzhi:
@rakuco @lincsoon Please help to review and merge, thanks.

wuhengzhi:
Should it be backported to crosswalk-15? Crosswalk-16 will be stable soon? Thanks.

baleboy:
16 is enough, we should promote it soon

PR #3448, opened 2015-12-08
[Android] Fix error of missing class in the app package

lincsoon:
Don't throw an exception when the object of new added class is created
in the library package but the app package lacks corresponding class
because of old version.

Relate to XWALK-5796
(cherry picked from commit 1a90ce31e27823bfba542204b3835b5653e5d28f)


lincsoon:
@rakuco Please help to merge

PR #3449, opened 2015-12-08
[Linux] Add JavaScript modal dialog

XiaosongWei:
Leverage components app_modal's JavaScript modal dialog implementation.

BUG=XWALK-5819

(cherry picked from commit c1dd999fce5ed1982431bc48bb0bee9394f14060)

### DIFF
--- a/runtime/browser/runtime.cc
+++ b/runtime/browser/runtime.cc
@@ -10,6 +10,9 @@
 #include "base/command_line.h"
 #include "base/message_loop/message_loop.h"
 #include "base/strings/utf_string_conversions.h"
+#if defined(OS_LINUX)
+#include "components/app_modal/javascript_dialog_manager.h"
+#endif
 #include "content/public/browser/navigation_entry.h"
 #include "content/public/browser/notification_details.h"
 #include "content/public/browser/notification_service.h"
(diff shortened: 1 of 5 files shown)


XiaosongWei:
@rakuco

rakuco:
This looks more like a new feature implementation than a bug fix. Do we really need it in 17?

darktears:
Nope per policy, no new features in beta branches.

XiaosongWei:
It's a critical bug reported by Deepin, if it can't be merged to crosswalk-17 then crosswalk-17 becomes useless for them and they have to start from crosswalk-18.

zhiqiangyu:
Well,  this is a critial bug fix for Deepin. Please prioritize to merge it if possible..
copy @baleboy ..

baleboy:
Is 17 the last version DeepIn will ever use?

zhiqiangyu:
the situation is that deepin is going to release a new version soon,  and during the app test  this major issue was reported by deepin.  so if we can solve this issue in crosswalk 17, it will be the best, otherwise the schedule might be postponded

zhiqiangyu:
and we will further double check w/ Deepin on the release schedule .

darktears:
Sure. We understand their pressure but they also need to understand our release model. It's a 2 way relationship. Plus their schedule slipped a lot the past months and we kind of ramped down because of that. Also they just figured out now right before their release that the JS dialogs are not working?

How we can merge a feature in the beta branch (which is against our policy) without even a QA report on master?

For the record I'm still debugging this patch on Windows. I simply ported the patch to Windows (just removing the ifdef OS_LINUX) because most of the code is cross-platform aka Aura + Toolkit Views and so far it asserts in Debug. I haven't had time to double check that on Linux but based on the fact that the assert is in non windows specific code right in the middle of toolkit views I strongly suspect it asserts as well on Linux. I will verify that first thing tomorrow morning.

baleboy:
Is DeepIn using the beta branch? We are not building nor testing it for Linux

darktears:
And the code asserts in debug on Linux as well. As-is anyway it can't be backported.

XiaosongWei:
Just double confirm with Deepin, they agree to use the latest canary build. So I will close this PR.
Sorry for any inconvenience.

PR #3451, opened 2015-12-09
[Lite] App crush when use --disable-lzma flag.

heke123:
Flag "--disable-lzma" in make_apk.py cause crush.

### DIFF
--- a/app/tools/android/make_apk.py
+++ b/app/tools/android/make_apk.py
@@ -320,10 +320,6 @@ def CopyCompressedLibrary(native_path, library_path, raw_path, arch):
 
 
 def CopyNativeLibrary(native_path, library_path, raw_path, arch):
-  dummy_library = os.path.join(native_path, DUMMY_LIBRARY);
-  if os.path.isfile(dummy_library):
-    os.remove(dummy_library)
-
   shutil.copytree(native_path, os.path.join(library_path, arch))
 
 


heke123:
Found this error when rebasing.  So fix in 15 because it is critical.  I'll port it into lite-17 after it is merged.

heke123:
@PeterWangIntel  @halton  thanks.

app/tools/android/make_apk.py
```diff
@@ -320,10 +320,6 @@ def CopyCompressedLibrary(native_path, library_path, raw_path, arch):
 
 
 def CopyNativeLibrary(native_path, library_path, raw_path, arch):
-  dummy_library = os.path.join(native_path, DUMMY_LIBRARY);
-  if os.path.isfile(dummy_library):
-    os.remove(dummy_library)
```
halton: @heke123 the dummy library is to ensure correct architecture information when user apk upload to Goggle Play. Please ensure your change does not break that case.

heke123: By default lzma is enabled. if user decides to disable lzma, the function "CopyNativeLibrary()" is called, and it mistakenly delete the "libxwalkdummy.so" by running "os.remove(dummy_library)". That's why I deleted those three lines.
After modification, it will be same as it is in master. It is mistakenly introduced when implementing "--disable-lzma" which flag only exists in Lite.

halton:
lgtm

PR #3452, opened 2015-12-09
[Windows] Move JavaScript dialog support to app_modal component.

darktears:
Remove previous code from b19a154 and use the code landed for Linux.

Please note that this commit also fix the ASSERT triggered in debug
build (which is also triggered on Linux). The fix is to make sure
that we initialize the vm state as we're adding transient windows
alongside with the root window.


darktears:
@pozdnyakov @joone @XiaosongWei PTAL

runtime/browser/runtime.cc
```diff
@@ -225,13 +222,10 @@ void Runtime::DidNavigateMainFramePostCommit(
 
 content::JavaScriptDialogManager* Runtime::GetJavaScriptDialogManager(
     content::WebContents* web_contents) {
-#if defined(OS_LINUX)
+#if defined(USE_AURA)
```
XiaosongWei: The guard defined(USE_AURA) is a little bit confused, actually app_modal::JavaScriptDialogManager is platform independent. It also applies to Android.

runtime/browser/runtime.cc
```diff
@@ -225,13 +222,10 @@ void Runtime::DidNavigateMainFramePostCommit(
...
   return app_modal::JavaScriptDialogManager::GetInstance();
 #else
-  if (!javascript_dialog_manager_) {
-    javascript_dialog_manager_.reset(new RuntimeJavaScriptDialogManager);
-  }
-  return javascript_dialog_manager_.get();
```
XiaosongWei: We can't remove RuntimeJavaScriptDialogManager as it's used in Android. Probably we may need to move Android's JS dialog implementation to based on app_modal.

runtime/browser/runtime.h
```diff
@@ -178,8 +178,6 @@ class Runtime : public content::WebContentsDelegate,
   // Notification manager.
   content::NotificationRegistrar registrar_;
 
-  scoped_ptr<content::JavaScriptDialogManager> javascript_dialog_manager_;
-
```
XiaosongWei: it's needed for Android

runtime/browser/runtime.cc
darktears: Right but in xwalk Android doesn't use it yet.

runtime/browser/runtime.cc
darktears: Yes I've uploaded a new version since that change. Not removing it.

darktears: Plus if you carefully read the message and open the original https://github.com/crosswalk-project/crosswalk/commit/b19a154ecbb1bd422b0489cac33c7aace9aae691 you'll see that I've added this code. Android is not using this method but is using the RuntimeJavaScriptDialogManager.

runtime/browser/runtime.h
darktears: Not here. It's created in XWalkWebContentsDelegate::GetJavaScriptDialogManager

darktears: And yes we should probably unify this, I'll do that in a following patch.

darktears:
@pozdnyakov

pozdnyakov:
lgtm

PR #3453, opened 2015-12-09
[Linux][Windows] Fix crash when triggering the inspector.

darktears:
This is due to https://codereview.chromium.org/1123783002 which
make javascript execution more strict and add various runtime checks.
There is a dedicated method that we should use for URLs such as the one
used by the inspector.

BUG=XWALK-5953


darktears:
@pozdnyakov R?

pozdnyakov:
lgtm

PR #3454, opened 2015-12-10
[Android] Update new shouldInterceptRequest usage sample

xzhan96:
As new shouldInterceptRequest has been implemented, the usage sample
also need to be updated.

BUG=XWALK-5956

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -97,6 +97,8 @@ import org.xwalk.core.internal.extension.BuiltinXWalkExtensions;
  *   import org.xwalk.core.internal.XWalkResourceClientInternal;
  *   import org.xwalk.core.internal.XWalkUIClientInternal;
  *   import org.xwalk.core.internal.XWalkViewInternal;
+ *   import org.xwalk.core.internal.XWalkWebResourceRequestInternal;
+ *   import org.xwalk.core.internal.XWalkWebResourceResponseInternal;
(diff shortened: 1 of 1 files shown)


xzhan96:
@fujunwei @wuhengzhi @lincsoon @XiaosongWei @axinging 
please help review.

wuhengzhi:
LGTM, thanks.

xzhan96:
@fujunwei @lincsoon @XiaosongWei @axinging 
Please help to review

xzhan96:
As XWalkWebResourceResponseInternal will also used by xwalk core, so it needs a tag "createInternally = true", which means reflection script would not reflect its parameters, so the constructor of XWalkWebResourceResponse.java is XWalkWebResourceResponse(Object bridge), so the application can't use it.

xzhan96:
@fujunwei  @lincsoon  @axinging please help to review, thanks

fujunwei:
The expose API and call by native should  be independent of each other in the XWalkWebResourceResponseInternal.java ?

xzhan96:
sure, I already separate the expose API and call by native function in XWalkWebResourceResponseInternal.java

fujunwei:
Do you mean we can use `new XWalkWebResourceResponse` after your reorganization?

xzhan96:
No, they are the different thing.
I see what's you mean, you want 1 more file to just reflect API to user, so that application can use "new XWalkWebResourceResponse", right? 
I tried this way before, but not successful, I forget the root cause.
I think I can do it again to get this purpose when I have time, this API update is just for current implementation.

xzhan96:
@lincsoon could you help to review?

PR #3460, opened 2015-12-17
[Android] Fix javadoc build after 5942d33.

rakuco:
Commit 5942d33 ("[Android] Notify the host application that an HTTP
response has been received from the server") introduced invalid javadoc
markup. Build output with OpenJDK8:

```
../out/Release/gen/xwalk_core_reflection_layer/wrapper/org/xwalk/core/XWalkResourceClient.java:331: error: bad use of '>'
     * The feature the same as Android WebView's onReceivedHttpError when HTTP errors have status codes >= 400
```

While here, improve the wording in the comments and, in particular, stop
saying "customers" when we just mean users. Not all users are customers.


r0b5t4:
rs+

rakuco:
@fujunwei: FYI ^

PR #3462, opened 2015-12-18
[Android] Add test APKs for shared mode

lincsoon:
Add XWalkSharedShell and XWalkSharedTest for shared mode test, the
package name, test cases, assets of these APKs are just like
XWalkCoreShell and XWalkCoreTest except that they must work with
XWalkRuntimeLib.

### DIFF
--- a/xwalk_android_tests.gypi
+++ b/xwalk_android_tests.gypi
@@ -35,6 +35,7 @@
         'xwalk_core_java',
         'xwalk_core_shell_apk_pak',
         'libxwalkdummy',
+        'xwalk_xwview_assets',
       ],
       'variables': {
         'apk_name': 'XWalkCoreShell',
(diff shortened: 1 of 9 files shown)


rakuco:
Please remember to file JIRA tickets for those APKs to start being shipped with each build, as well as added to the bot test list.

lincsoon:
@rakuco These new APKs don't need to be shipped nor to be included into the bot because they are the exactly same with XWalkCoreShell and XWalkCoreTest. They are only for compatibility test of shared mode. Before every beta release, QA will run XWalkSharedTest on different versions of XWalkRuntimeLib manually.

lincsoon:
@rakuco Could you give a LGTM so that I can merge this.

xwalk_android_tests.gypi
```diff
@@ -138,6 +182,7 @@
         'xwalk_core_shell_apk_java',
         '../tools/android/md5sum/md5sum.gyp:md5sum',
         '../tools/android/forwarder2/forwarder.gyp:forwarder2',
+        'xwalk_xwview_test_assets',
```
wuhengzhi: Reason for adding this?

lincsoon:
@rakuco

rakuco:
lgtm
I still think you need to file a ticket to get this into the bot/release build workflow, otherwise we won't even know if it's minimally running, and if it's not part of the zip files QA will need to build the APKs on their own.

lincsoon:
Thanks for the reply. I'll file a ticket to ship this.

PR #3463, opened 2015-12-21
[Android] Use regexp to match url when load.

axinging:
Use regexp to guarantee the bellow urls match:
http://www.abc.com==http://www.abc.com/
http://www.abc.cn==http://www.abc.cn/

BUG=XWALK-6027

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -11,6 +11,7 @@ import android.content.Intent;
 import android.content.SharedPreferences;
 import android.graphics.Bitmap;
 import android.graphics.Rect;
+import android.net.Uri;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Bundle;
@@ -33,6 +34,7 @@ import android.widget.FrameLayout;
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.annotation.Annotation;
+import java.util.List;
 import java.util.Locale;
 
 import org.chromium.base.annotations.CalledByNative;
@@ -262,10 +264,32 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
         mContentViewCore.onShow();
     }
 
+    private static boolean urlMatch(String url1, String url2) {
+        if (url1 == null || url2 == null) return url1 == url2;
+
+        Uri uri1 = Uri.parse(url1);
+        Uri uri2 = Uri.parse(url2);
+        if (!TextUtils.equals(uri1.getScheme(), uri2.getScheme())
+                || !TextUtils.equals(uri1.getHost(), uri2.getHost())) {
+            return false;
+        }
+
+        List<String> pathSegments1 = uri1.getPathSegments();
+        List<String> pathSegments2 = uri2.getPathSegments();
+        int pathSize1 = pathSegments1.size();
+        int pathSize2 = pathSegments2.size();
+        if (pathSize1 != pathSize2) return false;
+        for (int i = 0; i < pathSize1; i++) {
+            if (!TextUtils.equals(pathSegments1.get(i), pathSegments2.get(i))) return false;
+        }
+        return true;
+    }
+
     void doLoadUrl(String url, String content) {
         // Handle the same url loading by parameters.
-        if (url != null && !url.isEmpty() &&
-                TextUtils.equals(url, mWebContents.getUrl())) {
+        String webContentsUrl = mWebContents.getUrl();
+        if (!TextUtils.isEmpty(url) && !TextUtils.isEmpty(webContentsUrl)
+                && urlMatch(url.toLowerCase(), webContentsUrl.toLowerCase())) {
             mNavigationController.reload(true);
         } else {
             LoadUrlParams params = null;


axinging:
@lincsoon  @fujunwei  @XiaosongWei @wuhengzhi

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -87,6 +87,16 @@
...
 
+    /**
+    * Use URL_PATTERN to guarantee the bellow urls match:
+    * www.abc.com==http://www.abc.com/
+    * www.abc.com==http://www.abc.com
+    * www.abc.com/==http://www.abc.com/
+    * ////www.abc.com==http://www.abc.com/
+    * http://www.abc.com==http://www.abc.com/
+    */
+    private static final String URL_PATTERN="(^(http://)?)(/*)|((/$)?)";
```
wuhengzhi: need space before and after ```=```.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -265,7 +275,8 @@ private void receivePopupContents(long popupNativeXWalkContents) {
     void doLoadUrl(String url, String content) {
         // Handle the same url loading by parameters.
         if (url != null && !url.isEmpty() &&
-                TextUtils.equals(url, mWebContents.getUrl())) {
+                TextUtils.equals(url.replaceAll(URL_PATTERN,""),
+                mWebContents.getUrl().replaceAll(URL_PATTERN,""))) {
```
wuhengzhi: need space before and after ```,```.

wuhengzhi:
And we need to add test case for this change.

rakuco:
> BUG=XWALK-6027

The mentioned bug just says "Use regexp to match url when load" and points to XWALK-5065. Looking at it I have no idea why you want to do that and what is currently broken. If I go to XWALK-5065, there are zero references to XWALK-6027, so I am still left wondering what is going on.

axinging:
@rakuco, thanks for this comments.
If you check the comments of XWALK-5065, you can see I add the following workaround in customer's code:
view.load("http://www.huskers.com/", null); //Note: Add slash at the end of URL.

The original code or customer is:
view.load("http://www.huskers.com", null);

In deed, the code of customer is nothing wrong.  But in XWalkContent.doLoadUrl,  it will compare the url of load(http://www.huskers.com) with mWebContents.getUrl().  However, mWebContents.getUrl() will fulfill this url with slash(http://www.huskers.com/). So url of load and  mWebContents.getUrl() don't equal even if they are the same page, they have a slash difference.

As to this patch, we can guarantee that  user's http://www.huskers.com match http://www.huskers.com/ from mWebContents.getUrl().

rakuco:
OK, but why do you want to make the code consider http://crosswalk-project.org equal to http://crosswalk-project.org/ in the first place? What's the problem with `mNavigationController.loadUrl()` that's called in the code path you are trying to avoid? In XWALK-5065 you just say this is a "workaround", but what are you trying to work around?

wuhengzhi:
Other scheme will has same issue? e.g ```https://```

axinging:
@wuhengzhi  Yes, I am also thinking about this.

axinging:
@rakuco 
1), About your first place, customer can load http://crosswalk-project.org or http://crosswalk-project.org/ in his code. This is out of our control.
2), Avoid loadUrl,  using reload instead.
As we know, load http://crosswalk-project.org is the same as load http://crosswalk-project.org/. So if user load http://crosswalk-project.org first, then again, this should trigger reload. 

But as mention above,  when load http://crosswalk-project.org again,  mWebContents.getUrl() will be changed to http://crosswalk-project.org/. In the original code, XWalkContent.doLoadUrl will trigger loadUrl instead of reload, this is not what we want, reload is better option here.  So we can not simply use equal here.

rakuco:
I understand that you are trying to change the comparison, my question has always been why: why is reload a better option? Do things _not_ work if you don't reload?

axinging:
@rakuco,  I didn't investigate the root cause, but it seems reload has a much better memory usage than loadurl, please refer to: 5065_memory10s_xiaomi4_no_aboutblank.xlsx at https://crosswalk-project.org/jira/browse/XWALK-5065

rakuco:
Then it sounds like this patch is somewhat blindly fixing a symptom of a bigger problem. Without a proper explanation of why this helps and what exactly is wrong, I don't see a reason to commit this patch to the tree. It would be better to spend some time figuring out if there's a memory leak in the other code path, and checking what Chromium itself does in this case.

axinging:
@rakuco 
Even memory case is not verified. I think it worth to commit this patch because it is not correct to use equal to compare if two urls match.

rakuco:
That's what the `loadUrl()` implementations in Chromium do so far. See `content/shell/android/java/src/org/chromium/content_shell/Shell.java`, `chromecast/browser/android/apk/src/org/chromium/chromecast/shell/CastWindowAndroid.java` and `android_webview/java/src/org/chromium/android_webview/AwContents.java`, for example. In a way, it all boils down to comparing URIs: arguably, "http://crosswalk-project.org" has an empty path while "http://crosswalk-project.org/"'s path is "/". And in any case, hand-rolling your own URI parser with a regexp is almost never a good idea, if you really want to be thorough you should be using `android.net.Uri` and comparing schemas, paths etc. I just don't see the benefit in doing that here.

axinging:
Shell.java is not a good example. This may not be fully tested, if tested with case of 5065, the behavior of Shell.java should be wrong.

Will try this with Uri.

axinging:
@rakuco, use URI instead.
@lincsoon @fujunwei @XiaosongWei @wuhengzhi

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -262,10 +264,31 @@ private void receivePopupContents(long popupNativeXWalkContents) {
...
+        }
+        return true;
+    }
+
     void doLoadUrl(String url, String content) {
         // Handle the same url loading by parameters.
-        if (url != null && !url.isEmpty() &&
-                TextUtils.equals(url, mWebContents.getUrl())) {
+        if (url != null && !url.isEmpty()
+                && urlMatch(url.toLowerCase(), mWebContents.getUrl().toLowerCase())) {
```
wuhengzhi: If ```mWebContents.getUrl()``` is null, will lead to null pointer exception.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -262,10 +264,31 @@ private void receivePopupContents(long popupNativeXWalkContents) {
...
+
+        Uri uri1 = Uri.parse(url1);
+        Uri uri2 = Uri.parse(url2);
+        if (!TextUtils.equals(uri1.getScheme(), uri2.getScheme())
+                || !TextUtils.equals(uri1.getHost(), uri2.getHost())) {
+            return false;
+        }
+
+        List<String> pathSegments1 = uri1.getPathSegments();
+        List<String> pathSegments2 = uri2.getPathSegments();
```
wuhengzhi: Could ```Uri.equals()``` be used to compare two objects directly?

wuhengzhi: Oh, I know ```Uri.equals()``` can't be used directly since one of the url maybe contain ```/```.

axinging: Yes.

PR #3465, opened 2015-12-24
[Android] Refine the error message for mismatched architecture.

wuhengzhi:
If the library was packaged into wrong architecture folder, e.g. library
for x86 was put in "armeabi-v7a", it will throw "UnsatisfiedLinkError"
exception when load library, this situation belongs to mismatch status
instead of incomplete library.
Check the error cause and give the right status.

BUG=XWALK-5803

(cherry picked from commit c145065728e67d056444bd58ca3f1843e58e712d)


wuhengzhi:
@lincsoon Please help to merge. Thanks.

PR #3469, opened 2015-12-25
[Android] Fix the issue that video can't play with blob url.

wuhengzhi:
Blob data is stored in Crosswalk cache, the cache path is local address,
e.g "/sdcard/cache/" or other relative urls, its scheme is null.
In Crosswalk, "ResourceLoadingFilter" has been implemented to customize
the resource url, if resource url is cache path, the scheme is null, then
"uri.getScheme().equals()" will lead to null pointer exception and be
caught in "MediaPlayerBridge.java:setDataSource()", data source can't be
set exactly, then report media error format.
Add test case for blob url to make sure the video from indexedDB can be
played successfully.

BUG=XWALK-6004


js0701:
LGTM. Thanks@wuhengzhi

wuhengzhi:
@lincsoon @fujunwei @xzhan96 @XiaosongWei @axinging PTAL, thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkMediaPlayerResourceLoadingFilter.java
```diff
@@ -25,12 +25,12 @@
     @Override
     public boolean shouldOverrideResourceLoading(MediaPlayer mediaPlayer,
             Context context, Uri uri) {
-        if (uri.getScheme().equals(AndroidProtocolHandler.APP_SCHEME)) {
+        String scheme = uri.getScheme();
+        if (scheme == null) return false;
```
halton: This is kind of danger to identify schema null is for the cache uri. For eg: aaa, /root all are negative. Please check the cache case explicitly.

wuhengzhi: Sync with @halton, updated the comment, thanks.

halton:
With offline sync with @wuhengzhi , the updated git log and code LGTM. Will merge when trybot green.

rakuco:
can we have a test for this bug fix?

wuhengzhi:
I will investigate how to add test case.

wuhengzhi:
I have tried to implement the test case by ```TestWebServer```, but failed. Then we can implement it by online urls, e.g https://vidolio.com/crosswalk-bug/, but it maybe deleted in future and the internet access was disabled on test device.

rakuco:
What problems did you encounter with `TestWebServer`? My understanding was that you could put a small video and use IndexedDB to test this. I'd like to land this with a test, otherwise we can't be sure this won't regress in the future.

wuhengzhi:
@rakuco The previous idea is try to get video data from XHR by ```TestWebServer```, then write to indexedDB.
Now with following steps to implement test case:
1. Open database, write video data to indexedDB.
2. Open database, read video data from indexedDB.
3. Verify the play status.
@halton PTAL again.
Thanks.

test/android/core/javatests/src/org/xwalk/core/xwview/test/VideoWithBlobUrlTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+import android.test.suitebuilder.annotation.SmallTest;
+import android.util.Pair;
+
+import org.chromium.base.test.util.Feature;
+import org.chromium.net.test.util.TestWebServer;
+import org.xwalk.core.JavascriptInterface;
+import org.xwalk.core.xwview.test.util.CommonResources;
+
+import java.util.List;
```
halton: nit, please order

test/android/data/read_video_data.html
```diff
@@ -0,0 +1,38 @@
...
+        db.onerror = function (event) {
+            console.log("Error accessing IndexedDB database");
+        };
+        readVideoFromDb();
+    }
+</script>
+<body>
+<h3>Read video data from indedexDB and play.</h3>
+<video id="videoid" src="" controls autoplay></video>
+</body>
```
halton: You need use 2-space indent. Same for other html/js files.

test/android/data/read_video_data.html
```diff
@@ -0,0 +1,38 @@
+<script>
+var dbVersion = 2;
+var indexedDB = window.indexedDB
+var objectStore = "videotest"
+var request = indexedDB.open("videoFiles", dbVersion), db
+
+    readVideoFromDb = function () {
+        var transaction = db.transaction([objectStore], "readwrite");
```
halton: isn't read permission sufficient?

test/android/data/save_video_data.html
```diff
@@ -0,0 +1,51 @@
+<script>
+  var videodata =
+      "GkXfo0AgQoaBAUL3gQFC8oEEQvOBCEKCQAR3ZWJtQoeBAkKFgQIYU4BnQN8VSalmQCgq17FAAw9C" +
+      "QE2AQAZ3aGFtbXlXQUAGd2hhbW15RIlACECPQAAAAAAAFlSua0AxrkAu14EBY8WBAZyBACK1nEAD" +
+      "dW5khkAFVl9WUDglhohAA1ZQOIOBAeBABrCBlrqBlh9DtnVAdOeBAKNAboEAAIDyCACdASqWAJYA" +
+      "Pk0ci0WD+IBAAJiWlu4XdQTSq2H4MW0+sMO0gz8HMRe+0jRo0aNGjRo0aNGjRo0aNGjRo0aNGjRo" +
+      "0aNGjRo0aNGjRo0VAAD+/729RWRzH4mOZ9/O8Dl319afX4gsgAAA";
```
halton: The one last concern is where this video comes from? And what is licence and copyright. We need make sure you're not using any property content.

wuhengzhi: Video data comes from https://github.com/crosswalk-project/crosswalk/blob/master/test/android/core_internal/javatests/src/org/xwalk/core/internal/xwview/test/util/VideoTestWebServer.java#L25

halton: Thanks for pointing me out, turn out it re-uses https://code.google.com/p/chromium/codesearch#chromium/src/android_webview/javatests/src/org/chromium/android_webview/test/util/VideoTestWebServer.java&q=one_pixel_one_frame.webm&sq=package:chromium&type=cs&l=18, could you add comment to indicate?

Then you have my LGTM.

wuhengzhi: Yes, thanks.

halton:
LGTM, will merge after trybot green.

PR #3470, opened 2015-12-25
Allow media access permission check by default

huningxin:
mediaDevices.enumerateDevices() API depends on this check.

BUG=XWALK-6060

### DIFF
--- a/runtime/browser/runtime.cc
+++ b/runtime/browser/runtime.cc
@@ -334,9 +334,11 @@ bool Runtime::CheckMediaAccessPermission(
     content::WebContents* web_contents,
     const GURL& security_origin,
     content::MediaStreamType type) {
-  // TODO(xiang): Pepper flash plugin will trigger this check a lot, return
-  // false at the moment.
-  return false;
+  // Requested by Pepper Flash plugin and mediaDevices.enumerateDevices().
+  // FIXME(huningxin): current code grants the access by default. It should
+  // follow the generic permssions mechanism (XWALK-5475). Open XWALK-6083 to
+  // track it.
+  return true;
 }
 
 void Runtime::LoadProgressChanged(content::WebContents* source,


huningxin:
With this fix, mediaDevices.enumerateDevices() API works correctly.

See screenshot:

image

huningxin:
@darktears @halton , PTAL. Thanks!

halton:
LGTM

darktears:
I'm not sure we want to return a always true here. On Android there is a permission system so the user is aware, on Windows there is one for Universal Apps as well as Chrome. I think we want to put the bar higher than just always granting permissions IMHO.

We have https://crosswalk-project.org/jira/browse/XWALK-5475?filter=-1 to implement. We better start with that.

huningxin:
If I understand correctly, Crosswalk apps should behave like native apps as much as possible. So from the beginning, we grant the geolocation API, getUserMedia API etc., by default without prompting to user. So does this PR, allow the enumerateDevices API, part of getUserMedia API automatically.

As you said, Crosswalk apps, just like the native apps, follow the permission system of the OS, like Android and Windows Universal Apps. So even we return true here, on  Android, if user doesn't grant the camera access for the app, the enumerateDevices API will fail. 

This PR focuses on fixing enumerateDevices API doesn't work (XWALK-6060). I suggest not block it by XWALK-5475. Once XWALK-5475 implemented, enumerateDevices API, with other device APIs (geolocation, getUserMedi etc.,) together, will follow the permission system.

@darktears , WDYT?

darktears:
That's not correct, we do not grant geolocation for example. I'm actually working on fixing this and then working on prompting the user to grant or not access. People running Universal Apps are looking at a permission model similar to iOS where one grant access when it needs (just like Android 6.0+).

I understand you want to make it work, I suggest you add a FIXME, open a bug in JIRA to keep that tracked because it's not an ideal fix.

huningxin:
> I understand you want to make it work, I suggest you add a FIXME, open a bug in JIRA to keep that tracked because it's not an ideal fix.

It sounds good to me. I will do that and let you review soon. Thanks!

huningxin:
Done. FIXME has been added. https://crosswalk-project.org/jira/browse/XWALK-6083 has been opened. 

@darktears , please take another look. Thanks!

darktears:
lgtm

huningxin:
Thanks for review! Please help merge it at your convenience.

halton:
@darktears could you please merge it? I'm not the runtime OWNER.

PR #3471, opened 2015-12-30
M48 gardening: Fix XWalk core shell crashed with proxy service.

wuhengzhi:
From https://codereview.chromium.org/1356933002, in
aw_browser_context.cc, we can see the proxy config service was not reset
to the casted service.
This is the part 1 to fix the core shell crashed.

BUG=XWALK-6046


wuhengzhi:
@mrunalk PTAL, thanks.

mrunalk:
LGTM

PR #3472, opened 2016-01-04
[Android] Update XWalkContentView to track ContentView

lincsoon:
Move the methods for JellyBean to XWalkContentViewApi16 to imitate
JellyBeanContentView.
Move the mothods for AndroidM to XWalkContentViewApi23 to imitate
ContentView.ContentViewApi23.
All methods remaining in XWalkContentView is for XWalkView's usage.

XWalkContentView is a temporary solution. We are targeting to optimize
the hierarchy of XWalkView to make it more like WebView. This task can
be tracked via XWALK-6118.

BUG=XWALK-6014

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = 'f6dc12323e2705833c7b9b3cc25f264efd4714b4'
+chromium_crosswalk_rev = '6576bdb08fdecc552328fd856f5c818279be279d'
 v8_crosswalk_rev = '8e0f20386c9d322eb9fb5738d0dd9e63a3a0b7b2'
 
 crosswalk_git = 'https://github.com/crosswalk-project'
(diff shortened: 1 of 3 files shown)


lincsoon:
This pull request is after https://github.com/crosswalk-project/chromium-crosswalk/pull/304 gets merged, and then roll deps to new commits of chromium-crosswalk.

rakuco:
The code looks OK, but I have a broader question about the need for `XWalkContentView` in the first place. Sorry if it sounds obvious to you, but this gigantic class hierarchy still confuses me.

Upstream code (including `WebViewChromium` in `android_webview`) seems to work fine without having to extend `ContentView`, so I was wondering if we really need to do that. Having it around means spending all this time getting the build fixed and basically repeating a lot of code just to make some calls to `XWalkView`. The Android WebView code seems to work only one way (`WebViewChromium` calls` AwContents` that calls `ContentView` and `ContentViewCore`, the latter do not call the former).

lincsoon:
Embedding `ContentView` in `XWalkView` is for some historical reason. Now it begins to cause some problems like the commit comment said. We're actually trying to refactor `XWalkView` and resolve this matter fundamentally. But before that, `XWalkContentView` is just a temporary solution.

rakuco:
Please at least mention XWALK-6118 in this commit and make it clear in the commit message that this is supposed to be a temporary solution.

lincsoon:
Rolled chromium to latest fix. @rakuco Please help to merge

rakuco:
FYI, I also had to make `XWalkContent` call `onCreateInputConnectionSuper` again, as we'd end up with an infinite loop otherwise.

PR #3474, opened 2016-01-06
[Android] Fix the failed test cases.

wuhengzhi:
1. Fix all of the failed test cases about url override since the throttle
for navigation was not registered. Please refer to
https://codereview.chromium.org/1363483007
2. Modify the compare condition for OnDocumentLoadedInFrame since the
frame id is not invariable. The routing id is increased when the widget
routing id was added, please refer to
https://codereview.chromium.org/1303773002
3. Remove partial test code for SetInitialScale since the scale factor
also be changed since upstream's change.

BUG=XWALK-6046

### DIFF
--- a/test/android/core/javatests/src/org/xwalk/core/xwview/test/OnDocumentLoadedInFrameTest.java
+++ b/test/android/core/javatests/src/org/xwalk/core/xwview/test/OnDocumentLoadedInFrameTest.java
@@ -52,7 +52,7 @@ public class OnDocumentLoadedInFrameTest extends XWalkViewTestBase {
         final String url = addPageToTestServer(mWebServer, path, pageContent);
 
         loadUrlSync(url);
-        assertEquals(1, mOnDocumentLoadedInFrameHelper.getFrameId());
+        assertTrue(mOnDocumentLoadedInFrameHelper.getFrameId() > 0);
         assertEquals(1, mOnDocumentLoadedInFrameHelper.getCallCount());
     }
 }


wuhengzhi:
@mrunalk @axinging @lincsoon PTAL, thanks.

rakuco:
Do you have links for the upstream CLs that caused these changes?

wuhengzhi:
@rakuco For the third one, I have not found the CL. The removed code was not from upstream, which I added before, please look at https://github.com/crosswalk-project/chromium-crosswalk/blob/master/android_webview/javatests/src/org/chromium/android_webview/test/AwSettingsTest.java#L2698, we can't follow the upstream's code since the implementation and callback are different.

rakuco:
Thanks for the patch; I'm using it locally to run the M48 Android tests and the failures I'm facing are not related to this change. Can you please rebase it on top of the current next branch for me to merge it?

rakuco:
I'm looking at the changes again after testing them locally.

The first 2 make sense, but I'm still in doubt with the page scale part: the implementations may be different, but I thought the effects should be the same here. Our scale calculations look different from android_webview, and I do not know if this should be the case. We need to find out what exactly changed between M46 and M48 and determine if removing that part of the test is the best course of action or if it's just hiding a deeper issue.

rakuco:
I've cherry-picked and committed the first 2 changes as you can see. The 3rd part is pending as I mentioned above.

rakuco:
> I've cherry-picked and committed the first 2 changes as you can see. The 3rd part is pending as I mentioned above.

Actually, the navigation throttling part needs work: the code only works on Android. Building on Linux fails:

```
In file included from ../../xwalk/runtime/browser/xwalk_content_browser_client.cc:13:
In file included from ../../components/navigation_interception/intercept_navigation_delegate.h:8:
../../base/android/jni_weak_ref.h:8:10: fatal error: 'jni.h' file not found
#include <jni.h>
         ^
```

rakuco:
So to be clear: right now I've cherry-picked fix number 2, but number 1 and 3 still need some work.

baleboy:
@wuhengzhi are there any other PRs for points 1 and 3 above?

rakuco:
OK, I've re-pushed next and fixed point 1: I just defined and implemented `XWalkContentBrowserClient::CreateThrottlesForNavigation` for Android only (we did Throttle-related stuff in `RuntimeResourceDispatcherHostDelegateAndroid::RequestBeginning`).

Point 3 is still pending.

wuhengzhi:
I will find out the CL for point 3, thanks.

PR #3475, opened 2016-01-06
[Android] Add Cookie header to download requests.

dvicory:
Use any cookies for the URL that is being downloaded from with the default XWalkDownloadListenerImpl handler. This allows files that require Cookie-based authentication to be downloaded.

BUG=XWALK-5981
(cherry picked from commit ce005ec)


rakuco:
lgtm!

PR #3483, opened 2016-01-11
[Android] Fix javadoc build after cc194840.

rakuco:
OpenJDK8 is stricter and produces an error instead of a warning (the
documentation would have looked wrong anyway):

```
../out/Release/gen/xwalk_core_reflection_layer/wrapper/org/xwalk/core/XWalkView.java:111:
error: malformed HTML
 *               //             int statusCode, String reasonPhrase, Map<String, String> responseHeaders)
                                                                        ^
../out/Release/gen/xwalk_core_reflection_layer/wrapper/org/xwalk/core/XWalkView.java:111:
error: bad use of '>'
 *               //             int statusCode, String reasonPhrase, Map<String, String> responseHeaders)
                                                                                       ^
```


rakuco:
@xzhan96: ^

PR #3484, opened 2016-01-11
[Windows][Linux] Fix geolocation support and implement a system to as…

darktears:
…k for

permissions.

XWalk will now ask for permission using a dialog and the answer will
be kept persistent across runs. XWalkPermissionManager is the entity
responsible to create the dialog asking for permissions and it is also
the one responsible to verify if a permission was granted/denied in the
past.

XWalkContentSettings is the class responsible to store whether a
permission was granted or not and store it onto the filesystem.

Currently there is no way to reset the permissions once granted/denied.
We should probably implement a way to do so (like a command line to
reset the permissions).

Concerning the geolocation, right now the only backend we have is using
the Google Maps API and thus one need to follow these instructions to get
a key https://www.chromium.org/developers/how-tos/api-keys. At the moment
you need GOOGLE_API_KEY, GOOGLE_DEFAULT_CLIENT_ID and
GOOGLE_DEFAULT_CLIENT_SECRET to be defined as environment variables. This
isn't going to work for crosswalk-app-tool : we either need these to be
defined in say the manifest.json or it has to be a command line argument
which could be passed by crosswalk-app-tools.

BUG=XWALK-4989, XWALK-5475


darktears:
@pozdnyakov PTAL.

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
```diff
@@ -0,0 +1,89 @@
...
+#include "grit/components_strings.h"
+#include "xwalk/runtime/browser/ui/desktop/xwalk_permission_modal_dialog.h"
+#include "xwalk/runtime/browser/xwalk_content_settings.h"
+
+DEFINE_WEB_CONTENTS_USER_DATA_KEY(xwalk::XWalkPermissionDialogManager);
+
+namespace xwalk {
+
+XWalkPermissionDialogManager::XWalkPermissionDialogManager(
+    content::WebContents* web_contents) {
```
pozdnyakov: why do we need this arg?

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
```diff
@@ -0,0 +1,89 @@
...
+    const GURL& origin_url,
+    const std::string& accept_lang,
+    const base::string16& message_text,
+    const base::Callback<void(bool)>& callback) {
+  ContentSetting content_setting =
+      XWalkContentSettings::GetInstance()->GetPermission(
+          type,
+          origin_url,
+          web_contents->GetLastCommittedURL().GetOrigin());
+  if (content_setting == CONTENT_SETTING_ALLOW ||
```
pozdnyakov: IMO switch would be better:
`switch(content_setting) {
case CONTENT_SETTING_ALLOW:
  callback.Run(true);
  break;
case..
}`

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
```diff
@@ -0,0 +1,89 @@
...
+      content_setting == CONTENT_SETTING_BLOCK) {
+    callback.Run(content_setting == CONTENT_SETTING_ALLOW ? true : false);
```
pozdnyakov: `content_setting == CONTENT_SETTING_ALLOW` returns `bool` already, `? true : false` is not needed.

pozdnyakov: shouldn't `callback` be checked?

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
```diff
@@ -0,0 +1,89 @@
...
+    return;
+  }
+
+  app_modal::AppModalDialogQueue::GetInstance()->AddDialog(
+      new XWalkPermissionModalDialog(
+          web_contents,
+          message_text,
+          base::Bind(&XWalkPermissionDialogManager::OnPermissionDialogClosed,
+          base::Unretained(this), web_contents, type, origin_url, callback)));
```
pozdnyakov: is it guarantied that `this` lives longer than the dialog?

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
```diff
@@ -0,0 +1,89 @@
...
+}
+
+void XWalkPermissionDialogManager::CancelPermissionRequest(
+    content::WebContents* web_contents) {
+  app_modal::AppModalDialogQueue* queue =
+      app_modal::AppModalDialogQueue::GetInstance();
+  app_modal::AppModalDialog* active_dialog = queue->active_dialog();
+  for (app_modal::AppModalDialogQueue::iterator i = queue->begin();
```
pozdnyakov: `for (auto it = `

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
```diff
@@ -0,0 +1,89 @@
...
+    const base::Callback<void(bool)>& callback,
+    bool success) {
+  ContentSetting content_setting =
+      success ? CONTENT_SETTING_ALLOW : CONTENT_SETTING_BLOCK;
+  XWalkContentSettings::GetInstance()->SetPermission(
+      type,
+      origin_url,
+      web_contents->GetLastCommittedURL().GetOrigin(),
+      content_setting);
+  callback.Run(success);
```
pozdnyakov: check callback?

runtime/browser/ui/desktop/xwalk_permission_modal_dialog_views.cc
```diff
@@ -0,0 +1,132 @@
...
+namespace xwalk {
+
+XWalkPermissionModalDialogViews::XWalkPermissionModalDialogViews(
+    XWalkPermissionModalDialog* parent)
+  : parent_(parent) {
+  int options = views::MessageBoxView::DETECT_DIRECTIONALITY;
+
+  views::MessageBoxView::InitParams params(parent->message_text());
+  params.options = options;
+  message_box_view_ = new views::MessageBoxView(params);
```
pozdnyakov: how is it freed?

runtime/browser/ui/desktop/xwalk_permission_modal_dialog_views.cc
```diff
@@ -0,0 +1,132 @@
...
+  DCHECK(message_box_view_);
```
pozdnyakov: how `message_box_view_` can be null?

runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc
darktears: This will be destroyed when the WebContent is about to die. But you're right we should invalidate the dialog when this dies so that the callback will never be called.

darktears:
@pozdnyakov PTAL

runtime/browser/runtime_geolocation_permission_context.cc
```diff
@@ -82,6 +101,47 @@ RuntimeGeolocationPermissionContext::RequestGeolocationPermission(
...
+      requesting_frame, pref_service->GetString(kIntlAcceptLanguage),
+      url_formatter::kFormatUrlOmitUsernamePassword |
+      url_formatter::kFormatUrlOmitTrailingSlashOnBareHostname,
+      net::UnescapeRule::SPACES, nullptr, nullptr, nullptr));
+
+  permission_dialog_manager->RequestPermission(
+      CONTENT_SETTINGS_TYPE_GEOLOCATION,
+      requesting_frame, "", text,
+      base::Bind(
+      &RuntimeGeolocationPermissionContext::OnPermissionRequestFinished,
```
pozdnyakov: nit:ident

runtime/browser/runtime_geolocation_permission_context.cc
```diff
@@ -82,6 +101,47 @@ RuntimeGeolocationPermissionContext::RequestGeolocationPermission(
...
+      base::Unretained(this), result_callback));
```
pozdnyakov: same comment, is it guaranteed that `this` lives longer than `permission_dialog_manager`?

runtime/browser/runtime_geolocation_permission_context.cc
```diff
@@ -82,6 +101,47 @@ RuntimeGeolocationPermissionContext::RequestGeolocationPermission(
...
+#endif
+}
+
+#if !defined(OS_ANDROID)
+void RuntimeGeolocationPermissionContext::OnPermissionRequestFinished(
+    base::Callback<void(bool)> result_callback, bool success) {
+  if (success) {
+    content::GeolocationProvider::GetInstance()
+        ->UserDidOptIntoLocationServices();
+    result_callback.Run(true);
```
pozdnyakov: can `result_callback` be null?

runtime/browser/ui/desktop/xwalk_permission_modal_dialog_views.cc
```diff
@@ -0,0 +1,131 @@
...
+
+  views::MessageBoxView::InitParams params(parent->message_text());
+  params.options = options;
+  message_box_view_ = new views::MessageBoxView(params);
+  message_box_view_->AddAccelerator(
+      ui::Accelerator(ui::VKEY_C, ui::EF_CONTROL_DOWN));
+}
+
+XWalkPermissionModalDialogViews::~XWalkPermissionModalDialogViews() {
+  delete message_box_view_;
```
pozdnyakov: scoped_ptr?

runtime/browser/runtime_geolocation_permission_context.cc
darktears: No, it's passed around and it's checked above before.

runtime/browser/runtime_geolocation_permission_context.cc
darktears: Not sure but it doesn't matter because both are linked to web_contents, if the later gets destroyed, permission_dialog_manager is going to shutdown and the callback will never be called if that's your concern.

runtime/browser/ui/desktop/xwalk_permission_modal_dialog_views.cc
darktears: Fixed indeed.

pozdnyakov:
lgtm

PR #3486, opened 2016-01-12
crosswalk-17: Fix some memory leaks in the widget handling code.

rakuco:
They only manifest themselves in case ParsePreferenceItem() fails: in
this case, DictionaryValue::Set() is not called and |preference| leaks.

Fix it by turning most raw pointers in WidgetHandler::Parse() into
scoped_ptr's, even where it is not strictly necessary.

WidgetInfo::Set()'s prototype had to be adjusted a little to accept and
forward scoped_ptr's instead of base::Value*'s, but it is a change for
the good that prevents similar mistakes in the future.

CID=323513
CID=323514

(cherry picked from commit 56450318523a98f63e9c0aeccce5110443c9b73d)

### DIFF
--- a/application/common/manifest_handlers/widget_handler.cc
+++ b/application/common/manifest_handlers/widget_handler.cc
@@ -99,8 +99,8 @@ void WidgetInfo::SetString(const std::string& key, const std::string& value) {
   value_->SetString(key, value);
 }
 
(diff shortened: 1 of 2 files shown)


pozdnyakov:
lgtm

PR #3487, opened 2016-01-12
[Android] Remove make_apk and associated tests from the tree.

rakuco:
An "intent to deprecate" make_apk was sent in November 2015, and it is
finally time to remove our venerable tool from the tree. app-tools is
the preferred way to create packages now.

The entirety of app/tools/android has been removed. Care has been taken
to change as little as possible outside that directory: several targets
have been removed from the gyp files, references to make_apk were
dropped from comments and generate_app_packaging_tool.py no longer adds
make_apk & friends to xwalk_app_template (which is the basis for the
.zip files we ship in 01.org).

make_apk's test suite contained a few test extensions that might be
useful elsewhere; XWALK-6132 is tracking whether it makes sense to
revive them.

Related to: XWALK-5629


rakuco:
@baleboy @darktears @kenchris

baleboy:
LGTM!

rakuco:
the make_apk_test failures in the bots are expected: I'll remove the make_apk code from our bots before landing

PR #3488, opened 2016-01-13
Enable computeHorizontalScrollRange related API except computeScroll

shaoboyan:
Delegate computeHorizontalScrollRange related methods to ContentView.

BUG=XWALK-4394

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -713,6 +713,26 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
         mContentView.scrollBy(x, y);
     }
 
+    public int computeHorizontalScrollOffset() {
+        return mContentView.computeHorizontalScrollOffset();
+    }
+
+    public int computeHorizontalScrollRange() {
+        return mContentView.computeHorizontalScrollRange();
+    }
+
+    public int computeVerticalScrollExtent() {
+        return mContentView.computeVerticalScrollExtent();
+    }
+
+    public int computeVerticalScrollOffset() {
+        return mContentView.computeVerticalScrollOffset();
+    }
+
+    public int computeVerticalScrollRange() {
+        return mContentView.computeVerticalScrollRange();
+    }
+
     //--------------------------------------------------------------------------------------------
     private class XWalkIoThreadClientImpl extends XWalkContentsIoThreadClient {
         // All methods are called on the IO thread.
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentView.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentView.java
@@ -102,4 +102,29 @@ public class XWalkContentView extends ContentView {
         mXWalkView.onFocusChangedDelegate(gainFocus, direction, previouslyFocusedRect);
         super.onFocusChanged(gainFocus, direction, previouslyFocusedRect);
     }
+
+    @Override
+    public int computeHorizontalScrollOffset() {
+        return super.computeHorizontalScrollOffset();
+    }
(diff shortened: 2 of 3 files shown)


shaoboyan:
@axinging @XiaosongWei @wuhengzhi  PTAL, thx

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -713,6 +713,26 @@ public void scrollBy(int x, int y) {
         mContentView.scrollBy(x, y);
     }
 
+    public int computeHorizontalScrollRange() {
```
axinging: Order:  computeHorizontalScrollOffset/computeHorizontalScrollRange, 
computeVerticalScrollExtent/computeVerticalScrollOffset/computeVerticalScrollRange.

And test case?

shaoboyan: in XWALK-4394, I saw it related to XWALK-5417 which description as TC dev for feature XWALK-4394. So the test case is not include.

shaoboyan:
@axinging order modified done!

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
axinging: I refered to test case here: xwalk/test/android/core/javatests/src/org/xwalk/core/xwview/test/.

This is different from QA's.  But this raise a good question,  is there any possiblity to align test cases under xwalk/test/android/core/javatests/src/org/xwalk/core/xwview/test/ and QA's test?  @lincsoon  @fujunwei @baleboy

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -1424,6 +1424,36 @@ public void scrollBy(int x, int y) {
         mContent.scrollBy(x, y);
     }
 
+    @Override
+    @XWalkAPI
```
wuhengzhi: Does the comment is needed? e.g return, api version.

lincsoon: Yes, all need comment

shaoboyan: All right, I'll add them.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
shaoboyan: Since it's not easy for testing Scroll related API, I'd like to use an sample for testing.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentView.java
```diff
@@ -102,4 +102,29 @@ protected void onFocusChanged(boolean gainFocus, int direction, Rect previouslyF
...
+
+    @Override
+    public int computeVerticalScrollOffset() {
+        return super.computeVerticalScrollOffset();
+    }
+
+    @Override
+    public int computeVerticalScrollRange() {
+        return super.computeVerticalScrollRange();
+    }
```
lincsoon: If there's no special transition, no need to implement these methods in XWalkContentView.

shaoboyan: since these APIs are all protected in contentView, I think these public APIs are need.

lincsoon:
@axinging I think it's different between our unit test and QA's test. One is run by the try-bot for every commit, one is managed by QA for their test cycle. There might be some overlap, but no need to unify them all.

shaoboyan:
I'll make another PR later to fix these problems and some logic problems. Thx all.

PR #3489, opened 2016-01-13
[Android] Fix crash.

darktears:
Make sure to use the content settings only for desktop platforms
as Android does not need it.


rakuco:
rs=me

PR #3490, opened 2016-01-13
[Windows][Linux] Favor application name when asking permission.

darktears:
Instead of originating URL. If an application is launched from the
manifest.json then the URL is something like app:// and it's not
really what we want to show to the user.


darktears:
@pozdnyakov what do you think?

pozdnyakov:
lgtm

PR #3491, opened 2016-01-13
[Linux][Windows] Save permissions for applications launched with the

darktears:
manifest.

As they are with the app:// protocol we need to tell content_settings
to allow it.


darktears:
@pozdnyakov PTAL

pozdnyakov:
lgtm

PR #3493, opened 2016-01-15
[Android] Switch to the non-deprecated version of JSONReader::Read().

rakuco:
Prepare for the M48 update by switching away from
`JSONREader::DeprecatedRead()`, which is gone in that milestone.

In `XWalkContent::SetManifest()`, we just need to receive a `scoped_ptr`
instead, and it is also possible to simplify the code:
`manifest_dictionary` and `manifest_dictionary_ptr` are unnecessary and
only clutter the code.

### DIFF
--- a/runtime/browser/android/xwalk_content.cc
+++ b/runtime/browser/android/xwalk_content.cc
@@ -281,18 +281,13 @@ jboolean XWalkContent::SetManifest(JNIEnv* env,
   std::string json_input =
       base::android::ConvertJavaStringToUTF8(env, manifest_string);
 
-  base::Value* manifest_value = base::JSONReader::DeprecatedRead(json_input);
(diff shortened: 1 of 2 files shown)


rakuco:
@pozdnyakov @Bysmyyr

Bysmyyr:
One comment: change from Value to Dictionaryvalue can be done nicely with this: https://code.google.com/p/chromium/codesearch#chromium/src/base/values.h&q=Dictionaryvalue::from&sq=package:chromium&type=cs&l=214

rakuco:
I know, but that's not present in M46 which we use in master. Converting to `From` needs to be done only after we move to M48.

Bysmyyr:
ah, thats so new, Ok

pozdnyakov:
lgtm

PR #3495, opened 2016-01-18
[Android] Fix the issue that video can't play with blob url.

wuhengzhi:
Blob data is stored in Crosswalk cache, the cache path is local address,
e.g "/sdcard/cache/" or other relative urls, its scheme is null.
In Crosswalk, "ResourceLoadingFilter" has been implemented to customize
the resource url, if resource url is cache path, the scheme is null,
then
"uri.getScheme().equals()" will lead to null pointer exception and be
caught in "MediaPlayerBridge.java:setDataSource()", data source can't be
set exactly, then report media error format.
Add test case for blob url to make sure the video from indexedDB can be
played successfully.

BUG=XWALK-6004

(cherry picked from commit 3a0f3343d5ad86f64650335708a9b99c5c1f7f68)


wuhengzhi:
@rakuco @halton PTAL and merge, thanks.

PR #3497, opened 2016-01-21
[Windows] Maximize the window over the taskbar in the fullscreen mode.

joone:
It allows the users to close the fullscreen application.

BUG=XWALK-4861

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -193,7 +193,10 @@ void Application::SetWindowShowState<Manifest::TYPE_MANIFEST>(
 
   if (display_string == values::kDisplayModeFullscreen) {
     params->display_mode = blink::WebDisplayModeFullscreen;
-    params->state = ui::SHOW_STATE_FULLSCREEN;
+    // Set ui::SHOW_STATE_MAXIMIZED instead of ui::SHOW_STATE_FULLSCREEN
+    // because there is no way to exit from full screen mode.
+    // Refer to XWALK-4861 for more details.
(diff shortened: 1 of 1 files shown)


joone:
It needs https://github.com/crosswalk-project/chromium-crosswalk/pull/309

application/browser/application.cc
```diff
@@ -193,7 +193,10 @@ void Application::SetWindowShowState<Manifest::TYPE_MANIFEST>(
 
   if (display_string == values::kDisplayModeFullscreen) {
     params->display_mode = blink::WebDisplayModeFullscreen;
-    params->state = ui::SHOW_STATE_FULLSCREEN;
+    // Set ui::SHOW_STATE_MAXIMIZED instead of ui::SHOW_STATE_FULLSCREEN
+    // because there is no way to exit from full screen mode.
+    // Refer to XWALK-4861 for more details.
+    params->state = ui::SHOW_STATE_MAXIMIZED;
```
pozdnyakov: this would affect appearance on linux. And IMO this is wrong: fullscreen != maximized. we just need a UI element for the user to close the window..

joone: I know, but I couldn't find a way to add the title bar in full screen state for Windows. I should add a guard only for Windows.

joone: here is the update CL:
https://github.com/crosswalk-project/crosswalk/pull/3501

joone:
This CL was updated to https://github.com/crosswalk-project/crosswalk/pull/3501

rakuco:
Why not just update this pull request?

PR #3501, opened 2016-01-22
[Windows] Maximize the window over the taskbar in the fullscreen mode.

joone:
It allows the users to close the fullscreen application.

BUG=XWALK-4861

### DIFF
--- a/application/browser/application.cc
+++ b/application/browser/application.cc
@@ -193,7 +193,12 @@ void Application::SetWindowShowState<Manifest::TYPE_MANIFEST>(
 
   if (display_string == values::kDisplayModeFullscreen) {
     params->display_mode = blink::WebDisplayModeFullscreen;
-    params->state = ui::SHOW_STATE_FULLSCREEN;
+    // Set ui::SHOW_STATE_MAXIMIZED instead of ui::SHOW_STATE_FULLSCREEN
+    // because there is no way to exit from full screen mode.
+    // Refer to XWALK-4861 for more details.
+#if defined(OS_WIN)
+    params->state = ui::SHOW_STATE_MAXIMIZED;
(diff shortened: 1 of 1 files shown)


application/browser/application.cc
```diff
@@ -193,7 +193,12 @@ void Application::SetWindowShowState<Manifest::TYPE_MANIFEST>(
 
   if (display_string == values::kDisplayModeFullscreen) {
     params->display_mode = blink::WebDisplayModeFullscreen;
-    params->state = ui::SHOW_STATE_FULLSCREEN;
+    // Set ui::SHOW_STATE_MAXIMIZED instead of ui::SHOW_STATE_FULLSCREEN
+    // because there is no way to exit from full screen mode.
+    // Refer to XWALK-4861 for more details.
+#if defined(OS_WIN)
```
pozdnyakov: looks like it breaks fullscreen on linux desktop

application/browser/application.cc
```diff
@@ -193,7 +193,12 @@ void Application::SetWindowShowState<Manifest::TYPE_MANIFEST>(
...
+    params->state = ui::SHOW_STATE_MAXIMIZED;
```
pozdnyakov: It should remain ui::SHOW_STATE_FULLSCREEN, and new ui elements shall be introduced for the fullscreen window..maybe a buttons sliding down when the user is moving the mouse over the top of the window or so..  @darktears wdyt?

joone: We're going to use metro driver for that kind of UI.

rakuco:
What's the status of this pull request?

darktears:
@joone ???

joone:
We don't need this commit.

PR #3502, opened 2016-01-22
[Android] Fix audio can't play issue.

wuhengzhi:
For the uri of "app://", it was converted to "file://", so the scheme
should be re-got to check.

BUG=XWALK-6173


wuhengzhi:
@lincsoon @axinging @fujunwei @halton PTAL, thanks.

lincsoon:
LGTM

PR #3505, opened 2016-01-26
Change scoped_ptr.Pass() to std::move(scoped_ptr)

Bysmyyr:
Scoped ptr is deprecated in M49 and it is easier to
fix them directly here. There will be most
probably another similar commit because it is easier to
find them from build failures.


Bysmyyr:
@pozdnyakov PTAL

pozdnyakov:
lgtm

PR #3512, opened 2016-01-27
Move the variable setting logic from gyp_xwalk to build/common.gypi.

rakuco:
Setting variables in gyp_xwalk was a historical mistake and likely came
from failures to understand GYP's bizantine scoping and evaluation
rules. Additionally, gyp_xwalk will not be used once we move to GN, so
this logic needs to be moved anyway.

The following variables were being set in gyp_xwalk to the values they
already have as default in `src/build/common.gypi` and have thus just been
removed from gyp_xwalk and not added to `src/xwalk/build/common.gypi`:
 * `notifications`
 * `use_aura`
 * `use_openmax_dl_fft`

Quick explanation about GYP's weird scoping and variable evaluation
rules: one cannot set a variable in a scope and use it in a conditional
block in the same scope level, or set it in a conditional block and use
it in another check in the same block and scope level. In other words,
both conditionals below will evaluate to false:
```python
  'variables': {
    'foo%': 1,
    'conditionals': [
      ["foo == 1", {
        'bar%': 42,
      }],
      ["bar == 42", {
        'baz%': 34,
      }],
    ],
  }
```
For it to work, we need to define some variable-within-variable blocks
so that the variables and their values are properly seen by the
conditional blocks. To complicate things further, since this .gypi file
is automatically merged into all other .gyp and .gypi files we need to
put any variables we set in a conditions block into another variables
scope to avoid the situation where another .gypi file checks for the
variable's value in the same scope we're setting it in a conditional.

A lot of comments were added to the file to make this whole mess clear
to future readers.

Also related to the above, we are now including `src/build/common.gypi`
before `src/xwalk/build/common.gypi`, so that the values we set in the
latter do not get overridden by the former.

Testing done: run gyp_xwalk without this patch, build a target like
"xwalk", apply the patch, run gyp_xwalk again. Building the target again
should do nothing (on Android it's easier to pass `-Denable_webcl=0`,
otherwise `-DENABLE_WEBCL=1` will be passed in another position in the
command line and needlessly make everything be rebuilt).

Related BUG=XWALK-3674

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -1,41 +1,135 @@
+# This file is similar to src/build/common.gypi. It contains common variable
+# definitions and other actions that are then used in other build system files.
+#
+# The values can be overriden on the GYP command line (-D) or by setting them
+# in ~/.gyp/include.gypi.
+#
+# IMPORTANT: Like src/build/common.gypi, this file must not be included by
+# other .gyp or .gypi files, it is included automatically when gyp_xwalk is
+# called. This file is always included immediately after src/build/common.gypi
+# itself so that variables originally set there can have their values
+# overridden here.
+
 {
+  # Organization of the variables below:
+  # 1. Variables copied from inner scopes.
+  # 2. New variables defined in current scope.
+  # 3. Conditions.
+  # Some variables are new, but most come from other .gyp files and are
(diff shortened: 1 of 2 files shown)


rakuco:
@darktears @halton

PR #3513, opened 2016-01-28
[Android] Fix onReceivedHttpAuthRequest() was not called issue.

wuhengzhi:
This patch is to fix the issue about onReceivedHttpAuthRequest() was not
called when this function was overridden in class of XWalkResourceClient.
Modify the attribute of XWalkHttpAuthHandlerInternal to generate right
parameter for XWalkResourceClientBridge.
Move test case from XWalkViewInternal to XWalkView.

BUG=XWALK-5117

(cherry picked from commit b89f5dd217fddb9b7f8ba84ca89383a09c37f8eb)


wuhengzhi:
@rakuco Please help to review and merge to crosswalk 17, thanks.

PR #3514, opened 2016-01-28
Remove deprecated base/basictypes.h includes part 1

Bysmyyr:
base/basictypes.h is deprecated and will be removed from Chromium in m49. Mostly it is not
indlude all of them.


Bysmyyr:
@pozdnyakov PTAL

pozdnyakov:
lgtm

PR #3468, opened 2015-12-24
[Windows] Enable widevine support

qjia7:
This PR will let xwalk register widevine CDM component and add
corresponding key system. Based on this, Xwalk will have the ability
to play content protected media with widevine CDM.

The steps to verify this feature:
1. python gyp_xwalk.py -Dmediacodecs_EULA=1
2. ninja -C ../out/Debug xwalk.exe
3. Run '../out/Debug/xwalk
http://shaka-player-demo.appspot.com/?dash;asset=assets/car_segmenttemplate.mpd'

After above three steps, you will see the video can be played well.
Notice that adding mediacodecs_EULA=1 flag is because the test video is mp4
format. If your test video is webm, this flag is not needed.(We have opened -Denable_widevine=1 in gyp_xwalk )

BUG=XWALK-6102

PR #3516, opened 2016-01-28
[Windows] Disable widevine for windows

qjia7:
This reverts commit 53a0752.

Conflicts:
	gyp_xwalk

There are still some unresovled issues to bring in widevine to windows.
1. In current design, we utilize the component updater to download the
widevine component from google's server which will include a ~7M
widevinecdm.dll that will result that the first update consts several minutes
which is a bad user experience. Besides that, even we don't use widevine,
the download process still happen.
2. There is no prompt machanism to tell users the download process. Once
the download finished, users have to refresh their apps to use this feature.
3. If we set the component as autoupdate, it will ping google's server
every 6 hours to check whether to need to update WidevineCDM (In fast-update mode,
it will happen every 60 seconds)
4. Since we need to download some files from google's server, it maybe
exist ip issue.

Some details discussion can be found in XWALK-6207. So the PR will
revert widevine support untill we are ready for it.


rakuco:
Thanks for addressing this. Doesn't it make sense to just revert #3468 altogether? Since XWALK-6207 is currently restricted, I think you could also briefly mention what are the most pressing issues that have led to this change.

qjia7:
>Doesn't it make sense to just revert #3468 altogether?

I think just disable widevine  is also ok. If we don't open this flag, it won't build widevine part. So the corresponding dll file also won't be generated. All have merged part are open source, I think it should be no ip issue. Anyway, I have reverted #3468. Please take a look. Thanks.

rakuco:
lgtm, thanks.
The reason I suggested reverting everything altogether is that we may end up implementing things differently in the end, and until then we have dead code in the tree for no reason. Reverting isn't bad, and it happens all the time in Chromium itself.

PR #3524, opened 2016-02-02
Fix build warning: initialize order

Bysmyyr:
warning: field 'initialized_' will be initialized after
field 'library_path_' [-Wreorder]


rakuco:
lgtm

PR #3525, opened 2016-02-02
[Android] Stop setting ENABLE_WEBCL=1 in all files.

rakuco:
With chromium-crosswalk commit 64b4265 ("[Blink] WebCL: Only set
ENABLE_WEBCL in the relevant files"), the `ENABLE_WEBCL` macro is defined
only for the relevant files in Blink.

This means we can now remove it from `target_defaults` and stop passing
it to all the 15000+ files part of Crosswalk's build since the vast
majority of them do not need to be aware of the macro at all.


rakuco:
@shaoboyan @darktears

darktears:
LGTM of course

shaoboyan:
LGTM.

PR #3526, opened 2016-02-03
Fix build warning: initialize order

Bysmyyr:
"warning: field 'handle_binary_msg_callback_'
will be initialized after field 'library_path_' [-Wreorder]"


Bysmyyr:
PTAL @rakuco

rakuco:
`library_path_` is defined before `xw_extension_` in the header file, are you sure you won't get another warning with this change?

rakuco:
lgtm

PR #3527, opened 2016-02-03
[Android] Start shipping VERSION in the app-template zip file again.

rakuco:
It was thought to be used only by make_apk, so we stopped putting it in
the xwalk_app_template directory (and later publishing it in the zip
file) when make_apk was removed in commit acbf950 ("[Android] Remove
make_apk and associated tests from the tree") , but app-tools can use it
as a hint deduce Crosswalk's version number.


darktears:
SGTM

PR #3536, opened 2016-02-05
Remove GetDefaultCSPInfo().

rakuco:
This was only consumed by Tizen code that got removed in
3500f1d ("[Tizen] Remove support of Tizen"), and its presence is causing
build errors with the upcoming M49 because of -Wunused-function.


pozdnyakov:
lgtm

PR #3537, opened 2016-02-05
Remove GetNativeModule().

rakuco:
This function was added in 1959733 ("Add application extension and
getManifest(), getMainDocument() API support") but was never used
anywhere.

Get rid of it so that M49 builds do not fail due to `-Wunused-function`.


pozdnyakov:
lgtm

PR #3540, opened 2016-02-08
Remove deprecated base/basictypes.h includes part 2

Bysmyyr:
base/basictypes.h is removed from Chromium. Mostly it is not
indlude all of them.


runtime/browser/xwalk_pref_store.cc
```diff
@@ -71,6 +71,7 @@ PersistentPrefStore::PrefReadError XWalkPrefStore::ReadPrefs() {
 void XWalkPrefStore::ReadPrefsAsync(ReadErrorDelegate* error_delegate_raw) {
 }
 
-void XWalkPrefStore::ReportValueChanged(const std::string& key, uint32 flags) {
+void XWalkPrefStore::ReportValueChanged(const std::string& key,
+                                                 uint32_t flags) {
```
rakuco: You probably want to align this line to the `const std::string&` above.

Bysmyyr:
@pozdnyakov PTAL

rakuco:
lgtm

PR #3541, opened 2016-02-08
Remove deprecated base/basictypes.h includes part 3

Bysmyyr:
base/basictypes.h is removed from Chromium. Mostly it is not
indlude all of them.


Bysmyyr:
@pozdnyakov PTAL

rakuco:
Can't this just be merged into #3540?

PR #3544, opened 2016-02-09
Change scoped_ptr.Pass() to std::move(scoped_ptr) part4

Bysmyyr:
Scoped ptr is deprecated in M49 and it is easier to
fix them directly here. There will be most
probably another similar commit because it is easier to
find them from build failures.

### DIFF
--- a/application/browser/application_service_provider_linux.cc
+++ b/application/browser/application_service_provider_linux.cc
@@ -70,7 +70,7 @@ void OnPing(dbus::MethodCall* method_call,
       dbus::Response::FromMethodCall(method_call);
   dbus::MessageWriter writer(response.get());
   writer.AppendString("Pong");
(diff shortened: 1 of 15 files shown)


rakuco:
Are there any other remaining parts after this one? If so, what's the benefit of landing this in parts versus changing everything in one PR?

Bysmyyr:
Those are all from linux build. There can be some when building with android. Those was just miss from last one. I can wait if that makes things much easier.

rakuco:
I don't really see the value in splitting this change into several pull requests, at least I don't see any logical separation being done. So if you have any additional Android changes, please add them to this PR so we change everything that's left at once.

Bysmyyr:
Now there should be all.

rakuco:
lgtm, let's do this

PR #3549, opened 2016-02-11
[Win] Add a target to generate a Crosswalk ZIP file.

rakuco:
The new `generate_crosswalk_win_zip` target, built as part of the
`xwalk_builder` target on Windows, invokes a script that creates a ZIP
file (called `crosswalk_win.zip`) with everything required to run
Crosswalk on Windows. In other words, it's a zip file with the same
contents we publish on download.01.org (without a version number or
signed files, though).

Doing so helps with generating Crosswalk canaries for Windows as part of
our release building process.


darktears:
LGTM

PR #3554, opened 2016-02-16
Remove the notification of NOTIMPLEMENTED().

wuhengzhi:
This patch is to remove the notification of not implemented function for
OnFirstUserGestureObserved().
The notification will be displayed in logcat when user touch in XWalkView.
It's misleading when user encounter other issues, e.g crash, features
don't work.

### DIFF
--- a/runtime/browser/xwalk_autofill_client.cc
+++ b/runtime/browser/xwalk_autofill_client.cc
@@ -113,7 +113,6 @@ void XWalkAutofillClient::DidFillOrPreviewField(
 }
 
 void XWalkAutofillClient::OnFirstUserGestureObserved() {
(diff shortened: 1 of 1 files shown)


rakuco:
While this solves this specific issue mentioned in XWALK-6067, we still have dozens of other NOTIMPLEMENTED() calls in our tree that can be equally confusing.
I believe the best solution would be for someone to take on XWALK-6033 and check whether we can build our official releases with the `OFFICIAL_BUILD` macro set.

wuhengzhi:
Agree with that, we should investigate how to build with `OFFICIAL_BUILD`.

rakuco:
Closing this PR following the comment above.

PR #3555, opened 2016-02-16
Make sure to handle all the required messages in our devtools frontend

darktears:
This will unbreak the inspector on desktop platforms. Note that
the message has been added back in September 2015
(https://codereview.chromium.org/1362943005) but only broke when we moved
to M48 (probably now required due to work on
https://code.google.com/p/chromium/issues/detail?id=529471).

BUG=XWALK-6307


darktears:
@rakuco PTAL

rakuco:
lgtm

PR #3561, opened 2016-02-18
[Android] Remove deprecated base/basictypes.h includes

Bysmyyr:
base/basictypes.h is removed from Chromium in m49. Mostly it is not
needed, sometimes old types are replased with _t types. There
is still 4 mac and win uses of these.


rakuco:
lgtm

PR #3566, opened 2016-02-22
Fix FileApiPersistent API SecurityError.

darktears:
When launched through the manifest Crosswalk apps
are using the app:// protocol so we need to add it
as part of the allowed schemes for the API.


darktears:
@pozdnyakov maybe you can review?

pozdnyakov:
lgtm. Though I wish we had platform-specific files instead of operating with ifdefs

PR #3567, opened 2016-02-24
Intt

Bysmyyr:
Change last basictybes.h ints to std versions
    
basictypes.h is removed in m49 so we need to change all old
int types to _t versions.


pozdnyakov:
lgtm

PR #3568, opened 2016-02-24
[Android] Add isFirstAttempt as part of public XWalkHttpAuthHandler API

marcuspridham:
The Android HttpAuthHandler exposes if this is the first auth attempt
through the method useHttpAuthUsernamePassword. The Crosswalk
XWalkHttpAuthHandler also has this information but it is
currently not made available to the public API.

BUG=XWALK-6389


fujunwei:
LGTM, thanks.
@wuhengzhi PTAL.

XiaosongWei:
LGTM

marcuspridham:
One use case:
1. isFirstAttempt is true.  If you have old credentials that you want to try you can use them at this point.
2. isFirstAttempt is false.  Old credentials have failed need to acquire new ones through a prompt or service.

PR #3572, opened 2016-02-25
Remove application_service_provider_linux.{cc,h}.

rakuco:
These files have not been used since 192fcaca ("Remove the D-Bus code
from Crosswalk").


darktears:
LGTM

PR #3573, opened 2016-02-25
Add a "_linux" suffix to xwalk_notification_manager.{cc,h}.

rakuco:
Doing that allows it to be automatically excluded from builds based on
the rules in `src/build/filename_rules.gypi`, and we do not have to add
those files to the build system in a special condition check.


darktears:
LGTM

PR #3577, opened 2016-02-26
[Android] Remove "All" target.

rakuco:
It was added with commit 6863ff1b ("Bump crosswalk to Chromium
34.0.1809.2"), but it is not used anywhere (the `xwalk` target it
depended on does not even work on Android).


darktears:
lgtm

PR #3579, opened 2016-03-01
[Lite] Move compressed library file into lib folder

heke123:
In Lite, the compressed library is placed in /res/raw as a resource
file. It is named as libxwalkcore.so.x86 or libxwalkcore.so.armeabi_v7a.
This causes several problem:
1) In a "bundled(contains both libs for x86 and arm)" aar or webview.zip, these two files are placed together,
   Android regards them as duplicated resources.
2) The gradle of latest version regards it as illegal resource name.
3) Gradle cannot build multiple apks because it cannot filter resource
   files for different abi target.

In this PR, we don't put it res/raw any more, we rename it, put it in
lib/x86/libxwalkcoreCompressed.so (x86 for example) and modify accordingly
in make_apk.py and in java class which handles decompressing in runtime.

One WEAKNESS: During installation, Android will copy the compressed lib
 into a read only folder, which occupy more flash of the phone.

BUG=XWALK-6379


heke123:
@halton  @PeterWangIntel  Please help give a review, thanks.

heke123:
One idea:  If we don't provide "bundled" aar to user, if we provide xxx-x86.aar and xxx-arm.aar separately,  then we don't need this modification.  Bundled aar apparently increase apk size, it is not so meaningful especially for Lite.
@halton  Could you please help make decision for us? thanks.

halton:
@heke123 Does bundled aar mean developer will build bunlded Crosswalk app? If yes, I would go separated aar.

heke123:
@halton  Yes, user will get a bundled and bigger apk. The advantage is this one apk can run in both x86 and arm, while the weakness is the apk is bigger because it contains multiple .so files for different target.

heke123:
With this modification, all the situations will be supported, but occupy more flash of the phone.

halton:
> With this modification, all the situations will be supported, but occupy more flash of the phone.

If this PR fix both cases for bundled and by-arch aar issues, I would suggest go ahead to review and merge.

halton:
@PeterWangIntel please review.

PeterWangIntel:
LGTM as face to face review.

lianghanzhen:
I don't think it is a good solution.
We should rename so files as format libxwalkcore_(cpu_arch).so, find compressed so file according to cpu arch and decompress it.

heke123:
@lianghanzhen   No need to rename the so file.  Android package manager service can find the right so file for the right target. The so files is placed in the cpu_arch folders like .../x86/libxwalkcoreCompressed.so and .../arm/libxwalkcoreCompressed.so
Did I understand your comment correctly?

lianghanzhen:
@heke123 No, you don't. SO file can not be placed in raw folders because its name has two dots, it breaks the compile process. we can rename it like libxwalkcore_x86.so and decompress it according to cpu arch (R.raw.libxwalkcore_x86 or something), this is the simplest way.

heke123:
@lianghanzhen    The gradle supports build multiple apks for x86 and arm separately. If we just rename it, both the libxwalkcore_x86.so and libxwalkcore_arm.so are regarded as resources files and will be packaged even if user wants to generate apk for x86 only(or for arm only),  user will get a bigger apk which contains a redundant compressed lib.

lianghanzhen:
Yes, it will. What about providing different aars or libraries? If not, users have other options to package according cpu arch.

heke123:
Now all the situations are supported.  Both different aars and separate aars.

ks32:
Crosswalk-lite 17.46.455.1 aar is giving error couldn't find "libxwalkcore.so".

```
22 16:03:23.040 7840-7880/com.testt I/OpenGLRenderer: Initialized EGL, version 1.4
03-22 16:03:23.080 7840-7880/com.testt D/OpenGLRenderer: Enabling debug mode 0
03-22 16:03:23.666 7840-7840/com.testt D/XWalkLib: Init embedded mode
03-22 16:03:23.667 7840-7840/com.testt D/XWalkLib: lib version, api:5, min api:5
03-22 16:03:23.667 7840-7840/com.testt D/XWalkLib: app version, api:5, min api:5
03-22 16:03:23.667 7840-7840/com.testt D/XWalkLib: XWalk core version matched
03-22 16:03:23.688 7840-7853/com.testt I/art: Background sticky concurrent mark sweep GC freed 7462(470KB) AllocSpace objects, 7(110KB) LOS objects, 4% free, 11MB/12MB, paused 7.996ms total 68.976ms
03-22 16:03:23.708 7840-7840/com.testt D/XWalkLib: java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.testt-1/base.apk"],nativeLibraryDirectories=[/data/app/com.testt-1/lib/arm, /vendor/lib, /system/lib]]] couldn't find "libxwalkcore.so"
03-22 16:03:23.708 7840-7840/com.testt D/XWalkLib: Mismatch of CPU architecture
03-22 16:03:23.715 7840-7840/com.testt D/AndroidRuntime: Shutting down VM
03-22 16:03:23.726 7840-7840/com.testt E/AndroidRuntime: FATAL EXCEPTION: main
```

heke123:
@ks32  Thanks for reporting this error.  Could you tell me your Android device version?

ks32:
Tested on Nexus 5X and Moto X 1st gen. Same error on both devices.

heke123:
@ks32  Sorry I cannot reproduce this error on my device(Xiaomi, Android 4.4). Could you please report this problem in JIRA?  Please describe the detail steps how did you setup your project. Thanks.

ks32:
@heke123 How to report problem in JIRA?
I have created a test project showing this error:
https://github.com/ks32/crosswalk-lite-test

heke123:
@ks32   I have created a JIRA to track this problem, let's discuss in https://crosswalk-project.org/jira/browse/XWALK-6671

PR #3580, opened 2016-03-01
[Windows] Fix application origin

pozdnyakov:
Before, the application URL origin (which is used in granting API
permissions) was generated from the installation path. Instead the
URL origin must be unique and it must remain the same for the app
wherever it is installed.
Now the application origin is generated from the upgrade id
GUID that remains the same through all the application versions.

BUG=XWALK-6442


pozdnyakov:
@darktears @r0b5t4 PTAL

pozdnyakov:
ping for review

PR #3581, opened 2016-03-02
[Android] Add support to enable/disable theme-color property

xzhan96:
Starting version 39 of Chrome for Android on Lollipop, the app that embeds WebView changes toolbar color if the opened web page uses theme-color. Crosswalk WebView also support it now, but in some cases, the feature is not desired, the problem is that the app developer has no control on the pages that are opened, so it would be nice if this feature can be configured in Crosswalk.

BUG=XWALK-6445

(cherry picked from commit deed2fec61e8c1eded7129b295df664626511b45)

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClient.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClient.java
@@ -61,7 +61,9 @@ abstract class XWalkContentsClient extends ContentViewClient {
 
         @Override
         public void didChangeThemeColor(int color) {
-            onDidChangeThemeColor(color);
+            boolean themecolor = XWalkPreferencesInternal.getValue(
+                    XWalkPreferencesInternal.ENABLE_THEME_COLOR);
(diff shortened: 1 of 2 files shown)


xzhan96:
@baleboy please help to merge it, thanks

wuhengzhi:
It seems that the cherry pick info was lost.

darktears:
Why you would want to configure it? You can just leave the meta-tag if you don't want that.

Crosswalk-18 is in beta, I don't think we should land a new API.

baleboy:
> Why you would want to configure it? You can just leave the meta-tag if you don't want that

The problem is if you have an app that embeds the crosswalk webview and opens a browser page. According to user reports, in that case the app changes color if the embedded page uses theme-color, which is not desired. The app developer has no control on the pages that are opened. In that sense it can be considered a bug because there is no way to avoid this effect and not having the theme-color feature at all is desirable in this case.

darktears:
ok then, would have been nice to explain that in the commit message.

xzhan96:
@baleboy @wuhengzhi @darktears please help to review, thanks

PR #3582, opened 2016-03-02
v8tools: Port the lifecycleTracker code to non-deprecated API.

rakuco:
V8 is deprecating its previous `v8::PersistentBase::SetWeak()` calls
that turned a persistent object into a weak reference. The new SetWeak()
API turns the handle into a phantom reference, a concept borrowed from
Java that means a callback set when calling SetWeak() will be run
_after_ the object being referenced has already been garbage-collected.
This is happening most likely due to performance reasons, as memory can
be reclaimed in one GC cycle instead of two. Relevant CLs and bugs:
 - https://bugs.chromium.org/p/v8/issues/detail?id=3246
 - https://codereview.chromium.org/687003005
 - https://codereview.chromium.org/753553002
 - https://codereview.chromium.org/989153003

For us, this means the lifecycleTracker code needs to be rewritten, as
we were relying on the referenced lifecycleTracker object still being
alive when running the cleanup callback in order to retrieve the
function associated with the object's `destructor` property.

What we do now is track the function assigned to the `destructor`
property separately at the time of the assignment and invoke it without
going through the referenced object during the cleanup.

The code gets more verbose, but I believe it makes things more clear:
- There's now a getter for the `destructor` property, which returns the
  function assigned, or null when there is none. Previously, if no
  destructor had been set accessing the property would return undefined.
- Related to this, the `destructor` property is no longer deletable.
  Users need to set it to null in order to remove a previously
  associated function. Combined with the above point, this makes it more
  explicit that `destructor` is not a regular property and is always
  present (and enumerable) in the object.
- The property's setter performs type checking on the value being
  assigned, raises a `TypeError` when an invalid type is passed and
  clears up the function if null is passed. Previously one could pass
  any type and the checking would only be done when the cleanup function
  was being run and there would be no JS error if an invalid type had
  been passed (only a debug warning in C++).

Related BUG=XWALK-6372

### DIFF
--- a/extensions/renderer/xwalk_v8tools_module.cc
+++ b/extensions/renderer/xwalk_v8tools_module.cc
@@ -23,6 +23,9 @@ namespace extensions {
 
 namespace {
 
+// ================
+// forceSetProperty
+// ================
 void ForceSetPropertyCallback(const v8::FunctionCallbackInfo<v8::Value>& info) {
   // FIXME(cmarcelo): is this the default?
   info.GetReturnValue().SetUndefined();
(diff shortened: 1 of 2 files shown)


rakuco:
Ugh, this is crashing `SysAppsRawSocketTest.SysAppsRawSocket`.

rakuco:
ping @darktears @huningxin @halton

halton:
I have to admit this PR is so great with code change and explanation.  I would love give my LGTM even I can not understand the code change totally.

rakuco:
Thanks, I'll wait a bit for @huningxin then.

huningxin:
@rakuco , Thanks for this PR! 

LGTM.

PR #3583, opened 2016-03-02
[Windows][Linux] Limit min/max zoom from viewport meta

pozdnyakov:
This patch brings 'user agent' values which override page
min and max zoom from meta viewport tag.

BUG=XWALK-6269


pozdnyakov:
@darktears @kenchris  PTAL

PR #3585, opened 2016-03-02
[Windows] Read the Google API keys from the registry to set

darktears:
them as environment variables.

When packing a crosswalk app with app-tools and a developer provided
their keys to access the Google APIs they are going to be added
in the registry. Make sure to fetch them in Crosswalk and set them
as env variables so they can be used later on during requests (such
as geolocation.)

### DIFF
--- a/runtime/browser/xwalk_runner.cc
+++ b/runtime/browser/xwalk_runner.cc
@@ -26,6 +26,10 @@
 #include "xwalk/runtime/common/xwalk_runtime_features.h"
 #include "xwalk/runtime/common/xwalk_switches.h"
 
+#if defined(OS_WIN)
+#include "xwalk/runtime/browser/xwalk_runner_win.h"
+#endif
+
 namespace xwalk {
 
 namespace {
@@ -156,11 +160,17 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
   main_parts->CreateInternalExtensionsForExtensionThread(
       host, &extension_thread_extensions);
 
+  InitializeEnvironmentVariablesForGoogleAPIs(host);
+
   scoped_ptr<base::ValueMap> runtime_variables(new base::ValueMap);
   InitializeRuntimeVariablesForExtensions(host, runtime_variables.get());
   extension_service_->OnRenderProcessWillLaunch(
       host, &ui_thread_extensions, &extension_thread_extensions,
       std::move(runtime_variables));
+}
+
+void XWalkRunner::InitializeEnvironmentVariablesForGoogleAPIs(
+    content::RenderProcessHost* host) {
 }
 
 void XWalkRunner::OnRenderProcessHostGone(content::RenderProcessHost* host) {
@@ -187,7 +197,11 @@ void XWalkRunner::DisableRemoteDebugging() {
 
 // static
 scoped_ptr<XWalkRunner> XWalkRunner::Create() {
+#if defined (OS_WIN)
+  return scoped_ptr<XWalkRunner>(new XWalkRunnerWin);
+#else
   return scoped_ptr<XWalkRunner>(new XWalkRunner);
+#endif
 }
 
 content::ContentBrowserClient* XWalkRunner::GetContentBrowserClient() {
--- a/runtime/browser/xwalk_runner_win.cc
+++ b/runtime/browser/xwalk_runner_win.cc
@@ -1,0 +1,65 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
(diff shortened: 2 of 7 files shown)


darktears:
@pozdnyakov @r0b5t4 PTAL

runtime/browser/xwalk_runner.cc
```diff
@@ -163,6 +169,48 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
       std::move(runtime_variables));
 }
 
+void XWalkRunner::InitializeEnvironmentVariablesForGoogleAPIs(
+    content::RenderProcessHost* host) {
+#if defined (OS_WIN)
```
pozdnyakov: Please create XWalkRunnerWin, let's not multiply ifdefs

runtime/browser/xwalk_runner.cc
```diff
@@ -163,6 +169,48 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
...
+  std::wstring registry_path(L"SOFTWARE\\");
+  // Assuming xwalk_package_id is ASCII.
+  registry_path.append(std::wstring(vendor.begin(), vendor.end()));
+  registry_path.append(L"\\");
+  registry_path.append(std::wstring(app_name.begin(), app_name.end()));
+  base::win::RegKey key(HKEY_CURRENT_USER,
+                        registry_path.c_str(),
+                        KEY_READ);
+  std::wstring google_api_key;
+  if (key.ReadValue(L"GOOGLE_API_KEY",
```
pozdnyakov: L"GOOGLE_API_KEY" and others deserve having a const variable

runtime/browser/xwalk_runner.cc
```diff
@@ -163,6 +169,48 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
...
+      GetApplicationByRenderHostID(host->GetID());
+  if (!app)
+    return;
+  std::string app_name;
+  if (!app->data()->GetManifest()->GetString("xwalk_package_id", &app_name))
+    return;
+  std::string vendor =
+      app_name.substr(0, app_name.find_last_of("."));
+  std::wstring registry_path(L"SOFTWARE\\");
+  // Assuming xwalk_package_id is ASCII.
```
pozdnyakov: might mot be true..

runtime/browser/xwalk_runner.cc
```diff
@@ -163,6 +169,48 @@ void XWalkRunner::OnRenderProcessWillLaunch(content::RenderProcessHost* host) {
...
+  application::Application* app =
+      app_system()->application_service()->
+      GetApplicationByRenderHostID(host->GetID());
+  if (!app)
+    return;
+  std::string app_name;
+  if (!app->data()->GetManifest()->GetString("xwalk_package_id", &app_name))
+    return;
+  std::string vendor =
+      app_name.substr(0, app_name.find_last_of("."));
```
pozdnyakov: can we make a simpler registry key (any that is existing in manifest), so that we do not parse xwalk_package_id (not a major problem but still..)?

runtime/browser/xwalk_runner.cc
darktears: Better idea to convert from a string to a wstring?

runtime/browser/xwalk_runner.cc
darktears: Sure.

runtime/browser/xwalk_runner.cc
darktears: Want to add inheritance for that?

runtime/browser/xwalk_runner.cc
pozdnyakov: http://stackoverflow.com/questions/2573834/c-convert-string-or-char-to-wstring-or-wchar-t
I would try std::wstringstream but you might prefer other ways..

darktears:
@pozdnyakov what about the new version?

runtime/browser/xwalk_runner_win.cc
```diff
@@ -0,0 +1,63 @@
...
+
+void XWalkRunnerWin::InitializeEnvironmentVariablesForGoogleAPIs(
+  content::RenderProcessHost* host) {
+  application::Application* app =
+      app_system()->application_service()->
+          GetApplicationByRenderHostID(host->GetID());
+  if (!app)
+    return;
+  std::string app_name;
+  if (!app->data()->GetManifest()->GetString("xwalk_package_id", &app_name))
```
pozdnyakov: nit: pls add a const variable to xwalk/application/common/application_manifest_constants.h/cc for "xwalk_package_id"

pozdnyakov:
LGTM thanks

darktears:
Indeed

PR #3586, opened 2016-03-03
extensions: Port away from deprecated V8 APIs.

rakuco:
In preparation for Chromium M49 (which will enable stricter compilation
flags for more code by default), port the extensions subsystem away from
deprecated V8 API calls. They will be removed soon, and it is better to
avoid being surprised late in the game.

While here, enable the `chromium_code` variable in `extensions.gyp` so
that we build all targets there (currently, only `xwalk_extensions`)
with more compiler warnings and errors so that we catch uses of
deprecated APIs earlier in the future.

Related BUG=XWALK-6276
BUG=XWALK-6372

### DIFF
--- a/extensions/extensions.gyp
+++ b/extensions/extensions.gyp
@@ -1,6 +1,9 @@
 {
   'targets': [
     {
+      'variables': {
+        'chromium_code': 1,
+      },
       'target_name': 'xwalk_extensions',
       'type': 'static_library',
       'dependencies': [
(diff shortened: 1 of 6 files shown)


rakuco:
This continues the work started in #3582.
Ping @darktears @halton @huningxin @kenchris

huningxin:
@rakuco , thanks much for working on this! 

Code LGTM.

PR #3589, opened 2016-03-04
Fix a -Wreorder error after ec0720da.

rakuco:
Commit ec0720da ("[Windows] Fix application origin") introduced an
ordering issue when initializing `ApplicationData`'s members which
causes build failures with stricted compiler flags (as it happens by
default in M49):

```
../../xwalk/application/common/application_data.cc:109:7: error: field 'application_id_' will be initialized after field 'path_' [-Werror,-Wreorder]
      application_id_(id),
```


rakuco:
@pozdnyakov ptal

pozdnyakov:
thanks for catch

PR #3590, opened 2016-03-04
xwalk_browsertest: Port to embedded_test_server.

rakuco:
Upstream is moving towards removing `test_server()` and moving everything
to `embedded_test_server()`. See, for example, https://codereview.chromium.org/1486633002.

Do the same for `xwalk_browsertest`. There should be no visible effects
after this change.

BUG=XWALK-6288


pozdnyakov:
lgtm

PR #3593, opened 2016-03-07
v8tools: Run the lifecycleTracker destructor asynchronously.

rakuco:
`SysAppsRawSocketTest.SysAppsRawSocket` was crashing on M49 after commit
92bdf3c ("v8tools: Port the lifecycleTracker code to non-deprecated API") when
built in debug mode.

It turns out our weak callbacks (before and after turning them into phantom
callbacks in the commit mentioned above) are run between Blink's garbage
collection prologue and epilogue. The prologue forbids script execution, and
the epilogue reenables it. In debug mode, when all asserts are on, the checks
for whether JS code is being run fail and Crosswalk crashes.

Commit d43f001 ("Fix XWalkExtensionsV8ToolsTest.V8ToolsWorks asserting in
debug") was an early indication of this issue, and was working around the issue
by somewhat blindly disabling some of those checks. That was not enough, and
doing something as simple as calling `console.log()` from the destructor
callback was enough to crash Crosswalk.

We now adopt the same solution used by Chromium's extension system and its
`GCCallback` class: instead of running the code that calls the destructor
inside the weak callback we've registered, we post a task to the message loop
and run it asynchronously _after_ the garbage collection code has run.

Some of our tests had to be rewritten to take into account the fact that the
destructor functions are no longer run synchronously, which is why this
commit got quite big (the changes to the v8tools module itself are
really small).

### DIFF
--- a/extensions/renderer/xwalk_v8tools_module.cc
+++ b/extensions/renderer/xwalk_v8tools_module.cc
@@ -50,11 +50,14 @@ void LifecycleTrackerCleanup2(
     const v8::WeakCallbackInfo<LifecycleTrackerWrapper>& data) {
   LifecycleTrackerWrapper* wrapper = data.GetParameter();
   if (!wrapper->destructor.IsEmpty()) {
+    v8::HandleScope handle_scope(data.GetIsolate());
     v8::Local<v8::Context> context = v8::Context::New(data.GetIsolate());
-    v8::TryCatch try_catch(data.GetIsolate());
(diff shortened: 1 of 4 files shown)


rakuco:
I *really* hope this is the last pull request related to this; V8 is definitely not for the faint of heart.

ping @darktears @halton @huningxin

huningxin:
Thanks for the efforts!

LGTM.

PR #3598, opened 2016-03-09
[Lite]Add fake APIs to make Lite-17 compatible with Cordova Plugin 1.7.0

heke123:
New XWALK APIs are added in Crosswalk-18, accordingly the Cordova
Plugin uses those APIs. It complains compile error because in Lite-17
those new added APIs are not exist.

So we add fake APIs into Lite-17 to make it compatible with Cordova Plugin.
This modification is not needed in Lite-19.

BUG=XWALK-6488


heke123:
@PeterWangIntel  @halton  Please help to review, thanks.

PeterWangIntel:
LGTM， thx

PR #3599, opened 2016-03-09
Roll DEPS.xwalk

heke123:
Chromium-crosswalk:
6156aae51f40807bd969baed541fc004c80f2021
  [Lite] Set aapt to do compressing on all the apk contents.


heke123:
I'll continue to investigate if we can do something in AppTool or MakeAPK.py to reduce size.

heke123:
@halton  @PeterWangIntel

PR #3602, opened 2016-03-14
[Android] Change color of root layer to avoid white screen on startup

axinging:
If set the WindowBackground of Activity to RED, and webpage page's backgroudn to RED,
when application starts or resumes after click back key, it will showup a white screen
shortly. We can change the first screen to a fix color by setting the root layer color
of ContentViewCore. This can help to avoid white screen.

BUG=XWALK-4809, XWALK-4995

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = 'dfbc583e04cd1530d2cd6ec6721a8eea6227770c'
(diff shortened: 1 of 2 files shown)


axinging:
@lincsoon  @fujunwei @XiaosongWei @wuhengzhi

rakuco:
Please split the DEPS roll and the code itself into two separate commits. This makes things more logical and also easier to backport if necessary (you only need to cherry-pick the code change without any conflicts in DEPS.xwalk).

axinging:
@rakuco  split is done.

PR #3603, opened 2016-03-15
Fix the xwalk build target on android

mrunalk:
Not sure if this is important or should be workign on Android but atleast
it will be good to fix if anyone builds xwalk target.

BUG=XWALK-6531


rakuco:
Like I've just written in JIRA, the `xwalk` target does not generate anything useful on Android. With that said, I don't oppose at least getting it to build.

This has always been broken, so I suggest you send a PR against the master branch and include the error you get in the commit/PR messages.

mrunalk:
Agreed, not that important from Android point of view but just wanted to fix it. Closing this PR here and will open a new one with updated commit message for master. Thanks.

PR #3608, opened 2016-03-17
M49 gardening: Remove redundant inludes for java_apk.gypi

mrunalk:
Strangely these redundant includes are causing the build failure
mentioned in XWALK-6547

BUG=XWALK-6547

### DIFF
--- a/xwalk_android_tests.gypi
+++ b/xwalk_android_tests.gypi
@@ -194,7 +194,6 @@
           ],
         },
       ],
-      'includes': [ '../build/java_apk.gypi' ],
     },
     {
       'target_name': 'xwalk_core_unittests',
@@ -487,7 +486,6 @@
           ],
         },
       ],
-      'includes': [ '../build/java_apk.gypi' ],
     },
     {
       'target_name': 'xwalk_runtime_client_embedded_shell_apk_java',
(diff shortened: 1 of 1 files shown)


mrunalk:
@rakuco @lincsoon PTAL at this simple fix.

rakuco:
Can you elaborate on how these includes are redundant? As far as I can see, those targets are not being directly `include`d by others so without that include their respective APKs should stop being generated.

xwalk_android_tests.gypi
```diff
@@ -194,7 +194,6 @@
           ],
         },
       ],
```
mrunalk: @rakuco for example this include is redundant as _xwalk_core_test_apk_ depends on _xwalk_core_shell_apk_java_ which in turn depends on _xwalk_core_shell_apk_ and _xwalk_core_shell_apk_ includes _java_apk.gypi_

rakuco: Yeah, and inclusions are not transitive: the fact that `xwalk_core_shell_apk` includes `java_apk.gypi` only means that it will generate an APK; the fact that this target is a dependency of `xwalk_core_test_apk` only means that it will be built before `xwalk_core_test_apk`, not that it will automatically make it create an APK too.

mrunalk: Ahh ok. I just verified the output folder and it didn't create those apk's so you are right. :)
I will close this PR for now. Atleast it's good that I didn't commit this patch directly without reviewing. Thanks for reviewing.

PR #3612, opened 2016-03-18
build: Disable use_sysroot by default.

rakuco:
This is a no-op in M48, but starting with M49 `use_sysroot` is enabled
by default on Linux. This means the build will use a Debian Wheezy
chroot to build and use as few system dependencies as possible.

It is a nice idea, but we cannot use it with Crosswalk at the moment
because we have an additional dependency on libnotify that is not part
of the Wheezy chroot shipped by Chromium.

BUG=XWALK-5806

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -29,8 +29,20 @@
     # check for a variable set in a condition block below, so we need to put
     # variables such as |enable_widevine| into an inner scope to avoid having
     # the conditions block here be at the same level as one from another file.
+    # Lastly, we have some extra conditional blocks (like for |use_sysroot|)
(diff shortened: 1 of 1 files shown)


rakuco:
@mrunalk @darktears @halton @PeterWangIntel

mrunalk:
LGTM, but shouldn't you be pushing this to next instead of master as the bug is for M49?

rakuco:
I like to keep the actual M49 rebasing patch as small as possible: since it contains dozens of squashed commits it's really annoying to do `git bisect` or `git blame` and stumble upon a huge rebasing commit that changes a ton of things.

Since this change also applies to M48 just fine, landing it before makes history tracking easier.

PR #3616, opened 2016-03-20
[Android] New API: Load URL with specified headers

lincsoon:
Added new interface for XWalkView: load URL with specified HTTP headers.

BUG=XWALK-4921


pine613:
:+1:

PR #3620, opened 2016-03-25
[Android] Copy carboard.jar to xwalk core library.

wuhengzhi:
This patch is to copy carboard.jar to xwalk core library instead of
merging it to xwalk_core_library_java.jar.
Next step app-tools could use carboard.jar to generate apk.

BUG=XWALK-6597


rakuco:
I've posted a comment about the general issue this PR is fixing in the ticket itself.

wuhengzhi:
If there is general method to fix this issue would be better.

wuhengzhi:
@rakuco PTAL, thanks.

lincsoon:
@rakuco  Is this PR still available?

rakuco:
The bug still needs to be fixed in master and crosswalk-19, but probably in a different way, I just haven't had time to go back to this issue.

PR #3621, opened 2016-03-28
[Android] Enable pull-to-refresh

axinging:
[Android] Enable pull-to-refresh
With this enabled, when user pull from the page top, page will be reloaded.
This can be disabled by flag --disable-pull-to-refresh-effect.
Pull-to-refresh differs from edge effect. For the edge effect, it is from
android and it "performs the graphical effect used at the edges of scrollable
widgets when the user scrolls beyond the content bounds in 2D space.".
Detailed documents is here:
http://developer.android.com/reference/android/widget/EdgeEffect.html.
As to the pull-to-refresh, it is from google chrome. It means reload the page
when user vertically overscrolling content at page top.

BUG=XWALK-6277

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
@@ -1,0 +1,185 @@
+// Copyright 2015 The Chromium Authors. All rights reserved.
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import android.content.Context;
+import android.view.ViewGroup.LayoutParams;
+
+import org.chromium.base.TraceEvent;
+import org.chromium.base.annotations.JNINamespace;
+import org.chromium.base.metrics.RecordUserAction;
+import org.chromium.content.browser.ContentViewCore;
+import org.chromium.content.browser.OverscrollRefreshHandler;
+import org.chromium.third_party.android.swiperefresh.SwipeRefreshLayout;
+
+import org.xwalk.core.internal.R;
+
+/**
+ * An overscroll handler implemented in terms a modified version of the Android
+ * compat library's SwipeRefreshLayout effect.
+ */
+@JNINamespace("xwalk")
+public class SwipeRefreshHandler implements OverscrollRefreshHandler {
+    // Synthetic delay between the {@link #didStopRefreshing()} signal and the
+    // call to stop the refresh animation.
+    private static final int STOP_REFRESH_ANIMATION_DELAY_MS = 500;
+
+    // Max allowed duration of the refresh animation after a refresh signal,
+    // guarding against cases where the page reload fails or takes too long.
+    private static final int MAX_REFRESH_ANIMATION_DURATION_MS = 7500;
+
+    // The modified AppCompat version of the refresh effect, handling all core
+    // logic, rendering and animation.
+    private final SwipeRefreshLayout mSwipeRefreshLayout;
+
+    // The ContentViewCore with which the handler is associated. The handler
+    // will set/unset itself as the default OverscrollRefreshHandler as the
+    // association changes.
+    private ContentViewCore mContentViewCore;
+
+    // Async runnable for ending the refresh animation after the page first
+    // loads a frame. This is used to provide a reasonable minimum animation time.
+    private Runnable mStopRefreshingRunnable;
+
+    // Accessibility utterance used to indicate refresh activation.
+    private String mAccessibilityRefreshString;
+
+    /**
+     * Simple constructor to use when creating an OverscrollRefresh instance from code.
+     *
+     * @param context The associated context.
+     */
+    public SwipeRefreshHandler(Context context) {
+        mSwipeRefreshLayout = new SwipeRefreshLayout(context);
+        mSwipeRefreshLayout.setLayoutParams(
+                new LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
+        mSwipeRefreshLayout.setColorSchemeResources(R.color.light_active_color);
+        // SwipeRefreshLayout.LARGE layouts appear broken on JellyBean.
+        mSwipeRefreshLayout.setSize(SwipeRefreshLayout.DEFAULT);
+        mSwipeRefreshLayout.setEnabled(false);
+    }
+
+    /**
+     * Pair the effect with a given ContentViewCore instance. If that instance is null,
+     * the effect will be disabled.
+     * @param contentViewCore The associated ContentViewCore instance.
+     */
+    public void setContentViewCore(final ContentViewCore contentViewCore) {
+        if (mContentViewCore == contentViewCore) return;
+
+        if (mContentViewCore != null) {
+            setEnabled(false);
+            cancelStopRefreshingRunnable();
+            mSwipeRefreshLayout.setOnRefreshListener(null);
+            mContentViewCore.setOverscrollRefreshHandler(null);
+        }
+
+        mContentViewCore = contentViewCore;
+
+        if (mContentViewCore == null) return;
+
+        setEnabled(true);
+        mSwipeRefreshLayout.setOnRefreshListener(new SwipeRefreshLayout.OnRefreshListener() {
+            @Override
+            public void onRefresh() {
+                cancelStopRefreshingRunnable();
+                mSwipeRefreshLayout.postDelayed(
+                        getStopRefreshingRunnable(), MAX_REFRESH_ANIMATION_DURATION_MS);
+                if (mAccessibilityRefreshString == null) {
+                    int resId = R.string.accessibility_swipe_refresh;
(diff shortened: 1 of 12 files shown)


axinging:
@lincsoon @fujunwei @xzhan96  @wuhengzhi

lincsoon:
LGTM

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
```diff
@@ -0,0 +1,185 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import android.content.Context;
+import android.view.ViewGroup.LayoutParams;
+
+import org.chromium.base.TraceEvent;
+import org.chromium.base.annotations.JNINamespace;
```
wuhengzhi: alpha sort.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -689,6 +694,26 @@ public void onGetFullscreenFlagFromManifest(boolean enterFullscreen) {
             }
         }
     }
+    /**
```
wuhengzhi: Need a blank line before this line.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkWebContentsDelegateAdapter.java
```diff
@@ -91,6 +91,12 @@ public boolean addMessageToConsole(int level, String message, int lineNumber,
     }
 
     @Override
+    public void showRepostFormWarningDialog() {
+        if (mXWalkContentsClient == null) return;
```
wuhengzhi: `if (mXWalkContentsClient != null) mXWalkContentsClient.resetSwipeRefreshHandler();` is better?

axinging: It seems both way exist on chromium android.

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
axinging: To align with origin copy chrome/android/java/src/org/chromium/chrome/browser/SwipeRefreshHandler.java, will change as little as possible.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
axinging: Done, thx!

rakuco:
> With this enabled, when user pull from the page top, page will be reloaded.

Are you sure this isn't already the case? I remember seeing a number of people complaining about this and actually asking how to turn it off.

axinging:
@rakuco, :), I was once confused by this too.  
There are two effects here: edge-effect, refresh-effect.  Currently  edge-effect is default on, while refresh-effect is missing. Before this, I always took edge-effect as refresh effect.So in previous threads, for customer wants to disable these, we always suggest they add flag: xwalk --disable-pull-to-refresh-effect --disable-overscroll-edge-effect. 

But indeed they are different. 

I am not sure if customers also mistakenly  regards edge effect as refresh effect.

As my observation and QA's,  pull to refresh is not enabled currently. This is why this commit happens.

rakuco:
I see. Can you explain what the difference between edge effect and refresh effect is (it would also be good to have it in the commit message)?

axinging:
@rakuco . I think no necessary. For Android developer, they are family with the edge effect, and also can be easily googled,  documents is here: http://developer.android.com/reference/android/widget/EdgeEffect.html.

As to the refresh, it is also from google chrome.

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
rakuco: "T" comes before "a" anyway.

axinging: You mean need to change it any way? For current, I think keep alignment with upstream is the better option.
"To align with origin copy chrome/android/java/src/org/chromium/chrome/browser/SwipeRefreshHandler.java, will change as little as possible."

axinging:
@rakuco

axinging:
@rakuco  again

rakuco:
This change is huge, I'll try to take a look tomorrow

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
```diff
@@ -0,0 +1,185 @@
...
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import android.content.Context;
+import android.view.ViewGroup.LayoutParams;
+
+import org.chromium.base.annotations.JNINamespace;
+import org.chromium.base.metrics.RecordUserAction;
+import org.chromium.base.TraceEvent;
```
rakuco: I meant the exact opposite of what you did with my previous comment. Sorting is case-sensitive; "T" (ASCII character 84) comes before "a" (ASCII character 97).

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
```diff
@@ -0,0 +1,185 @@
...
+import org.chromium.content.browser.OverscrollRefreshHandler;
+import org.chromium.third_party.android.swiperefresh.SwipeRefreshLayout;
+
+import org.xwalk.core.internal.R;
+
+/**
+ * An overscroll handler implemented in terms a modified version of the Android
+ * compat library's SwipeRefreshLayout effect.
+ */
+@JNINamespace("content")
```
rakuco: This doesn't look right.

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
```diff
@@ -0,0 +1,185 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import android.content.Context;
+import android.view.ViewGroup.LayoutParams;
+
+import org.chromium.base.annotations.JNINamespace;
+import org.chromium.base.metrics.RecordUserAction;
```
rakuco: I'm not sure it's worth duplicating even the tracings and metrics since they don't provide anything useful to us.

rakuco:
The implementation looks mostly right. What worries me is having to copy `SwipeRefreshHandler.java` to Crosswalk. This is the kind of thing that causes problems in the future since the implementations start diverging and we end up with an unfixable mess of old code with local workarounds.

It would be great if you or someone else from your team started an effort to move the original `SwipeRefreshHandler.java` to either `components/` or `content/`.

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
axinging: So， the original pull is correct? https://github.com/axinging/crosswalk/commit/3f21e93cf6487994379c1af917b35edfd1fac22d

Again, I hope we do not change this file too much. This is just a copy.

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
axinging: It's best to keep it  aligned with the original copy. I do not think it is good to introduce any new diff.

axinging:
@rakuco, good idea, it's good if we can move this to content or components.  I will investigate and negotiate with upstream if they are willing to move this to content or components.

BTW, now that this is not api related, how about enable it at crosswalk first?

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
rakuco: Yes, I was agreeing with you last week.

runtime/android/core_internal/src/org/xwalk/core/internal/SwipeRefreshHandler.java
rakuco: I agree with this rule of thumb, but in this case it's just wrong. We're not in the `content` namespace and this is not `src/content`. Like all other `JNINamespace` annotations in Crosswalk, it should use the "xwalk" namespace or something derived from it.

rakuco:
My only remaining objection is to the `JNINamespace` annotation as I described in my inline comment.

Meanwhile, please create a ticket about handling `SwipeRefreshHandler` upstream and make sure someone is taking care of it.

axinging:
Thanks @rakuco. Filed a bug here: https://bugs.chromium.org/p/chromium/issues/detail?id=603356.
It seems components has no place for a string, so I suggest put it under content.  
Also update the JNI namespace.

rakuco:
lgtm, thanks.

PR #3625, opened 2016-03-29
[Android] Enable onOverScrolled API in XWalkView. 

shaoboyan:
This PR is companied by [PR 330#](https://github.com/crosswalk-project/chromium-crosswalk/pull/330) in chromium-crosswalk.

BUG=XWALK-4871
BUG=XWALK-4894

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
@@ -247,6 +247,11 @@ class XWalkContentsClientBridge extends XWalkContentsClient
     }
 
     @Override
+    public void onOverScrolled(int scrollX, int scrollY, boolean clampedX, boolean clampedY) {
+        mXWalkView.onOverScrolledDelegate(scrollX, scrollY, clampedX, clampedY);
+    }
+
+    @Override
     public XWalkWebResourceResponseInternal shouldInterceptRequest(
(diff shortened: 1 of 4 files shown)


shaoboyan:
@axinging @lincsoon PTAL.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
```diff
@@ -247,6 +247,11 @@ public void onProgressChanged(int progress) {
     }
 
     @Override
+    public void onOverScrolled(int scrollX, int scrollY, boolean clampedX, boolean clampedY) {
+        mXWalkView.onOverScrolledDelegate(scrollX, scrollY, clampedX, clampedY);
+    }
+
+    @Override
```
axinging: I didn't understand why do it here and XWalkContentView.java.

axinging:
Thx shaobo, LGTM.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
shaoboyan: For the event is bottom up. I nned to call onOverScrolledDelegate.
And for the WebView onOverScrolled API, it will be called whenever a scroll event happens. so it is not only binding for overscroll event. To keep the API behavior the same as WebView. I add onOverScrolledDelegate call in onScrollChanged.

PR #3626, opened 2016-03-29
Disable viewport on desktop platforms

pozdnyakov:
Viewport is not supported on desktop platforms in Chromium, there
are multiple issues with it which won't be solved (see crbug.com/591326)

BUG=XWALK-6514


pozdnyakov:
@kenchris @darktears PTAL

kenchris:
fine with me. Bokan wanted to enable it but it doesnt seem to happen

rakuco:
@pozdnyakov, are you going to merge this?

PR #3627, opened 2016-03-29
Use std::move instead of scoped_ptr::Pass().

rakuco:
Follow-up to 44b2e8a ("Add a popup window to exit full screen mode"). In
the upcoming Chromium M49 it is not possible to use
`scoped_ptr::Pass()`. Prepare beforehand by using `std::move()` instead.


pozdnyakov:
lgtm

PR #3635, opened 2016-04-01
Add a NOLINT to avoid a build/header_guard cpplint warning.

rakuco:
`xwalk_common_message_generator.h` does not have include guards, and
according to a comment there this seems to be on purpose.

Add a `NOLINT` entry to it to avoid cpplint a build/header_guard error.

RELATED BUG=XWALK-1853


darktears:
LGTM

PR #3639, opened 2016-04-04
PRESUBMIT: Enable check for tabs in .grd files.

rakuco:
Enable the check after fixing the existing violations:

    ** Presubmit Warnings **
    Found a tab character in:
    ***************
    runtime/resources/xwalk_resources.grd:97
    runtime/resources/xwalk_resources.grd:98
    ***************


darktears:
LGTM

PR #3640, opened 2016-04-04
PRESUBMIT: Enable checks for CRLF line endings.

rakuco:
And fix the only existing violation:

    ** Presubmit Warnings **
    Are you sure that you want these files to contain Windows style line endings?
    runtime/browser/runtime_platform_util_win.cc


darktears:
LGTM

PR #3641, opened 2016-04-05
application: Remove unused JavaScript file.

rakuco:
`main.js` stopped being used with commit 7514f134 ('Drop "Main document"
support'), so we can remove it and get rid of an unlicensed file.


pozdnyakov:
lgtm

PR #3642, opened 2016-04-06
[Android] Refine the implementation of shouldOverrideUrlLoading

xzhan96:
Current implementation of shouldOverrideUrlLoading has confliction
with the original design intent. In native side, handle
shouldOverrideUrlLoading should via a sync IPC instead of current
scheme, which is only used to post onPageStarted in Android WebView.

And then, the test case also need to make corresponding changes,
for example, loading "file:///android_asset" and iframe will not
trigger shouldOverrideUrlLoading.

See CreateThrottlesForNavigation in xwalk_content_browser_client.cc
and HandleNavigation in xwalk_content_renderer_client.cc for more
details.

BUG=XWALK-6630

### DIFF
--- a/runtime/browser/android/xwalk_contents_client_bridge.cc
+++ b/runtime/browser/android/xwalk_contents_client_bridge.cc
@@ -285,6 +285,20 @@ void XWalkContentsClientBridge::OnWebLayoutPageScaleFactorChanged(
       env, obj.obj(), page_scale_factor);
 }
 
+bool XWalkContentsClientBridge::ShouldOverrideUrlLoading(
+    const base::string16& url,
+    bool has_user_gesture,
+    bool is_redirect,
+    bool is_main_frame) {
+  JNIEnv* env = AttachCurrentThread();
+  ScopedJavaLocalRef<jobject> obj = java_ref_.get(env);
+  if (obj.is_null())
+    return false;
+  ScopedJavaLocalRef<jstring> jurl = ConvertUTF16ToJavaString(env, url);
+  return Java_XWalkContentsClientBridge_shouldOverrideUrlLoading(
+      env, obj.obj(), jurl.obj(), has_user_gesture, is_redirect, is_main_frame);
+}
+
 void XWalkContentsClientBridge::ConfirmJsResult(JNIEnv* env,
                                                 jobject,
                                                 int id,
--- a/runtime/browser/xwalk_render_message_filter.cc
+++ b/runtime/browser/xwalk_render_message_filter.cc
@@ -1,3 +1,4 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
 // Copyright (c) 2014 Intel Corporation. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
@@ -5,12 +6,15 @@
 #include "xwalk/runtime/browser/xwalk_render_message_filter.h"
 
 #if defined(OS_ANDROID)
+#include "xwalk/runtime/browser/android/xwalk_contents_client_bridge_base.h"
 #include "xwalk/runtime/browser/android/xwalk_contents_io_thread_client.h"
 #include "xwalk/runtime/common/android/xwalk_render_view_messages.h"
 #endif
 
 #include "xwalk/runtime/common/xwalk_common_messages.h"
 #include "xwalk/runtime/browser/runtime_platform_util.h"
+
+using content::BrowserThread;
 
 namespace xwalk {
 
@@ -32,6 +36,8 @@ bool XWalkRenderMessageFilter::OnMessageReceived(
     IPC_MESSAGE_HANDLER(ViewMsg_OpenLinkExternal, OnOpenLinkExternal)
 #if defined(OS_ANDROID)
     IPC_MESSAGE_HANDLER(XWalkViewHostMsg_SubFrameCreated, OnSubFrameCreated)
+    IPC_MESSAGE_HANDLER(XWalkViewHostMsg_ShouldOverrideUrlLoading,
+                        OnShouldOverrideUrlLoading)
 #endif
     IPC_MESSAGE_UNHANDLED(handled = false)
   IPC_END_MESSAGE_MAP()
@@ -49,6 +55,26 @@ void XWalkRenderMessageFilter::OnSubFrameCreated(
     int parent_render_frame_id, int child_render_frame_id) {
   XWalkContentsIoThreadClient::SubFrameCreated(
       process_id_, parent_render_frame_id, child_render_frame_id);
+}
+
+void XWalkRenderMessageFilter::OnShouldOverrideUrlLoading(
+    int render_frame_id,
+    const base::string16& url,
+    bool has_user_gesture,
+    bool is_redirect,
+    bool is_main_frame,
+    bool* ignore_navigation) {
+  DCHECK_CURRENTLY_ON(BrowserThread::UI);
+  *ignore_navigation = false;
+  XWalkContentsClientBridgeBase* client =
+      XWalkContentsClientBridgeBase::FromRenderFrameID(process_id_, render_frame_id);
+  if (client) {
+    *ignore_navigation = client->ShouldOverrideUrlLoading(
+        url, has_user_gesture, is_redirect, is_main_frame);
+  } else {
+    LOG(WARNING) << "Failed to find the associated render view host for url: "
+                 << url;
+  }
 }
 #endif
 
--- a/runtime/browser/xwalk_render_message_filter.h
+++ b/runtime/browser/xwalk_render_message_filter.h
@@ -22,6 +22,12 @@ class XWalkRenderMessageFilter : public content::BrowserMessageFilter {
   void OnOpenLinkExternal(const GURL& url);
 #if defined(OS_ANDROID)
   void OnSubFrameCreated(int parent_render_frame_id, int child_render_frame_id);
+  void OnShouldOverrideUrlLoading(int routing_id,
+                                  const base::string16& url,
+                                  bool has_user_gesture,
+                                  bool is_redirect,
+                                  bool is_main_frame,
+                                  bool* ignore_navigation);
 #endif
   ~XWalkRenderMessageFilter() override {}
 
--- a/runtime/common/android/xwalk_render_view_messages.h
+++ b/runtime/common/android/xwalk_render_view_messages.h
@@ -115,6 +115,20 @@ IPC_MESSAGE_ROUTED0(XWalkViewHostMsg_PictureUpdated) // NOLINT(*)
 IPC_MESSAGE_ROUTED1(XWalkViewHostMsg_DidActivateAcceleratedCompositing, // NOLINT(*)
                     int /* input_handler_id */)
 
+// Sent immediately before a top level navigation is initiated within Blink.
+// There are some exclusions, the most important ones are it is not sent
+// when creating a popup window, and not sent for application initiated
+// navigations. See XWalkContentRendererClient::HandleNavigation for all
+// cornercases. This is sent before updating the NavigationController state
+// or creating a URLRequest for the main frame resource.
+IPC_SYNC_MESSAGE_CONTROL5_1(XWalkViewHostMsg_ShouldOverrideUrlLoading, // NOLINT(*)
+                            int /* render_frame_id id */,
+                            base::string16 /* in - url */,
+                            bool /* in - has_user_gesture */,
+                            bool /* in - is_redirect */,
+                            bool /* in - is_main_frame */,
+                            bool /* out - result */)
+
 // Sent when a subframe is created.
 IPC_MESSAGE_CONTROL2(XWalkViewHostMsg_SubFrameCreated, // NOLINT(*)
                      int, /* parent_render_frame_id */
--- a/runtime/renderer/xwalk_content_renderer_client.cc
+++ b/runtime/renderer/xwalk_content_renderer_client.cc
@@ -1,3 +1,4 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
 // Copyright (c) 2013 Intel Corporation. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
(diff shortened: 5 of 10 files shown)


xzhan96:
@lincsoon @axinging @fujunwei @wuhengzhi please help to review, thanks！

runtime/browser/android/xwalk_contents_client_bridge.cc
```diff
@@ -285,6 +285,19 @@ void XWalkContentsClientBridge::OnWebLayoutPageScaleFactorChanged(
       env, obj.obj(), page_scale_factor);
 }
 
+bool XWalkContentsClientBridge::ShouldOverrideUrlLoading(const base::string16& url,
+                                                      bool has_user_gesture,
```
wuhengzhi: indent.

runtime/renderer/xwalk_content_renderer_client.h
```diff
@@ -62,6 +62,14 @@ class XWalkContentRendererClient
                        GURL* new_url) override;
 
   void AddKeySystems(std::vector<media::KeySystemInfo>* key_systems) override;
+  bool HandleNavigation(content::RenderFrame* render_frame,
```
wuhengzhi: This function should be wrapped with `#ifdef OS_ANDROID  #endif`, so build failed on Linux.

runtime/common/android/xwalk_render_view_messages.h
```diff
@@ -115,6 +115,20 @@ IPC_MESSAGE_ROUTED0(XWalkViewHostMsg_PictureUpdated) // NOLINT(*)
 IPC_MESSAGE_ROUTED1(XWalkViewHostMsg_DidActivateAcceleratedCompositing, // NOLINT(*)
                     int /* input_handler_id */)
 
+// Sent immediately before a top level navigation is initiated within Blink.
+// There are some exlusions, the most important ones are it is not sent
```
wuhengzhi: exclusions?

xzhan96: so nice.. thanks!

runtime/browser/xwalk_render_message_filter.cc
```diff
@@ -32,24 +35,46 @@ bool XWalkRenderMessageFilter::OnMessageReceived(
...
+    *ignore_navigation = client->ShouldOverrideUrlLoading(
+        url, has_user_gesture, is_redirect, is_main_frame);
+  } else {
+    LOG(WARNING) << "Failed to find the associated render view host for url: "
+                 << url;
+  }
+}
 #endif
 
+void XWalkRenderMessageFilter::OnOpenLinkExternal(const GURL& url) {
```
axinging: How about put this at its original place? Is there any necessary to do this here?

axinging:
LGTM

runtime/browser/xwalk_render_message_filter.cc
xzhan96: @axinging this is to correct the format error before (not need to define OnSubFrameCreated as private, this is from my previous patch ) and align with function declaration in xwalk_render_message_filter.h

xzhan96:
@wuhengzhi @XiaosongWei @lincsoon please help to review, thanks

rakuco:
Does this cause any conflicts with #3651?

wuhengzhi:
@rakuco `conflicts` means duplicated process or just the code conflicts?
For duplicated process, there is no conflicts since https://github.com/crosswalk-project/crosswalk/pull/3651 is just for custom urls.
For code conflicts, I implemented https://github.com/crosswalk-project/crosswalk/pull/3651 base on this PR, then submitted https://github.com/crosswalk-project/crosswalk/pull/3651.

xzhan96:
@wuhengzhi we modify the same file "XWalkContentsClientBridge.java", I think this is the only one potential conflict, right?
I checked #3651 , it has no code logic conflict with this PR. @rakuco

runtime/browser/xwalk_render_message_filter.cc
XiaosongWei: I prefer to keep those unchanged, if you want to correct the format, it would be better to submit a separate PR for that.

runtime/browser/xwalk_render_message_filter.h
```diff
@@ -17,12 +17,17 @@ class XWalkRenderMessageFilter : public content::BrowserMessageFilter {
   explicit XWalkRenderMessageFilter(int process_id);
 #endif
   bool OnMessageReceived(const IPC::Message& message) override;
-
- private:
-  void OnOpenLinkExternal(const GURL& url);
 #if defined(OS_ANDROID)
   void OnSubFrameCreated(int parent_render_frame_id, int child_render_frame_id);
+  void OnShouldOverrideUrlLoading(int routing_id,
```
XiaosongWei: it would be nice to keep these two OnXXX as private as they are both your private implementation.

xzhan96: nice, thanks

runtime/browser/xwalk_render_message_filter.cc
xzhan96: ok

XiaosongWei:
LGTM

runtime/browser/xwalk_render_message_filter.cc
```diff
@@ -50,6 +55,26 @@ void XWalkRenderMessageFilter::OnSubFrameCreated(
   XWalkContentsIoThreadClient::SubFrameCreated(
       process_id_, parent_render_frame_id, child_render_frame_id);
 }
+
+void XWalkRenderMessageFilter::OnShouldOverrideUrlLoading(
```
rakuco: This function was basically copied from Chromium, yet there's no copyright attribution.

runtime/renderer/xwalk_content_renderer_client.cc
```diff
@@ -130,6 +130,62 @@ void XWalkContentRendererClient::RenderThreadStarted() {
 #endif
 }
 
+#if defined(OS_ANDROID)
+bool XWalkContentRendererClient::HandleNavigation(
```
rakuco: Ditto for the copyright of this function.

rakuco:
After looking at previous commits related to this:
- https://github.com/crosswalk-project/crosswalk/commit/09567ffc444f84dad3ecc553d52d35b26948cb53
- https://github.com/crosswalk-project/crosswalk/commit/823d34c9385d14899c5bf333c0967a66d8c2d174
- https://github.com/crosswalk-project/crosswalk/commit/f846cdbfb9103a6653fea6d9e3c4c8da3810fc9e

I have to say I got even more confused than I was before. Should people use `shouldOverrideUrlLoading` or `XWalkNavigationHandler`? Why do we have both in the first place and how do they interact? Were we allowing people to intercept all kinds of requests, including those that fail the checks in `XWalkContentRendererClient::HandleNavigation()`? Restricting what can be intercepted looks like a sane change, but we have to announce this behavior change very loudly.

runtime/renderer/xwalk_content_renderer_client.cc
xzhan96: thanks

xzhan96:
@rakuco 
Q: Should people use shouldOverrideUrlLoading or XWalkNavigationHandler? 
A: My understanding is that people will not use XWalkNavigationHandler, because it is a default navigation handler for handling platform-specific schemes, like "wtai://", as descripted  in 09567ff and it will not be exposed to external, please check XWalkNavigationHandlerImpl.java.
However, as a command API in both WebViewClient and XWalkResourceClient, shouldOverrideUrlLoading is frequently used.

Q: Why do we have both in the first place and how do they interact? 
A: I think they will not interact, because they are for different url schemes, shouldOverrideUrlLoading  only intercept resource from internet.

Q: Were we allowing people to intercept all kinds of requests, including those that fail the checks in XWalkContentRendererClient::HandleNavigation()?
A: Why need to intercept all kinds of requests? In WebView, local resource loading like "file:///android_asset" will not trigger shouldOverrideUrlLoading, and for HTTP schemes, only top-level navigations can be overridden, I think it's reasonable.  But In XWalkView, those special cases are not taken into account, which will confuse developer.
I am not sure my answer is clear, feel free to let me know if any question.

xzhan96:
@rakuco is this patch ok?

xzhan96:
ping @rakuco

runtime/renderer/xwalk_content_renderer_client.cc
```diff
@@ -130,6 +131,62 @@ void XWalkContentRendererClient::RenderThreadStarted() {
...
+  if (application_initiated && !is_redirect)
+    return false;
+
+  bool is_main_frame = !frame->parent();
+  const GURL& gurl = request.url();
+  // For HTTP schemes, only top-level navigations can be overridden. Similarly,
+  // WebView Classic lets app override only top level about:blank navigations.
+  // So we filter out non-top about:blank navigations here.
+  if (!is_main_frame &&
+      (gurl.SchemeIs(url::kHttpScheme) || gurl.SchemeIs(url::kHttpsScheme) ||
```
rakuco: Are all those conditions being tested in `ShouldOverrideUrlLoadingTest.java`?

test/android/core/javatests/src/org/xwalk/core/xwview/test/ShouldOverrideUrlLoadingTest.java
```diff
@@ -417,7 +417,7 @@ public Boolean call() throws Exception {
         // Since the targetURL was loaded from the test server it means all processing related
         // to dispatching a shouldOverrideUrlLoading callback had finished and checking the call
         // is stable.
-        assertEquals(shouldOverrideUrlLoadingCallCount + 1,
+        assertEquals(shouldOverrideUrlLoadingCallCount,
```
rakuco: Shouldn't this be just "0"?

test/android/core/javatests/src/org/xwalk/core/xwview/test/ShouldOverrideUrlLoadingTest.java
```diff
@@ -469,7 +471,7 @@ public Boolean call() throws Exception {
             }
         });
 
-        assertEquals(1, mShouldOverrideUrlLoadingHelper.getCallCount());
+        assertEquals(shouldOverrideUrlLoadingCallCount, mShouldOverrideUrlLoadingHelper.getCallCount());
```
rakuco: I think it's easier to understand what's being tested here if you just use `0` instead. This is 0, right?

test/android/core/javatests/src/org/xwalk/core/xwview/test/ShouldOverrideUrlLoadingTest.java
```diff
@@ -488,20 +490,22 @@ private void doTestCalledOnRedirect(TestWebServer webServer,
...
         loadUrlSync(redirectUrl);
 
-        mShouldOverrideUrlLoadingHelper.waitForCallback(directLoadCallCount, 2);
+        mShouldOverrideUrlLoadingHelper.waitForCallback(directLoadCallCount, 1);
         assertEquals(redirectTarget,
                 mShouldOverrideUrlLoadingHelper.getShouldOverrideUrlLoadingUrl());
 
         int indirectLoadCallCount = mShouldOverrideUrlLoadingHelper.getCallCount();
         loadUrlSync(pageWithLinkToRedirectUrl);
+        assertEquals(indirectLoadCallCount, mShouldOverrideUrlLoadingHelper.getCallCount());
```
rakuco: Ditto here -- I'm assuming the value should be 1 here just like it is above, right?

rakuco:
>> Should people use shouldOverrideUrlLoading or XWalkNavigationHandler?

> My understanding is that people will not use XWalkNavigationHandler, because it is a default navigation handler for handling platform-specific schemes, like "wtai://", as descripted in 09567ff and it will not be exposed to external, please check XWalkNavigationHandlerImpl.java.
> However, as a command API in both WebViewClient and XWalkResourceClient, shouldOverrideUrlLoading is frequently used.

>> Why do we have both in the first place and how do they interact?

> I think they will not interact, because they are for different url schemes, shouldOverrideUrlLoading only intercept resource from internet.

From reading what you wrote it sounds like `XWalkNavigationHandler` is an internal version of `shouldOverrideUrlLoading` for certain schemes that users cannot override. If that's the case I feel they could be sharing more code. Someone should investigate (not for this pull request, of course) if that's really the case or even if `XWalkNavigationHandler` could be removed.

>> Were we allowing people to intercept all kinds of requests, including those that fail the checks in XWalkContentRendererClient::HandleNavigation()?

> Why need to intercept all kinds of requests? In WebView, local resource loading like "file:///android_asset" will not trigger shouldOverrideUrlLoading, and for HTTP schemes, only top-level navigations can be overridden, I think it's reasonable. But In XWalkView, those special cases are not taken into account, which will confuse developer.

Er, I was expecting a "yes" or "no" in response to my question :-) Let me rephrase: were we allowing API users to do too much before? Were they able to override the handling of non-GET requests, and requests to non-http(s) schemes, meaning they could essentially implement their own XWalkNavigationHandlers and bypass our checks? If so, that does not look great.

runtime/renderer/xwalk_content_renderer_client.cc
xzhan96: The same as WebView, there are testCalledForIframeHttpNavigations, and the test for "about:blank".

test/android/core/javatests/src/org/xwalk/core/xwview/test/ShouldOverrideUrlLoadingTest.java
xzhan96: yes, but do you think it's better to use shouldOverrideUrlLoadingCallCount which is get from mShouldOverrideUrlLoadingHelper.getCallCount() before getting request from web server instead of using "0" directly?

test/android/core/javatests/src/org/xwalk/core/xwview/test/ShouldOverrideUrlLoadingTest.java
xzhan96: yes, it's "0". I referred to WebView. I think this style means the test just compare the function call count before and after getting request from web server.

xzhan96:
> From reading what you wrote it sounds like XWalkNavigationHandler is an internal version of shouldOverrideUrlLoading for certain schemes that users cannot override. If that's the case I feel they could be sharing more code. Someone should investigate (not for this pull request, of course) if that's really the case or even if XWalkNavigationHandler could be removed.
Agree with you, I don't know clearly about #09567ff, it should be investigated before more changes.

> Er, I was expecting a "yes" or "no" in response to my question :-) Let me rephrase: were we allowing API users to do too much before? Were they able to override the handling of non-GET requests, and requests to non-http(s) schemes, meaning they could essentially implement their own XWalkNavigationHandlers and bypass our checks? If so, that does not look great.
sorry..:-)
Yes, API users do too much before, but it has conflict with the native design. Current implementation doesn't allow user to override XWalkNavigationHandlers to bypass our check.

test/android/core/javatests/src/org/xwalk/core/xwview/test/ShouldOverrideUrlLoadingTest.java
rakuco: Personally I prefer using a number here since it's easier to just look at a zero and say "hmm, so there shouldn't be any calls here". But I won't object if you prefer to keep it this way.

rakuco:
I think we're close to an agreement. Since neither of us understands this code very well, I have to ask: have you tested that the `HandleNavigation()` code is called _before_ the resource throttling code that triggers `InterceptNavigationDelegateImpl.shouldIgnoreNavigation()` and `XWalkNavigationHandler.handleNavigation()`? I don't think our tests currently cover this, and I'd like to double-check this with you since following the logic in Chromium is very confusing.

xzhan96:
The basic logic of this part is:
`HandleNavigation()` in aw_content_renderer_client.cc or xwalk_content_renderer_client.cc is called by  `decidePolicyForNavigation()`  in render_frame_impl.cc, and `decidePolicyForNavigation()` is called by FrameLoaderClientImpl.cpp. All above actions are **before** creating DocumentLoader, which will do the resource request. And **`InterceptNavigationDelegateImpl.shouldIgnoreNavigation()`** will be called by `StartRequest()` in resource_loader.cpp, which is **after** creating DocumentLoader, and as my previous, it is **only used for posting onPageStarted event in Android WebView**.

> have you tested that the HandleNavigation() code is called before the resource throttling code 

**So all the tests** in ShouldOverrideUrlLoadingTest.java is to test the logic path of `HandleNavigation()`, which will call `shouldOverrideUrlLoading()` in WebViewClient.java or XWalkResourceClient.java finally. 

I didn't see the test case for 09567ff, maybe we need to check why, and investigate `XWalkNavigationHandler.handleNavigation()` that if they could be sharing more code, but as your comments, it not this patch should do.

rakuco:
>> have you tested that the HandleNavigation() code is called before the resource throttling code
> 
> So all the tests in ShouldOverrideUrlLoadingTest.java is to test the logic path of HandleNavigation(), which will call shouldOverrideUrlLoading() in WebViewClient.java or XWalkResourceClient.java finally.

OK. What I meant with my question is that there doesn't seem to be any tests that implement both `shouldOverrideUrlLoading()` and `XWalkNavigationHandler.handleNavigation()` and verify that only the former is called instead of the latter. Food for thought for your future efforts.

xzhan96:
> OK. What I meant with my question is that there doesn't seem to be any tests that implement both shouldOverrideUrlLoading() and XWalkNavigationHandler.handleNavigation() and verify that only the former is called instead of the latter. 

You are right.
Actually, In my previous comments "I didn't see the test case for 09567ff, maybe we need to check why", I meant there is no test for XWalkNavigationHandler.handleNavigation(). Anyway, we will check it and discuss in weekly meeting.

xzhan96:
> OK. What I meant with my question is that there doesn't seem to be any tests that implement both shouldOverrideUrlLoading() and XWalkNavigationHandler.handleNavigation() and verify that only the former is called instead of the latter. 

You are right.
Actually, In my previous comments "I didn't see the test case for 09567ff, maybe we need to check why", I meant there is no test for XWalkNavigationHandler.handleNavigation(). Anyway, we will check it and discuss in weekly meeting.

PR #3643, opened 2016-04-06
[Android] Fix about setTextZoom does not work with CSS3 columns.

wuhengzhi:
With following CSS styles:

    <div id="div1" style="-webkit-column-width:1000px; height: 1000px;">

setTextZoom(200) does not work with the expected scale behavior.
Implement the layout algorithm related APIs to verify if the zoom should
be force enabled.
Add test case for layout algorithm.

BUG=XWALK-5929


runtime/android/core_internal/src/org/xwalk/core/internal/XWalkSettingsInternal.java
```diff
@@ -95,6 +106,8 @@
     private boolean mSpatialNavigationEnabled = true;
     private boolean mQuirksModeEnabled = false;
 
+    private LayoutAlgorithmInternal mLayoutAlgorithm = LayoutAlgorithmInternal.NARROW_COLUMNS;
```
FaultException: @wuhengzhi Are you positive this default is correct? Android's WebView says that NORMAL is the recommended mode, NARROW_COLUMS is deprecated, and I didn't change the mode in my test project.

wuhengzhi: Yes, from [WebSettings.LayoutAlgorithm](http://developer.android.com/reference/android/webkit/WebSettings.LayoutAlgorithm.html), `NORMAL` is recommended.
From [AwSettings.java#L55](https://github.com/crosswalk-project/chromium-crosswalk/blob/master/android_webview/java/src/org/chromium/android_webview/AwSettings.java#L55), the default value is still `NARROW_COLUMNS`.
Thanks.

wuhengzhi:
@axinging @lincsoon @fujunwei @xzhan96 PTAL, thanks.

rakuco:
lgtm
anyone else?

xzhan96:
lgtm

PR #3644, opened 2016-04-07
sysapps: Use real Promises instead of a polyfill.

rakuco:
Back when the sysapps code was being written, V8's and Chromium's
Promises implementation was still under development, and only reached
Crosswalk months later.

The sysapps code resorted to using a very basic polyfill at the
time (which has a different syntax and is not as asynchronous as a
proper implementation), and nobody updated the code once we were able to
use Chromium's Promises implementation.

This change finally does it: instead requiring the "sysapps_promise"
module, we now just use Chromium's Promise object, reducing the
dependency on the third-party `common_promise_api.js`.


sysapps/common/common_api.js
```diff
@@ -80,13 +71,18 @@ var BindingObjectPrototype = function() {
...
 
-  function addMethodWithPromise(name, Promise) {
+  function addMethodWithPromise(name) {
     Object.defineProperty(this, name, {
       value: function() {
-        var promise_instance = new Promise();
         var args = Array.prototype.slice.call(arguments);
-        this._postMessage(name, args, wrapPromiseAsCallback(promise_instance));
-        return promise_instance;
+        return new Promise(function(resolve, reject) {
```
kenchris: No need to create functions (resolve, reject) => {  } will do

sysapps/common/common_api.js
```diff
@@ -80,13 +71,18 @@ var BindingObjectPrototype = function() {
...
+          this._postMessage(name, args, function(data, error) {
+            if (error)
+              reject(error);
```
kenchris: It error a real Error object? otherwise new Error("...name...", { message: ... } )

sysapps/common/common_api.js
```diff
@@ -80,13 +71,18 @@ var BindingObjectPrototype = function() {
...
+            else
+              resolve(data);
+          });
+        }.bind(this));
```
kenchris: bind is not needed when using =>

sysapps/common/common_api.js
kenchris: return new Promise((resolve, reject) => error ? reject(error) : resolve(data));

should probably work. If you need to add { } then you need to add 'return'  as well

sysapps/common/common_api.js
```diff
@@ -80,13 +71,18 @@ var BindingObjectPrototype = function() {
...
     Object.defineProperty(this, name, {
       value: function() {
-        var promise_instance = new Promise();
         var args = Array.prototype.slice.call(arguments);
-        this._postMessage(name, args, wrapPromiseAsCallback(promise_instance));
-        return promise_instance;
+        return new Promise((resolve, reject) => {
+          this._postMessage(name, args, function(data, error_message) {
+            if (error_message)
+              reject(new Error(error_message));
```
kenchris: Did you test that? I believe the message has to be second argument, but I might be mistaken that that is just for ErrorEvent

kenchris: You are right... my mistake, this is fine

rakuco: I did, see how I had to change `common_api_browsertest.html`. I'm using `Error` as described here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

rakuco: so lgtm? :)

sysapps/common/common_api.js
```diff
@@ -80,13 +71,18 @@ var BindingObjectPrototype = function() {
...
-  function addMethodWithPromise(name, Promise) {
+  function addMethodWithPromise(name) {
     Object.defineProperty(this, name, {
       value: function() {
-        var promise_instance = new Promise();
         var args = Array.prototype.slice.call(arguments);
-        this._postMessage(name, args, wrapPromiseAsCallback(promise_instance));
-        return promise_instance;
+        return new Promise((resolve, reject) => {
+          this._postMessage(name, args, function(data, error_message) {
```
kenchris: You can do this easier. Does postMessage ALWAYS return?

then

```
this._postMessage(name, args, (data, error) => error ? Promise.reject(new Error(error)) : Promise.resolve(data));
```

kenchris: Hmm, might not work if the postMessage is async... sorry for the noice

kenchris: I would still make the function(data, err) using (data, err) => instead

kenchris: so lgtm with that change

sysapps/device_capabilities/device_capabilities_api.js
```diff
@@ -12,8 +12,6 @@ var v8tools = requireNative('v8tools');
 var common = requireNative('sysapps_common');
 common.setupSysAppsCommon(internal, v8tools);
 
-var Promise = requireNative('sysapps_promise').Promise;
```
kenchris: Are you removing the polyfill as well?

kenchris:
lgtm, will you remove the polyfill it self in another commit?

sysapps/device_capabilities/device_capabilities_api.js
rakuco: There are other parts of the code using it (like the native filesystem code), so I'm removing its uses one module at a time until I'm able to nuke it from orbit.

rakuco:
There are other parts of the code using it (like the native filesystem code), so I'm removing its uses one module at a time until I'm able to nuke it from orbit.

PR #3645, opened 2016-04-07
native filesystem extension: Clean up the code.

rakuco:
Oh boy, where do I even start? This extension has been pretty much
unmaintained since 2014, does not really implement any actual spec and
was using the `Promises` polyfill from the sysapps code. It was the
latter that prompted me to take a look at it, and this is what I came up
with.

All the commits but the last one are small groundwork, and the last
change is where the big cleanup is actually performed. Quoting this last
commit's commit message:

Make the code more readable and trim all the unnecessary fat from both the
C++ and the JS sides to the extent that keeping backwards compatibility
allows us (i.e. synchronous functions remain synchronous, async functions
that use callbacks instead of Promises continue doing so etc).

- Remove all the pseudo-Promises silliness: manually keeping a list of
  Promises (which were just using the sysapps polyfill instead of actual
  Promises) and manipulating it whenever a message was passed or
  received was just bonkers.

- Remove all the manual message passing code. Creating a `msg`
  dictionary with a loosely-defined syntax does not make sense, we can
  just leverage the support code from `xwalk_internal_api.js` that takes
  care of keeping track of each message, correctly forwarding all
  parameters and invoking an optional callback when a response is
  returned from the C++ side.

- Consequently, this also allows us to register a specific handler for
  the newly-introduced "requestNativeFileSystem" message instead of
  implementing a catch-all `HandleMessage()` function.

- Another consequence of this change is that we now return a simple
  error string when an error occurred instead of packing the whole thing
  into a JSON string with a lot of unnecessary additional data. This
  error message gets turned into an `Error` object. It is the only
  somewhat backwards-incompatible change in this commit. I believe the
  benefits of doing error handling more sanely outweigh the cons here.

- Add an IDL file describing the parameters we pass to
  `requestNativeFileSystem()` to avoid doing a lot of manual argument
  parsing in `OnRequestNativeFileSystem` (previously `HandleMessage`).

- Replace all the `FileSystemChecker` complication with a single,
  internal function.

With this change, the Promises polyfill in the `sysapps/` directory no
longer has any users, and can be safely removed in another commit.

### DIFF
--- a/experimental/native_file_system/native_file_system.idl
+++ b/experimental/native_file_system/native_file_system.idl
@@ -1,0 +1,15 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Native FileSystem API.
+// This does not correspond to any spec (or even the JS API we export in
+// native_file_system_api.js). This IDL is used to make it easier to parse
+// messages sent to the extension and prepare responses to the JavaScript side.
+namespace native_file_system {
+    callback RequestCallback = void (DOMString filesystem_id, DOMString error);
+
+    interface Functions {
+        static void requestNativeFileSystem(DOMString path, RequestCallback callback);
+    };
+};
--- a/experimental/native_file_system/native_file_system_api.js
+++ b/experimental/native_file_system/native_file_system_api.js
@@ -40,10 +40,6 @@ var requestNativeFileSystem = function(path, success, error) {
   postMessage(msg, get_file_system_id_success, error);
 }
 
-var getDirectoryList = function() {
-  extension.internal.sendSyncMessage("get");
-}
-
 var getRealPath = function(virtual_root) {
   var _msg = {
     cmd : "getRealPath",
@@ -55,7 +51,6 @@ var getRealPath = function(virtual_root) {
 NativeFileSystem.prototype = new Object();
 NativeFileSystem.prototype.constructor = NativeFileSystem;
 NativeFileSystem.prototype.requestNativeFileSystem = requestNativeFileSystem;
-NativeFileSystem.prototype.getDirectoryList = getDirectoryList;
 NativeFileSystem.prototype.getRealPath = getRealPath;
 
 exports = new NativeFileSystem();
--- a/experimental/native_file_system/native_file_system_api.js
+++ b/experimental/native_file_system/native_file_system_api.js
@@ -2,80 +2,38 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-var _promises = {};
-var _next_promise_id = 0;
+let isolated_fs = requireNative("isolated_file_system");
+let internal = requireNative("internal");
+internal.setupInternalExtension(extension);
 
-var Promise = requireNative('sysapps_promise').Promise;
-var IsolatedFileSystem = requireNative('isolated_file_system');
-
-var postMessage = function(msg, success, error) {
-  var p = new Promise();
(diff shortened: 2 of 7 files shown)


experimental/native_file_system/native_file_system.idl
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// Native FileSystem API.
+// This does not correspond to any spec (or even the JS api we export in
```
kenchris: API

experimental/native_file_system/native_file_system_api.js
```diff
@@ -2,85 +2,38 @@
...
-
-function handlePromise(msgObj) {
-  if (msgObj.data.error) {
-    if (_isFunction(_promises[msgObj._promise_id].reject)) {
-      _promises[msgObj._promise_id].reject(msgObj.data);
-    }
-  } else {
-    if (_isFunction(_promises[msgObj._promise_id].fulfill)) {
-      _promises[msgObj._promise_id].fulfill(msgObj.data);
+  requestNativeFileSystem(path, success, error) {
```
kenchris: I wonder if we should return a promise as well... so people can choose to add arguments or use it as promise?

rakuco: That was my original intention, but that probably breaks the code for existing users that are not expecting a Promise to be returned.

experimental/native_file_system/native_file_system_api.js
```diff
@@ -2,85 +2,38 @@
...
+      throw new TypeError("Wrong parameters passed to requestNativeFileSystem.");
     }
+    error = error || (_ => {});
+
+    internal.postMessage(
+      "requestNativeFileSystem", [path],
+      (filesystem_id, error_message) => {
+        if (error_message)
+          error(new Error(error_message));
+        else
```
kenchris: where is getDirectoryList?

experimental/native_file_system/native_file_system_api_browsertest.html
```diff
@@ -41,9 +54,9 @@
                 writer.write(blob);
                 runNextTest();
               },
-              function(e) {reportFail(JSON.stringify(e))});
+              function(e) {reportFail(e)});
             },
-          function(e) {reportFail(JSON.stringify(e))});
+          function(e) {reportFail(e.message)});
```
kenchris: why e.message here and e elsewhere

experimental/native_file_system/native_file_system_api.js
rakuco: It was removed, see the commit _"native filesystem: Remove getDirectoryList()"_. It never worked because the C++ side was never implemented.

experimental/native_file_system/native_file_system_api_browsertest.html
rakuco: Oops.

experimental/native_file_system/native_file_system_api.js
```diff
@@ -55,7 +51,6 @@ var getRealPath = function(virtual_root) {
 NativeFileSystem.prototype = new Object();
 NativeFileSystem.prototype.constructor = NativeFileSystem;
 NativeFileSystem.prototype.requestNativeFileSystem = requestNativeFileSystem;
-NativeFileSystem.prototype.getDirectoryList = getDirectoryList;
```
kenchris: Ah ok got it! gone it is!

kenchris:
lgtm

PR #3646, opened 2016-04-07
sysapps: Remove Promises polyfill.

rakuco:
Now that all users of the Promises polyfill have been converted to
native Promises (which have shipped with Chromium since M32), remove our
polyfill, which was not very complete. It also had a different license
and was not in our `third_party/` directory.


kenchris:
rs=me

PR #3648, opened 2016-04-08
[Android] Rewrite generate_app_packaging_tool.py.

rakuco:
Simplify the script a lot by not trying to deduce if an entry is a file
or a directory. Instead, keep two separate lists with directories and
files we want to copy to the `xwalk_app_template` directory.

Also stop hardcoding some paths and let gyp pass all the directory
information to the script.


darktears:
LGTM

PR #3650, opened 2016-04-11
[Android] Update warning in setAcceptFileSchemeCookies's API documentation

xzhan96:
Since M46, there is a change in upstream to match the comments
in CookieManager "Note that calls to this method will have no
effect if made after a WebView or CookieManager instance has
been created". For XWalkCookieManager, the function can't be
defined static for reflection limitation, so the comment is
not correct.

BUG=XWALK-6274

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
@@ -123,7 +123,7 @@ public class XWalkCookieManagerInternal {
      * of cookie data can take place.
      * <p>
      * Note that calls to this method will have no effect if made after a
-     * WebView or CookieManager instance has been created.
+     * CookieManager instance has done any real work.
      * @param accept Whether accept cookies for file scheme URLs
      * @since 5.0
      */


xzhan96:
@axinging @wuhengzhi @lincsoon please help to review and merge it, thanks

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
```diff
@@ -123,7 +123,7 @@ public boolean allowFileSchemeCookies() {
      * of cookie data can take place.
      * <p>
      * Note that calls to this method will have no effect if made after a
-     * WebView or CookieManager instance has been created.
+     * CookieManager instance have done any real work.
```
rakuco: have -> has

rakuco:
> Update the notic of setAcceptFileSchemeCookies

"notic" is not a word. How about _"Update warning in setAcceptFileSchemeCookies's API documentation"_?

> defined static for relection limitation, so the comment is

"relection" is not a word. Did you mean "reflection"?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
xzhan96: thanks a lot @rakuco

rakuco:
thanks!

PR #3651, opened 2016-04-12
[Android] Implement the feature about custom url scheme.

wuhengzhi:
This patch is to support the custom url scheme, e.g. "whatsapp://",
"intent://".
For "intent://", if the related app, e.g. Barcode Scanner, was
installed, then start the activity to launch this scanner, if it was
not installed, XWalkView will load the fallback url if it was specified.
For "whatsapp://", if whatsapp was installed, start the activity to
launch it, if it was not installed, then try to install it from market
according to the package name. If the package name is null, will report
"ERR_UNKNOWN_URL_SCHEME".

BUG=XWALK-6690

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
@@ -96,8 +96,15 @@ class XWalkContentsClientBridge extends XWalkContentsClient
                      mNavigationHandler.handleNavigation(navigationParams);
 
             if (!ignoreNavigation) {
-                // Post a message to UI thread to notify the page is starting to load.
-                mContentsClient.getCallbackHelper().postOnPageStarted(url);
+                // Check whether the fallback url is existed for scheme: intent://.
+                final String fallbackUrl = mNavigationHandler.getFallbackUrl();
+                if (fallbackUrl != null) {
+                    mNavigationHandler.resetFallbackUrl();
+                    mXWalkView.load(fallbackUrl, null);
+                } else {
+                    // Post a message to UI thread to notify the page is starting to load.
+                    mContentsClient.getCallbackHelper().postOnPageStarted(url);
+                }
             }
 
             return ignoreNavigation;
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/UrlUtilities.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/UrlUtilities.java
@@ -1,0 +1,164 @@
+// Copyright 2013 The Chromium Authors. All rights reserved.
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+import android.content.ComponentName;
+import android.content.Context;
+import android.content.Intent;
+import android.content.IntentFilter;
+import android.content.pm.ActivityInfo;
+import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
+import android.text.TextUtils;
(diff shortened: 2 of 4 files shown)


rakuco:
What worries me with this patch is that we're duplicating a lot of code and logic from Chromium (even specific hacks for youtube.com, for example). It will inevitably bitrot as we rebase on top of newer releases in the future.

Isn't there a way to duplicate less code? If this isn't urgent, perhaps work upstream to disentable this from the Chromium code and move this to a lower part of the stack?

wuhengzhi:
Most of the codes come from `src/chrome`, move the implementation to content will duplicate less code. There are some special processes for `chrome`, e.g. chrome scheme, chrome switch, I will discuss with upstream if the move is possible.

rakuco:
Can you please create a ticket about the upstream activities to keep us informed of the discussions?

wuhengzhi:
Please look at https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/QtsNyPNlfXI, there is no response now.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
```diff
@@ -97,8 +97,15 @@ public boolean shouldIgnoreNavigation(NavigationParams navigationParams) {
                       mNavigationHandler.handleNavigation(navigationParams));
 
             if (!ignoreNavigation) {
-                // Post a message to UI thread to notify the page is starting to load.
-                mContentsClient.getCallbackHelper().postOnPageStarted(url);
+                // Check whether the fallback url is existed for scheme: intent://.
+                if (mNavigationHandler.getFallbackUrl() != null) {
+                    final String fallbackUrl = mNavigationHandler.getFallbackUrl();
```
lincsoon: No need to call `getFallbackUrl` twice. And why `final`?

lincsoon:
Before we figure out a better solution, I suggest we copy the code from upstream firstly, but mark them clearly for further rebase or modification

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContentsClientBridge.java
wuhengzhi: ```final``` is better, please look at above code: ```final String url = navigationParams.url;```

wuhengzhi:
@rakuco Same as https://bugs.chromium.org/p/chromium/issues/detail?id=603356, it seems that upstream don't wanna split the code.

wuhengzhi:
For ```UrlUtilities.java```, the code was ported from [UrlUtilities.java](https://github.com/crosswalk-project/chromium-crosswalk/blob/master/chrome/android/java/src/org/chromium/chrome/browser/util/UrlUtilities.java), [IntentUtils.java](https://github.com/crosswalk-project/chromium-crosswalk/blob/master/chrome/android/java/src/org/chromium/chrome/browser/util/IntentUtils.java) and 
[ExternalNavigationDelegateImpl.java](https://github.com/crosswalk-project/chromium-crosswalk/blob/master/chrome/android/java/src/org/chromium/chrome/browser/externalnav/ExternalNavigationDelegateImpl.java).
For ```XWalkNavigationHandlerImpl.java: shouldOverrideUrlLoadingInternal()```, the code was ported from [ExternalNavigationHandler.java](https://github.com/crosswalk-project/chromium-crosswalk/blob/master/chrome/android/java/src/org/chromium/chrome/browser/externalnav/ExternalNavigationHandler.java)

lincsoon:
Could you mark it clearly on the source file? If a entire method is from upstream, we should take care to rebase in the future instead of reviewing it now. And we must can know about it from the soure file.

wuhengzhi:
@lincsoon Added the ported urls in code as comments, thanks.

PR #3652, opened 2016-04-12
Use a standard std::string() constructor instead of base::EmptyString.

rakuco:
The former is less costly than than the latter, which requires
thread-safe access and a singleton. We just need an empty `std::string()`
after all.


pozdnyakov:
lgtm

PR #3653, opened 2016-04-12
crosswalk-18: Use a standard std::string() constructor instead of base::EmptyString.

rakuco:
The former is less costly than than the latter, which requires
thread-safe access and a singleton. We just need an empty std::string()
after all.

(cherry picked from commit c1e8610f455a18574697cf73e0f902b0b8e776e3)


pozdnyakov:
lgtm

PR #3654, opened 2016-04-12
[Android] Fix xwalk_app_runtime_java.jar's location in the app template.

rakuco:
Fix a regression introduced by 7209eee ("[Android] Rewrite
generate_app_packaging_tool.py") which ended up putting
`lib/xwalk_app_runtime_java.jar` into the top-level app_template
directory. It must be under `template/`.

BUG=XWALK-6744


r0b5t4:
rs=me

PR #3656, opened 2016-04-13
[Android] Remove OS_ANDROID checks from an Android-specific file.

rakuco:
`xwalk_presentation_service_delegate_android.cc`, as the name says, is
only built on Android so there is no reason to have checks for the
`OS_ANDROID` macro there.

While here, fix a typo in a function name ("Andorid" -> "Android") and
adjust the coding style in surrounding regions.


rakuco:
@pozdnyakov @kenchris

kenchris:
lgtm :) how do you even find these things

PR #3657, opened 2016-04-14
[Android] Add notification while downloading is in progress and finished

xzhan96:
When open a link to download file in XWalkView, it's better the downloading
is visible and shows in the notifications while in progress and after
completion, so the downloaded file can be opened.

BUG=XWALK-6678


xzhan96:
@lincsoon @axinging @wuhengzhi @XiaosongWei @fujunwei please help review, thanks

rakuco:
I agree with the patch, but I don't think this solves XWALK-3108, only XWALK-6678.

XWALK-3108 ended up becoming a mixture of different issues reported by different people. Originally, it was only about the user's app supposedly freezing (hard to confirm, the original reporter is marked as inactive) and the error message shown in logcat. I'd like to treat it that way, and leave the "I wish that the users are notified when a download finishes" wishlist item to XWALK-6678. I don't think this patch gets rid of the logcat NOTIMPLEMENTED warning at all.

wuhengzhi:
LGTM, thanks.

XiaosongWei:
LGTM

rakuco:
thanks!

PR #3659, opened 2016-04-18
crosswalk-18: Build with enable_webvr=0 by default.

rakuco:
Disable WebVR support on Android by default, since it depends on
`third_party/cardboard-java` which in turn needs
`third_party/android_protobuf`.

The dependency on `android_protobuf` means `xwalk_core_library_java.jar`
bundle Protobuf and cause build conflicts for users depending on Protobuf
themselves (or using other libraries that pull it).

The WebVR code is still in a very experimental state, even more so in M48,
so disabling it altogether is the best solution to get Crosswalk 18 out of
the door. We will try to think of a better solution for Crosswalk 19 and
above.

RELATED BUG=XWALK-6597
BUG=XWALK-6746

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = 'd2f39946b24323c7d2ef510d1cb5451b3a30b4cc'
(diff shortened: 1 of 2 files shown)


rakuco:
ping @darktears

rakuco:
or @pozdnyakov or @kenchris

darktears:
LGTM

PR #3660, opened 2016-04-19
Remove fake external extension api

heke123:
Decide to add lite support in cordova plugin 1.6.x, so we don't need
this fake external extension api now.
Note that after cordova plugin upgrade into 1.7.0, we need to add
this api back.


heke123:
@halton @PeterWangIntel  Thanks.

halton:
@heke123 I think this change is reverting PR https://github.com/crosswalk-project/crosswalk/pull/3598? could you use reverting to make the two PRs them connected.

halton:
LGTM, will merge after trybot green.

heke123:
@halton @PeterWangIntel   changed into "git revert",  thanks:)

PR #3662, opened 2016-04-21
[Android] Implement getContentHeight API.

shaoboyan:
Implement getContentHeight API for sync invoking to get content height.

 BUG=XWALK-6521

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -380,6 +380,10 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
         return mContentsClientBridge.getXWalkWebChromeClient();
     }
 
+    public int getContentHeight() {
+        return (int)Math.ceil(mContentViewCore.getContentHeightCss());
+    }
+
     public void setXWalkClient(XWalkClient client) {
         if (mNativeContent == 0) return;
         mContentsClientBridge.setXWalkClient(client);
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -491,6 +491,18 @@ public class XWalkViewInternal extends android.widget.FrameLayout {
     }
 
     /**
+     * Get the content height of current web page/app.
+     * NOTE: When page finished loading, it's content size info may not be updated.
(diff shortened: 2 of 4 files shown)


shaoboyan:
@axinging @XiaosongWei @lincsoon  PTAL.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -380,6 +380,21 @@ public XWalkWebChromeClient getXWalkWebChromeClient() {
         return mContentsClientBridge.getXWalkWebChromeClient();
     }
 
+    public int getContentHeight() {
+        /**
+         * When page finished loading, it's content size info may not be updated.
```
lincsoon: `its`?
I think `\\` is enough for the comment inside of the method

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -380,6 +380,21 @@ public XWalkWebChromeClient getXWalkWebChromeClient() {
...
+         * Wait for a while for info updating.
+         */
+        if ((int)Math.ceil(mContentViewCore.getContentHeightCss()) == 0) {
+            try {
+                Thread.sleep(50);
```
lincsoon: I'm not sure the background of this API, but hang on the UI thread?

shaoboyan: I consider this for some time. and I think it is not a good idea to hold UI thread on our side. I'll modify this and add comment to remind users.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -491,6 +491,18 @@ public String getUrl() {
     }
 
     /**
+     * Get the content height of current web page/app.
+     * NOTE: When page finished loading, it's content size info may not be updated.
```
axinging: So does this mean when finished loading, its OK to call this? Otherwise may get a invalid data?

shaoboyan: No, this means when finished loading. it is possible that the value is still not be updated. but if it happens, just a little while(i test waiting 50 ms). it will be updated.

shaoboyan:
@lincsoon @axinging PTAL. Now the API have risks that couldn't return the correct value when the page loaded immediately.

PR #3663, opened 2016-04-21
[Android] Implement overScrollBy API.

shaoboyan:
Implement overScrollBy API for XWalkView.

BUG=XWALK-4894

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright (C) 2006 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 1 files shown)


shaoboyan:
This API could not be tested automatly.

shaoboyan:
@axinging @XiaosongWei  PTAL.

shaoboyan:
@axinging @XiaosongWei  PTAL.

XiaosongWei:
one question: the overScrollBy is View's protected method, now you get it promoted to public, what if developer extends XWalkView and override the protected overScrollBy? isn't it a compile error in public to protected downgrade ?

> 
The access modifier (§6.6) of an overriding or hiding method must provide at least as much access as the overridden or hidden method, as follows:
>
If the overridden or hidden method is public, then the overriding or hiding method must be public; otherwise, a compile-time error occurs.
If the overridden or hidden method is protected, then the overriding or hiding method must be protected or public; otherwise, a compile-time error occurs.
If the overridden or hidden method has default (package) access, then the overriding or hiding method must not be private; otherwise, a compile-time error occurs.

shaoboyan:
@XiaosongWei Yes. You are at point. Thx for the remind.I'll update it.

shaoboyan:
@XiaosongWei @lincsoon It is a pity that current @XWALKAPI annotation could only expose public methond to XWalkView.java. Since WebView have lots of protected method. shall we extend @XWALKAPI annotation for exposing protected method ? 
P.S: Although lots of protected method is invoked bottom-up but there are also some methods are invoked top-down, like: compute* APIs .

lincsoon:
About the problem of the access modifier, I think it's okay to use public temporarily if it's the only solution to implement feature. Because the developers who encounters a compilation error can fix it by using public too, although it's kind of strange. Then I'll create a ticket to improve the `@XWalkAPI` header.

But please shaobo give more comments on this patch:
* Why is it necessary to override the `overscrollBy` method and why other solution is not feasible?
* Which code is copied from the View class directly? And you should add copyright at the top of the source file too.

shaoboyan:
@lincsoon done.

PR #3668, opened 2016-04-22
PRESUBMIT: Stop relying on cpplint.FileInfo.RepositoryName().

rakuco:
This protects us from changes in depot_tools that cause problems for us.

For example, https://codereview.chromium.org/1897153003 breaks our use
case in the Buildbot slaves: Crosswalk is checked out inside the
chromium-crosswalk directory, which is checked out inside our Buildbot
git repository. That CL makes FileInfo.RepositoryName() find the
outermost git repository, which is the Buildbot one. In the end, the
suggested include guards will be wrong and look like this:

    XWALK_SLAVE_CROSSWALK_LINUX_BUILD_SRC_XWALK_RUNTIME_BROWSER_XWALK_PRESENTATION_SERVICE_HELPER_ANDROID_H_

Since we know our repository layout, we can take a shortcut and just
return the path we want with the right prefix by doing some string
manipulation.

### DIFF
--- a/PRESUBMIT.py
+++ b/PRESUBMIT.py
@@ -26,9 +26,13 @@ _LICENSE_HEADER_RE = (
 def _CheckChangeLintsClean(input_api, output_api):
   class PrefixedFileInfo(cpplint.FileInfo):
     def RepositoryName(self):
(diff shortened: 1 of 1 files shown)


pozdnyakov:
rs=me

PR #3669, opened 2016-04-25
Stop using std::wstring in our code.

rakuco:
Chromium's coding style (as well as Google's) says using `wstring` is
frowned upon for multiple reasons. There is even a check in Chromium's
main `PRESUBMIT.py` that prevents new `wstring`s from being added in
crosswalk-platform code.

The only use was in code copied from content_shell in
2013 (`RuntimeDownloadManagerDelegate::ChooseDownloadPath()`), which
could be rewritten using only `base::FilePath`s and Chromium's
`OPENFILENAME` wrapper class. In any case, Crosswalk on Windows still
does not support downloads so the change cannot break anything (even
though I have verified it does build).

This change is another step towards enabling more PRESUBMIT checks for
our code.

### DIFF
--- a/runtime/browser/runtime_download_manager_delegate.cc
+++ b/runtime/browser/runtime_download_manager_delegate.cc
@@ -29,6 +29,10 @@
 #include "base/nix/xdg_util.h"
 #endif
 
+#if defined(OS_WIN)
+#include "ui/base/win/open_file_name_win.h"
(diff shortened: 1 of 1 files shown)


rakuco:
ping @darktears @pozdnyakov

darktears:
LGTM

PR #3670, opened 2016-04-26
[Android] Give host application a chance to take over MediaPlayer

fujunwei:
Give the host application a chance to take over the control when a new
media is about to be played in the current XWalkViewInternal

BUG=XWALK-6770


lincsoon:
@fujunwei What's the latest status?

PR #3671, opened 2016-04-26
[Android] Update comments of set/getCookie in XWalkCooikeManager

xzhan96:
Some customers complain that set/getCookie of XWalkCooikeManager is invalid
without adding scheme into url domain, which isn't required in Android WebView.

This is because CookieManager already use WebAddress to add scheme into the
url, so the comments of according XWalk APIs need to be updated.

BUG=XWALK-6766

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
@@ -38,6 +38,8 @@ public class XWalkCookieManagerInternal {
      * Set cookie for a given url. The old cookie with same host/path/name will
      * be removed. The new cookie will be added if it is not expired or it does
      * not have expiration which implies it is session cookie.
+     *
+     * Scheme[https|http] must be added as the prefix of url domain.
      * @param url The url which cookie is set for
      * @param value The value for set-cookie: in http response header
      * @since 5.0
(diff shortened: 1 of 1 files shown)


xzhan96:
@lincsoon @fujunwei @XiaosongWei @wuhengzhi @axinging please help to review

rakuco:
Isn't it possible to change our implementation to match the Android WebView instead?

xzhan96:
> Isn't it possible to change our implementation to match the Android WebView instead?

I've thought about it, maybe it's ok, but I am not sure the potential impact. And since it is easy to add a scheme before url domain, I commit this patch.
We can discuss it, and if it is necessary, then implement it.

rakuco:
It may be easy to add a scheme to the URL, but in this case I guess it's safe to say that this is not intuitive: it's not required by the Android WebView (which can lead to subtle problems in existing code), and new and existing users will only find out about this if they read a note in the API documentation. At the very least, I'd expect an error message to be shown to users if the URL is malformed (not sure if that's already done on the native side).
If this is a matter of taking `url` and doing some processing to it (it's just not clear what the Android WebView does to choose the scheme) I don't see why not do it instead.

lincsoon:
I agree with rakuco. Few people pay attention to the API description.

xzhan96:
ok, I will change xwalk implementation to match WebView.

PR #3674, opened 2016-04-27
[Android] Indicate that a malformed URL isn't accepted by XWalkCooikeManager

xzhan96:
Some customers complain long debug time to know that url without scheme is invalid
when using XWalkCooikeManager, which isn't required in Android WebView.

This is because CookieManager already use WebAddress to add scheme into the url.
For the different design, xwalk will not do the same implementation, but it's better
to throw an exception to user in this case.

BUG=XWALK-6766

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
@@ -4,6 +4,11 @@
 
 package org.xwalk.core.internal;
 
+import android.util.Log;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
 import org.chromium.base.annotations.JNINamespace;
 
 /**
@@ -14,6 +19,8 @@ import org.chromium.base.annotations.JNINamespace;
 @JNINamespace("xwalk")
 @XWalkAPI(createExternally = true)
 public class XWalkCookieManagerInternal {
+    private static final String TAG = "XWalkCookieManager";
+
     /**
      * Control whether cookie is enabled or disabled
      * @param accept TRUE if accept cookie
@@ -44,7 +51,11 @@ public class XWalkCookieManagerInternal {
      */
     @XWalkAPI
     public void setCookie(final String url, final String value) {
-        nativeSetCookie(url, value);
+        try {
+            nativeSetCookie(new URL(url).toString(), value);
+        } catch (MalformedURLException e) {
+            Log.e(TAG, "Not setting cookie due to invalid URL", e);
+        }
     }
 
     /**
@@ -56,9 +67,14 @@ public class XWalkCookieManagerInternal {
      */
     @XWalkAPI
     public String getCookie(final String url) {
-        String cookie = nativeGetCookie(url.toString());
-        // Return null if the string is empty to match legacy behavior
-        return cookie == null || cookie.trim().isEmpty() ? null : cookie;
+        try {
+            String cookie = nativeGetCookie(new URL(url).toString());
+            // Return null if the string is empty to match legacy behavior
+            return cookie == null || cookie.trim().isEmpty() ? null : cookie;
+        } catch (MalformedURLException e) {
+            Log.e(TAG, "Unable to get cookies due to invalid URL", e);
+            return null;
+        }
     }
 
     /**


xzhan96:
@rakuco @lincsoon @wuhengzhi please help to review, thanks

xzhan96:
[https://github.com/crosswalk-project/crosswalk/pull/3671](url)

xzhan96:
#3671

runtime/android/core_internal/src/org/xwalk/core/internal/WebAddress.java
```diff
@@ -0,0 +1,189 @@
...
+            /* path      */ "(\\/?[^#]*)?" +
+            /* anchor    */ ".*", Pattern.CASE_INSENSITIVE);
+
+    /** parses given uriString. */
+    public WebAddress(String address) throws ParseException {
+        if (address == null) {
+            throw new NullPointerException();
+        }
+
+        // android.util.Log.d(LOGTAG, "WebAddress: " + address);
```
wuhengzhi: remove it?

xzhan96:
@lincsoon @rakuco

rakuco:
Er, I was expecting some one-liners that passed `url` to a parser and raised an exception if the URL was malformed, not some mammoth patch including a parser and even a new exception type...

It looks like having bug-per-bug compatibility with the classic WebView requires too much code, I don't see the value in importing these new files into the tree.

We just need to make the behavior intuitive for our users: doesn't my suggestion work (ie. passing the `url` parameters to something like `java.net.URL` or whatever more specialized solution Android or Chromium has and failing very loudly if it does not work)?

xzhan96:
> We just need to make the behavior intuitive for our users: doesn't my suggestion work (ie. passing the url parameters to something like java.net.URL or whatever more specialized solution Android or Chromium has and failing very loudly if it does not work)?

This is what I am thinking, let me investigate again.

xzhan96:
@rakuco please review if this patch is you want, :）

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
```diff
@@ -14,6 +19,8 @@
 @JNINamespace("xwalk")
 @XWalkAPI(createExternally = true)
 public class XWalkCookieManagerInternal {
+    private static final String LOGTAG = "XWalkCookieManager";
```
lincsoon: Basically, we just use `TAG` in crosswalk's code

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
```diff
@@ -132,6 +148,16 @@ public void setAcceptFileSchemeCookies(boolean accept) {
...
+    private static String judgeUrl(String url) throws MalformedURLException {
+        URL newUrl = null;
+        try{
+            newUrl = new URL(url);
+        }catch (MalformedURLException e){
+            throw new MalformedURLException("Invalid URL!");
+        }
+        return newUrl.toString();
+    }
+
```
lincsoon: Personally, the following code would be simpler.

```
try {
    return new URL(url).toString();
} catch (...) {
    throw ...
}
return null;
```

lincsoon: If this method is only used by `getCookie` and just has this simple functionality, I don't see the necessity to make it a standalone method instead of adding code in `getCookie` directly.

xzhan96: nice, thanks

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
xzhan96: Sorry, why XWalkCookieManager is not crosswalk's code?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManagerInternal.java
```diff
@@ -132,6 +148,14 @@ public void setAcceptFileSchemeCookies(boolean accept) {
         nativeSetAcceptFileSchemeCookies(accept);
     }
 
+    private static String judgeUrl(String url) throws MalformedURLException {
+        try {
+            return new URL(url).toString();
+        } catch (MalformedURLException e){
+            throw new MalformedURLException("Invalid URL!");
+        }
+    }
```
wuhengzhi: >     private String judgeUrl(String url) throws MalformedURLException {
        return new URL(url).toString();
    }

Or this function could be removed.

xzhan96: Have you tried if it can be built successfully?

xzhan96:
@lincsoon @wuhengzhi @XiaosongWei please help to review

xzhan96:
@rakuco  please review if this patch is you want.

wuhengzhi:
LGTM except the top comment: `So we do the same implementation in XWalk to match the Android WebView.`, it should also be updated.

xzhan96:
> So we do the same implementation in XWalk to match the Android WebView., it should also be updated.

This is the original commit message, not latest one

lincsoon:
LGTM

wuhengzhi:
LGTM, thanks.

xzhan96:
@rakuco please help review and merge, thanks

xzhan96:
@rakuco

xzhan96:
@rakuco please review again

xzhan96:
> Er, I was expecting some one-liners that passed url to a parser and raised an exception if the URL was malformed, not some mammoth patch including a parser and even a new exception type...
It looks like having bug-per-bug compatibility with the classic WebView requires too much code, I don't see the value in importing these new files into the tree.
We just need to make the behavior intuitive for our users: doesn't my suggestion work (ie. passing the url parameters to something like java.net.URL or whatever more specialized solution Android or Chromium has and failing very loudly if it does not work)?

@rakuco please review if this patch is you want, :）

xzhan96:
@rakuco please review if this patch is you want, :）

> Er, I was expecting some one-liners that passed url to a parser and raised an exception if the URL was malformed, not some mammoth patch including a parser and even a new exception type...
It looks like having bug-per-bug compatibility with the classic WebView requires too much code, I don't see the value in importing these new files into the tree.
We just need to make the behavior intuitive for our users: doesn't my suggestion work (ie. passing the url parameters to something like java.net.URL or whatever more specialized solution Android or Chromium has and failing very loudly if it does not work)?

PR #3675, opened 2016-04-27
Add release notes

baleboy:
This PR adds the full Crosswalk release notes (up to Crosswalk 19) to the source tree. Once this is merged, the process to update the release notes going forward would be something like:

- Anybody who submits new features is welcome and encouraged to add a description to the release notes in the same PR that implements the feature
- Latest before branching to beta the release notes are updated to reflect the content up to that point (my responsibility)
- Any notable bug fixes or changes done in the beta branch are updated in the branched release notes

The above will be communicated on crosswalk-dev once this PR is merged.

This first submission is huge because it includes all the release notes starting from Crosswalk 1 (collecting them was a nice trip down Memory Lane). That partly explains why the content and style change quite a bit, as the format has evolved over time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,3 +82,18 @@ license.
   exceptions being the amount of space used for indentation (2 instead of 4)
   and the use of `MixedCase` functions and methods instead of
   `lower_case_ones`.
+  
+## Updating the release notes
+  
+* When submitting a new feature, the author is welcome and encouraged to add a 
+  short description to the release notes in the same PR that implements the
+  feature.
+  
+* Latest before branching to beta the release notes are updated to reflect the
+  content up to that point.
+  
+* Any notable bug fixes or changes done in the beta branch are updated in the 
+  branched release notes.
+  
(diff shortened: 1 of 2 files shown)


baleboy:
@rakuco @bspencer @darktears @lincsoon @ibelem PTAL

rakuco:
I suggest adding the process for updating the release notes you described to the file itself so people know when to edit it.

baleboy:
@rakuco unless I make it very short (and therefore not so useful) I feel that it would clutter the release notes with information needed only by part of the intended audience. I could put it at the end, but that means nobody will read it. What about documenting the process on the website?

rakuco:
OK then; we could also put it in `CONTRIBUTING.md` or elsewhere in the tree so it's easy to point people to the instructions, but any place is fine to me in the end.

baleboy:
Note that I added the date of promotion to stable next to each release.

baleboy:
@rakuco good point, I added a section to CONTRIBUTING.md. Sorry for taking the bots' time with these doc updates...

rakuco:
lgtm :)

PR #3676, opened 2016-04-27
[Windows] Fix notifications on Windows.

darktears:
Make sure to set the AppUserModelID correctly so that it matches
what app-tools is setting on the shortcut. This will fix the bug
where packaged app would not show notifications : this was due to
some very old code inside xwalk setting AppUserModelID to the
application_id. This code is not needed anymore because we're going
to set AppUserModelID to the application name being launched. This
patch will also fix the grouping in the notification center on builds
leading to Anniversary Update because instead of showing the app_id
it will show the actual name of the app.

BUG=XWALK-4990


darktears:
@pozdnyakov your review love is appreciated.

pozdnyakov:
lgtm

PR #3677, opened 2016-04-27
Fix ASSERTs and low quality assets on high dpi screens

darktears:
Right now we're shipping only assets with 100% scale factor. This
patch address the issue by making sure that we at least support
better assets when the scale factor is not 100%.

This will require a change in app-tools as well.


darktears:
@lincsoon can you have a look please?

@r0b5t4 I need a fix in app-tools see : https://github.com/darktears/crosswalk-app-tools/commit/21443043f963f35679bb901e001a63cac63b401f but how do we proceed because these files will start to show up after a given version. Anything else I missed?

r0b5t4:
@darktears we can add all .pak files just like we do for dll files already. Would that work? The code is right below where you patched it.

darktears:
@r0b5t4 yep I guess we can do that indeed.

darktears:
PR for app tool is there : https://github.com/crosswalk-project/crosswalk-app-tools/pull/132

darktears:
ping @lincsoon

lincsoon:
LGTM for the Android part

PR #3679, opened 2016-04-29
[Windows][Linux] Allow localhost for media capture APIs

darktears:
We just follow upstream and ease local debugging, regular
http is still disallowed, only https.

BUG=XWALK-6562

### DIFF
--- a/runtime/browser/media/media_capture_devices_dispatcher.cc
+++ b/runtime/browser/media/media_capture_devices_dispatcher.cc
@@ -15,6 +15,7 @@
 #include "content/public/browser/render_process_host.h"
 #include "content/public/common/media_stream_request.h"
 #include "grit/xwalk_resources.h"
+#include "net/base/net_util.h"
 #include "ui/base/l10n/l10n_util.h"
 #include "xwalk/application/browser/application.h"
 #include "xwalk/application/browser/application_service.h"
@@ -127,7 +128,10 @@ void XWalkMediaCaptureDevicesDispatcher::RequestPermissionToUser(
     const content::MediaResponseCallback& callback) {
   DCHECK(content::BrowserThread::CurrentlyOn(content::BrowserThread::UI));
 
-  if (request.security_origin.SchemeIs(url::kHttpScheme))
+  // This function may be called for a media request coming from
+  // from WebRTC/mediaDevices. These requests can't be made from HTTP.
+  if (request.security_origin.SchemeIs(url::kHttpScheme) &&
+      !net::IsLocalhost(request.security_origin.host()))
     callback.Run(content::MediaStreamDevices(),
                  content::MEDIA_DEVICE_PERMISSION_DENIED,
                  scoped_ptr<content::MediaStreamUI>());
--- a/runtime/browser/runtime.cc
+++ b/runtime/browser/runtime.cc
@@ -25,6 +25,7 @@
 #include "ui/base/resource/resource_bundle.h"
 #include "ui/gfx/image/image_skia.h"
 #include "ui/gfx/native_widget_types.h"
+#include "net/base/net_util.h"
 #include "xwalk/application/browser/application.h"
 #include "xwalk/application/browser/application_service.h"
 #include "xwalk/application/browser/application_system.h"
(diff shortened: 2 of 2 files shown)


darktears:
@huningxin @pozdnyakov

runtime/browser/runtime.cc
```diff
@@ -356,7 +356,11 @@ bool Runtime::CheckMediaAccessPermission(
 #else
   // This function may be called for a media request coming from
   // from WebRTC/mediaDevices. These requests can't be made from HTTP.
-  if (security_origin.SchemeIs(url::kHttpScheme))
+  std::string host = security_origin.host();
+  bool is_localhost = host.compare("localhost") ||
```
pozdnyakov: don't we have a place like "utils" to put a function there so that this code is not duplicated?

darktears:
@pozdnyakov new version PTAL

runtime/browser/xwalk_utils.cc
```diff
@@ -0,0 +1,16 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/runtime/browser/xwalk_utils.h"
+
+namespace xwalk_utils {
+
+bool IsLocalHost(const GURL& url) {
+  std::string host = url.host();
+  return host.compare("localhost") ||
```
pozdnyakov: think it should rather be `host.compare("localhost") == 0` and same for others.. http://www.cplusplus.com/reference/string/string/compare/ (looks like browser tests failure on Linux might be related)

runtime/browser/xwalk_utils.h
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef XWALK_RUNTIME_BROWSER_XWALK_UTILS_H_
+#define XWALK_RUNTIME_BROWSER_XWALK_UTILS_H_
+
+#include "url/gurl.h"
+
+namespace xwalk_utils {
```
pozdnyakov: maybe nested 'utiils' namespace is better?

darktears: Using the existing namespace that we have. I don't mind either way.

runtime/browser/xwalk_utils.cc
darktears: Nope, they are not related. @rakuco moved the bot to Portland and we experiencing flakiness on the test (though it was flaky before).

runtime/browser/media/media_capture_devices_dispatcher.cc
```diff
@@ -127,7 +128,10 @@ void XWalkMediaCaptureDevicesDispatcher::RequestPermissionToUser(
     const content::MediaResponseCallback& callback) {
   DCHECK(content::BrowserThread::CurrentlyOn(content::BrowserThread::UI));
 
-  if (request.security_origin.SchemeIs(url::kHttpScheme))
+  // This function may be called for a media request coming from
+  // from WebRTC/mediaDevices. These requests can't be made from HTTP.
+  if (request.security_origin.SchemeIs(url::kHttpScheme) &&
+      !utils::IsLocalHost(request.security_origin))
```
huningxin: Can we use `net::IsLocalhost`? (https://github.com/crosswalk-project/chromium-crosswalk/blob/master/net/base/net_util.h#L102)

darktears: Yes of course. I did see that.

darktears:
@pozdnyakov should be good this time, @huningxin had a great suggestion.

pozdnyakov:
lgtm code-wise, but browser test failure looks related.

rakuco:
Alexis, please resubmit for us to make sure: the Linux bots were sick due to some behind-the-scenes changes (XWALK-6856) and I've reverted them to their previous, working state.

darktears:
They can't be related, I doubt any of these tests is using getUserMedia() JS API. My patch is only about code path related to getUserMedia.

darktears:
All green merging.

huningxin:
That's great! LGTM.

PR #3520, opened 2016-02-01
Show specail options with the accept attribute

fujunwei:
The following code is allow the user to upload an image from stored files or by taking a photo using camera, but the Camcorder and Voice Recorder options should not be included.
```
<input id="file" type="file" name="file" accept="image/*">
```

BUG=XWALK-3750

PR #3682, opened 2016-05-04
[Android] Fix the wrong choice for input element.

wuhengzhi:
Show special option was fixed in https://github.com/crosswalk-project/crosswalk/pull/3520,
there is an exception that is the write external permission was not
added, then select the wrong option even if the accept type was correct.

BUG=XWALK-6430

(cherry picked from commit fc3b73e673340879757cc882ffe159bdfe33da4a)


wuhengzhi:
@rakuco Please take a look and merge, thanks.

PR #3684, opened 2016-05-04
[Linux]Fix the race condition in XWalkExtensionService

XiaosongWei:
XWalkExtensionServer got deleted in XWalkExtensionData by
XWalkExtensionService while the XWalkExtensionServer pointers are still
accessed by ExtensionServerMessageFilter in IPC thread.

Use WeakPtr to avoid such race condition.

BUG=XWALK-6856

### DIFF
--- a/extensions/browser/xwalk_extension_service.cc
+++ b/extensions/browser/xwalk_extension_service.cc
@@ -51,8 +51,8 @@ class ExtensionServerMessageFilter : public IPC::MessageFilter,
  public:
   ExtensionServerMessageFilter(
       scoped_refptr<base::SequencedTaskRunner> task_runner,
-      XWalkExtensionServer* extension_thread_server,
-      XWalkExtensionServer* ui_thread_server)
+      const base::WeakPtr<XWalkExtensionServer>& extension_thread_server,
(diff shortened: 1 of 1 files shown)


XiaosongWei:
@rakuco I am unable to reproduce the issue in my local environment and I just added log to check the suspected point, need your help to double confirm whether it works or not.

@darktears @baleboy Please take a look!

rakuco:
I didn't have time to test this today, but looking at the patch I don't see how it would fix that bug: you're just adding weak pointers to the code without checking for their validity in `OnGetExtensions` (and possibly other places).

XiaosongWei:
@rakuco the weak pointers validity check is done in ```OnMessageReceived``` before calling ```OnGetExtension``` and ```OnCreateInstance```

rakuco:
Unfortunately it looks like you did not try a debug build, which would have also helped you try to reproduce the issue in the first place.

```
[89561:89575:0510/190546:FATAL:weak_ptr.cc(26)] Check failed: sequence_checker_.CalledOnValidSequencedThread(). WeakPtrs must be checked on the same sequenced thread.
```

This assert prevents the tests from even running in debug mode.

rakuco:
see also: #3694

PR #3686, opened 2016-05-05
[Windows] Download support

darktears:
Implement download support for Windows as it was already almost finished.
I removed the code when not built with Aura because it's dead code
(Windows, Linux are built with Aura by default). I also put the code
to figure out the Download path in the common place with other key paths.

BUG=XWALK-4969

### DIFF
--- a/runtime/common/xwalk_paths.cc
+++ b/runtime/common/xwalk_paths.cc
@@ -12,6 +12,9 @@
 
 #if defined(OS_WIN)
 #include "base/base_paths_win.h"
+#include "base/win/scoped_co_mem.h"
+#include <knownfolders.h>
+#include <shlobj.h>
 #elif defined(OS_LINUX)
 #include "base/environment.h"
 #include "base/nix/xdg_util.h"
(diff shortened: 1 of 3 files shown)


darktears:
@pozdnyakov @rakuco some review love.

runtime/common/xwalk_paths.cc
```diff
@@ -116,6 +119,53 @@ bool GetInternalPluginsDirectory(base::FilePath* result) {
...
+    return false;
+  }
+  *result = base::FilePath(path_buf);
+  return true;
+}
+
+bool GetUserDownloadDirectory(base::FilePath* result) {
+  typedef HRESULT(WINAPI *GetKnownFolderPath)(
+      REFKNOWNFOLDERID, DWORD, HANDLE, PWSTR*);
+  GetKnownFolderPath f = reinterpret_cast<GetKnownFolderPath>(
```
pozdnyakov: f?  :) maybe a more descriptive name?

rakuco:
looks sane to me

PR #3690, opened 2016-05-06
[Windows][Linux] Support of '*' CSP rules for applications

pozdnyakov:
This patch enables manifest CSP rules containing '*' such as:
'http://*/' or 'https://*.example.com'.

BUG=XWALK-6745


pozdnyakov:
@darktears PTAL

darktears:
LGTM

PR #3691, opened 2016-05-07
[Android] Implement the feature about custom url scheme.

wuhengzhi:
This patch is to support the custom url scheme, e.g. "whatsapp://",
"intent://".
For "intent://", if the related app, e.g. Barcode Scanner, was
installed, then start the activity to launch this scanner, if it was
not installed, XWalkView will load the fallback url if it was specified.
For "whatsapp://", if whatsapp was installed, start the activity to
launch it, if it was not installed, then try to install it from market
according to the package name. If the package name is null, will report
"ERR_UNKNOWN_URL_SCHEME".

BUG=XWALK-6690

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkNavigationHandlerImpl.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkNavigationHandlerImpl.java
@@ -1,3 +1,4 @@
+// Copyright 2013 The Chromium Authors. All rights reserved.
 // Copyright (c) 2013 Intel Corporation. All rights reserved
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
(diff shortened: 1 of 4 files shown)


runtime/android/core_internal/src/org/xwalk/core/internal/XWalkNavigationHandlerImpl.java
```diff
@@ -32,22 +41,28 @@
...
     public boolean handleNavigation(NavigationParams params) {
         final String url = params.url;
+        if (UrlUtilities.isAcceptedScheme(url)) return false;
         Intent intent = null;
         if (url.startsWith(PROTOCOL_WTAI_PREFIX)) {
             intent = createIntentForWTAI(url);
         } else {
             intent = createIntentForActionUri(url);
         }
+        if (intent == null && shouldOverrideUrlLoadingInternal(params)) return true;
```
wuhengzhi: Move `shouldOverrideUrlLoadingInternal()` to here, after the action uri.

wuhengzhi:
@lincsoon PTAL and merge, thanks.

PR #3694, opened 2016-05-10
Fix race condition when RenderProcess or ExtensionProcess dies.

darktears:
When either of these two processes are dying then it may be that there
are still messages being filtered in the IPC channel. If so then the filter
may have dangling pointers to the extension thread server or the ui thread
server (as those have been deleted already). The solution is to invalidate
the filter when deleting the XWalkExtensionData object. Please note that
this commit partially revert XWALK-4295 beside the call to RemoveFilter in
OnRenderProcessHostClosed which was the cause of the crash (double delete
because RemoveFilter also delete the filter).

BUG=XWALK-6856

### DIFF
--- a/extensions/browser/xwalk_extension_data.cc
+++ b/extensions/browser/xwalk_extension_data.cc
@@ -6,6 +6,7 @@
 
 #include "content/public/browser/browser_thread.h"
 #include "xwalk/extensions/browser/xwalk_extension_process_host.h"
+#include "xwalk/extensions/browser/xwalk_extension_service.h"
 #include "xwalk/extensions/common/xwalk_extension_server.h"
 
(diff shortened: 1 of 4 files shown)


darktears:
@rakuco PTAL and verify Linux. Windows is ok, I've also tested https://crosswalk-project.org/jira/browse/XWALK-4295 to make sure...

rakuco:
le lgtm :fr:

darktears:
Please try Linux then I'll merge :)

darktears:
So you tried ....

rakuco:
I've just tried `xwalk_browsertest` and `xwalk_extensions_browsertest` and both pass (debug builds).

PR #3700, opened 2016-05-16
[Android] Enable the wide viewport quirk and implement APIs about loading overview mode.

wuhengzhi:
Enable the wide viewport quirk by default.
Implement embedding APIs: setLoadWithOverviewMode(), getLoadWithOverviewMode().
Patch for overview mode should be merged after enabling wide viewport quirk, otherwise will lead to one test case(testSetInitialScale2) failed.

BUG=XWALK-6762
BUG=XWALK-6859

### DIFF
--- a/test/android/core/javatests/src/org/xwalk/core/xwview/test/SetInitialScaleTest.java
+++ b/test/android/core/javatests/src/org/xwalk/core/xwview/test/SetInitialScaleTest.java
@@ -62,6 +62,7 @@ public class SetInitialScaleTest extends XWalkViewTestBase {
     @SmallTest
     @Feature({"setInitialScale"})
     public void testSetInitialScale2() throws Throwable {
+        setQuirksMode(false);
         CallbackHelper onPageFinishedHelper = mTestHelperBridge.getOnPageFinishedHelper();
 
         WindowManager wm = (WindowManager) getInstrumentation().getTargetContext()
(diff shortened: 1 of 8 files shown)


wuhengzhi:
@axinging @fujunwei @xzhan96 @lincsoon PTAL, thanks.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SetInitialScaleTest.java
```diff
@@ -75,12 +76,12 @@ public void testSetInitialScale2() throws Throwable {
...
         final float defaultScaleFactor = 0;
-        final float defaultScale = 0.5f;
-        final float scaleFactor = 0.25f;
+        final float defaultScale = getInstrumentation().getTargetContext(
+                ).getResources().getDisplayMetrics().density;
 
         assertEquals(defaultScaleFactor, getScaleFactor(), .01f);
         loadDataSync(null, page, "text/html", false);
-        assertEquals(scaleFactor, getScaleFactor(), .01f);
+        assertEquals(defaultScale, getPixelScale(), .01f);
```
axinging: I didn't understand where is this 0.01f from?

wuhengzhi: Function prototype:
`void assertEquals (float expected, 
                float actual, 
                float delta)`
If the two number's subtraction is less than delta, they are equals.

axinging:
LGTM

xzhan96:
Per my understanding of the wide viewport from https://developer.android.com/reference/android/webkit/WebSettings.html and https://chromiumcodereview.appspot.com/23496047

I prefer you refine the summary and commit log of this patch, just for your reference:

> Enable the wide viewport quirks mode default. 

Wide viewport should be enabled **by** default. 

> This  patch is to enable wide viewport quirks mode default. Without
enabling, the page is zoom out which is different from WebView.
Porting the test case for wide viewport from Android WebView.
Fixed the failed case of setInitialScale().

is it better to just explain the reason in commit log?  Hope below comment is useful.
 
```
According to the upstream, this setting should be always enabled,
and this quirk is just to maintain compatibility with Android apps.

Porting the corresponding test case from Android WebView, and 
fix the failed case setInitialScale.

Please refer to https://chromiumcodereview.appspot.com/23496047
for more details.
```

wuhengzhi:
@xzhan96 Thanks for the comments, will refine it.
This patch is for wide viewport quirks mode, wide viewport's embedding APIs has been implemented.

xzhan96:
LGTM

lincsoon:
Please do the rebase

wuhengzhi:
@lincsoon Rebase is done.

PR #3701, opened 2016-05-16
DEPS.xwalk: Re-roll chromium-crosswalk.

rakuco:
No new commits have been added, and there are no functional changes. The
only difference is that the "Cr-Commit-Position" line in the commit
message for commit "[Backport] Move recently added functions to
V8_USE_EXTERNAL_STARTUP_DATA" has been removed.

chromium-crosswalk commits that are not present upstream must never have
a "Cr-Commit-Position" line, otherwise `src/build/util/lastchange.py`
will use a commit hash that is not present in Chromium and
DevTools/remote debugging will fail to work as expected.

BUG=XWALK-5139
BUG=XWALK-6208

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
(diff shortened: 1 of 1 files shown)


rakuco:
@mrunalk: fyi

kenchris:
lgtm

PR #3702, opened 2016-05-19
[Android] Upgrade version of API to 6

lincsoon:


lincsoon:
@rakuco This version is used to mark the version of embedding API internally, it should be the same with    the version that we published on [official website](https://crosswalk-project.org/documentation/apis/embedding_api.html). This file is is used by shared mode to check compatibility.

lincsoon:
@rakuco

PR #3708, opened 2016-05-24
Make sure default schemes (https, http...) are allowed to store cookies.

darktears:
This is a regression after moving to M50 and the actual refactor due
to CookieMonster being hidden as an API.

https://github.com/crosswalk-project/crosswalk/commit/e1f52c3660f3868a11cccbae67593b435672160d
landed an incorrect change.

BUG=XWALK-6934


pozdnyakov:
lgtm

PR #3709, opened 2016-05-25
[Android] Implement the method to set compositing surface type for si…

xzhan96:
…ngle XWalkView

Currently, XWalkView compositing surface type(SURFACE_VIEW or TEXTURE_VIEW)
can only be configured by XWalkPreferences, which is shared by all XWalkViews.
This implementation provide a method to config the surface type in application's
layout file, and it also make each XWalkView has separate configuration.

BUG=XWALK-5449

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -116,7 +116,7 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
     // (ie before it is destroyed).
     private CleanupReference mCleanupReference;
 
-    public XWalkContent(Context context, AttributeSet attrs, XWalkViewInternal xwView) {
+    public XWalkContent(Context context, String animatable, XWalkViewInternal xwView) {
         // Initialize the WebContensDelegate.
         mXWalkView = xwView;
         mViewContext = mXWalkView.getContext();
@@ -134,7 +134,7 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
         MediaPlayerBridge.setResourceLoadingFilter(
                 new XWalkMediaPlayerResourceLoadingFilter());
 
-        setNativeContent(nativeInit());
+        setNativeContent(nativeInit(), animatable);
 
         XWalkPreferencesInternal.load(this);
         initCaptureBitmapAsync();
@@ -157,7 +157,7 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
             new Rect(), mGetBitmapCallback);
     }
 
-    private void setNativeContent(long newNativeContent) {
+    private void setNativeContent(long newNativeContent, String animatable) {
         if (mNativeContent != 0) {
             destroy();
             mContentViewCore = null;
@@ -166,10 +166,15 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
         assert mNativeContent == 0 && mCleanupReference == null && mContentViewCore == null;
 
         // Initialize ContentViewRenderView
-        boolean animated = XWalkPreferencesInternal.getValue(
-                XWalkPreferencesInternal.ANIMATABLE_XWALK_VIEW);
+        boolean animated;
+        if(animatable == null)
+            animated = XWalkPreferencesInternal.getValue(
+                    XWalkPreferencesInternal.ANIMATABLE_XWALK_VIEW);
+        else
+            animated = animatable.equalsIgnoreCase("true");
         CompositingSurfaceType surfaceType =
                 animated ? CompositingSurfaceType.TEXTURE_VIEW : CompositingSurfaceType.SURFACE_VIEW;
+        Log.d(TAG, "CompositingSurfaceType is " + (animated ? "TextureView" : "SurfaceView"));
         mContentViewRenderView = new ContentViewRenderView(mViewContext, surfaceType) {
             protected void onReadyToRender() {
                 // Anything depending on the underlying Surface readiness should
@@ -259,7 +264,7 @@ class XWalkContent implements XWalkPreferencesInternal.KeyValueChangeListener {
     }
 
     private void receivePopupContents(long popupNativeXWalkContents) {
-        setNativeContent(popupNativeXWalkContents);
+        setNativeContent(popupNativeXWalkContents, null);
 
         mContentViewCore.onShow();
     }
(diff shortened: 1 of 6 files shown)


xzhan96:
@lincsoon @axinging @wuhengzhi @fujunwei @XiaosongWei

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -166,10 +166,11 @@ private void setNativeContent(long newNativeContent) {
         assert mNativeContent == 0 && mCleanupReference == null && mContentViewCore == null;
 
         // Initialize ContentViewRenderView
-        boolean animated = XWalkPreferencesInternal.getValue(
+        boolean animated = animatable ? true : XWalkPreferencesInternal.getValue(
                 XWalkPreferencesInternal.ANIMATABLE_XWALK_VIEW);
```
lincsoon: Do you mean:
```
boolean animated = animatable || XWalkPreferencesInternal.ANIMATABLE_XWALK_VIEW);
```

I thought we should prioritize the value of single XWalkView. Anyway, it can't be differentiated that whether the developer set the value to false or didn't set the value at all. So I think we should use Boolean object instead of primitive type.

At last, could you make it clear in the description that which is prioritized when the value of single XWalkView conflicts with global value.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -73,12 +73,51 @@
...
- * TextureView</a> is intentionally used to render web pages for animation support.
+ * TextureView</a>, which is intentionally used to render web pages for animation support.
  * Besides, XWalkViewInternal won't be rendered if it's invisible.</p>
  *
+ * <p>Crosswalk provides two ways to choose TextureView or SurfaceView:</p>
+ * <ol><li>[To Be Deprecated] Set preference key
+ * {@link XWalkPreferencesInternal#ANIMATABLE_XWALK_VIEW} to true to use TextureView,
+ * and vice versa. Notice that all XWalkViews share the same preference value.</li>
+ * <li>Application developer can set this attribute for a single XWalkView by XML without
+ * impact on other XWalkViews, notice that do not use XWalkPreferencesInternal at the same time,
```
lincsoon: Why can't use both of them at the same time? If it isn't really allowed, we should make it throw an exception or anything else instead of mentioning in the doc.

xzhan96:
> At last, could you make it clear in the description that which is prioritized when the value of single XWalkView conflicts with global value.

yes, I am refine this point. 
My thinking is: we should check if there is setup on single XWalkView(layout), if yes, we should use it and ignore XWalkPreferences, if not, we use XWalkPreferences.
Do you think it's ok?

tools/reflection_generator/wrapper_generator.py
```diff
@@ -101,6 +101,11 @@ def GenerateFields(self):
       fields_string += "    private ArrayList<Object> constructorTypes;\n"
       fields_string += "    private ArrayList<Object> constructorParams;\n"
       fields_string += "    private ReflectMethod postWrapperMethod;\n"
+    if self._java_data.wrapper_name == "XWalkView":
+      fields_string += "    private boolean sAnimatable = false;\n"
+      fields_string += "    \n"
+      fields_string += "    public static final String XWALK_ATTRS_NAMESPACE\
+ = \"http://schemas.android.com/apk/res-auto\";"
```
lincsoon: What's the use case if we make this value public?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -166,10 +166,15 @@ private void setNativeContent(long newNativeContent) {
...
 
         // Initialize ContentViewRenderView
-        boolean animated = XWalkPreferencesInternal.getValue(
-                XWalkPreferencesInternal.ANIMATABLE_XWALK_VIEW);
+        boolean animated;
+        if(animatable == null)
+            animated = XWalkPreferencesInternal.getValue(
+                    XWalkPreferencesInternal.ANIMATABLE_XWALK_VIEW);
+        else
+            animated = (animatable == "true")? true : false;
```
lincsoon: prefer `equalIngoreCaes` to operator `=`

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -269,6 +309,8 @@ public XWalkViewInternal(Context context) {
     @XWalkAPI(preWrapperLines = {
                   "        super(${param1}, ${param2});",
                   "        if (isInEditMode()) return;",
+                  "        mAnimatable = ${param2}.getAttributeValue(",
+                  "                XWALK_ATTRS_NAMESPACE, \"animatable\");",
```
lincsoon: Why not just use the string directly for the name space? Or use the static member for both of name space and attribute name.

tools/reflection_generator/java_method.py
```diff
@@ -269,6 +270,12 @@ def ParseMethodAnnotation(self, annotation):
       post_wrapline = self.FormatWrapperLine(match.group('post_wrapline'))
       self._method_annotations[self.ANNOTATION_POST_WRAPLINE] = post_wrapline
 
+    post_bridline_re = re.compile('postBridgeLines\s*=\s*\{\s*('
```
lincsoon: `brid` or `bird`? Anyway, I don't know why to prefer `post_brid_line` to `post_bridge_line`

tools/reflection_generator/java_method.py
```diff
@@ -482,12 +489,18 @@ def GenerateBridgeConstructor(self):
 
         this.wrapper = wrapper;
         reflectionInit();
+${POST_BRID_LINES}
     }
 
     """)
+
+      post_bird_string = self._method_annotations.get(
```
lincsoon: `bird` or `brid`?

xzhan96: typo, it should be brid.
I just refer to post_wrapline_re, I don't know why it not use post_wrapperline_re.
Do you think post_bridge_string is better?

lincsoon: It would be better to understand. But it doesn't really matter here.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
xzhan96: Actually, I used static member for both of them in previous version, I changed it after the bridge_generator.py can parse the ["]. 
I will modify it as you suggestion.

xzhan96:
@lincsoon please help to review

suleymanccelik:
Thanks @xzhan96 . Will your commit be added to next crosswalk-android version?

xzhan96:
@suleymanccelik 
sure, it should be added into Crosswalk-20, so stay tuned :)

PR #3710, opened 2016-05-25
[Linux][Windows] Implement navigation history of links.

darktears:
So they can change color appropriately if you already navigated that URL.
Mostly borrowing the work already done for Android.

BUG=XWALK-6935

### DIFF
--- a/runtime/browser/runtime.cc
+++ b/runtime/browser/runtime.cc
@@ -10,6 +10,7 @@
 #include "base/command_line.h"
 #include "base/message_loop/message_loop.h"
 #include "base/strings/utf_string_conversions.h"
+#include "base/threading/non_thread_safe.h"
 #include "components/app_modal/javascript_dialog_manager.h"
 #include "components/content_settings/core/common/content_settings.h"
 #include "components/content_settings/core/common/content_settings_types.h"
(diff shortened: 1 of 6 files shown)


darktears:
@pozdnyakov @lincsoon PTAL

lincsoon:
Sorry, I'm not familiar with the code around here. @XiaosongWei @axinging Could you have a look?

runtime/browser/runtime.cc
```diff
@@ -316,6 +317,16 @@ void Runtime::TitleWasSet(content::NavigationEntry* entry, bool explicit_set) {
     ui_delegate_->UpdateTitle(entry->GetTitle());
 }
 
+void Runtime::DidNavigateAnyFrame(
+    content::RenderFrameHost* render_frame_host,
+    const content::LoadCommittedDetails& details,
+    const content::FrameNavigateParams& params) {
+  DCHECK(thread_checker_.CalledOnValidThread());
+
+  XWalkBrowserContext::FromWebContents(web_contents())
```
pozdnyakov: web_contents_->GetBrowserContext()

darktears: I think it's better like it is, otherwise your suggestion needs a cast from a BrowserContext to a XWalkBrowserContext.

runtime/browser/android/renderer_host/xwalk_render_view_host_ext.cc
```diff
@@ -105,16 +104,6 @@ void XWalkRenderViewHostExt::RenderProcessGone(base::TerminationStatus status) {
   }
 }
 
-void XWalkRenderViewHostExt::DidNavigateAnyFrame(
```
axinging: Any background about move this from xwalk_render_view_host_ext.h to runtime.h?

darktears: because both (Runtime and XWalkRenderViewHost) are WebContentsObserver and they seem to observe the same web content object. xwalk_render_view_host_ext seems to be the added bit to handle Android specific code related to stuff happening with the web content object. I do not see why I should duplicate the same functionality/code. Runtime is used and created in Android so both are going to get the same method calls. That's my understanding.

XiaosongWei: @darktears Isn't Runtime only used in Linux/Windows? Per my understanding Android doesn't use Runtime.

darktears:
Since there is no more Android part and that @pozdnyakov already said LGTM on IRC, merging.

PR #3715, opened 2016-05-27
[Android] Filter JARs being merged by the classes they contain.

rakuco:
This builds upon the initial work done in commit 862201b ("[Android]
Only merge JARs which we have built ourselves").

That commit filtered the JARs being merged based on whether they were
located inside our build directory. That is not enough, as we sometimes
build a few JARs that contain classes in packages that can clash with
other JARs a user may use:
* `jsr_305_javalib.jar` has classes in the `javax.annotation` package.
* `protobuf_nano_javalib.jar` has classes in the
  `com.google.protobuf.nano` package.
The latter in particular is built as a dependency for the
`cardboard.jar` blob (itself a dependency for WebVR support) and causes
conflicts with LibPhoneNumber.

So instead of filtering JARs by their location, we now inspect each JAR
and skip those with classes outside the package namespaces we allow (at
the moment, `org.xwalk`, `org.chromium` and `SevenZip`).

For WebVR support, which is what originally prompted this change, this
means users must bundle `cardboard.jar` into their projects themselves
if they want WebVR support to work -- the alternative, bundling
`cardboard.jar` ourselves, means going back to XWALK-5092 (i.e. the
build would break for users who do bundle `cardboard.jar` themselves).

We also do some extra validation when generating
`xwalk_core_library_java.jar` and make sure the list of JARs we are
known to skip is entirely correct. We still need a more automated way to
let users know which JARs they need to embed together with Crosswalk.   

BUG=XWALK-5092
BUG=XWALK-6544
BUG=XWALK-6597

### DIFF
--- a/build/android/merge_jars.py
+++ b/build/android/merge_jars.py
@@ -7,12 +7,16 @@
 
 """
 Given a list of JAR files passed via --jars, produced one single JAR file with
-all their contents merged. JAR files outside --build-dir are ignored.
+all their contents merged. JAR files with classes outside org.xwalk,
+org.chromium and a few others are ignored to avoid clashing with other
+packages.
 """
 
-import optparse
+import argparse
+import fnmatch
 import os
 import sys
+import zipfile
 
 GYP_ANDROID_DIR = os.path.join(os.path.dirname(__file__),
                                os.pardir, os.pardir, os.pardir,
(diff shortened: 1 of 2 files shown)


rakuco:
@wuhengzhi: PTAL. This implements the idea I suggested in XWALK-6597, but I did not check if something along the lines of what you did in #3620 (whether bundling `cardboard.jar` in `libs/` causes clashes for any users who include cardboard from another source in their projects themselves).

build/android/merge_jars.py
```diff
@@ -25,20 +29,38 @@
...
 
+def IsMergeableJar(jar_name):
+  """
+  Returns True if a certain JAR does not have any classes outside the
+  org.chromium and org.xwalk namespaces.
+  """
+  with zipfile.ZipFile(jar_name) as zip_file:
+    for entry_name in zip_file.namelist():
+      if entry_name in ('META-INF/', 'META-INF/MANIFEST.MF') or \
+         fnmatch.fnmatchcase(entry_name, 'SevenZip/*.class') or \
```
wuhengzhi: > outside the org.chromium and org.xwalk namespaces.

This comment does not refer to  other namespaces  `SevenZip`.
Could we put `SevenZip/*.class`, `org/chromium/*.class` and `org/xwalk/*.class` to a list that we can give a general comment and expand it conveniently?

build/android/merge_jars.py
```diff
@@ -25,20 +29,38 @@
...
   parser.add_option('--output-jar', help='Name of the merged JAR file.')
 
   options, _ = parser.parse_args()
-  build_dir = os.path.abspath(options.build_dir)
 
   with build_utils.TempDir() as temp_dir:
     for jar_file in build_utils.ParseGypList(options.jars):
-      if not os.path.abspath(jar_file).startswith(build_dir):
+      # If a JAR has classes outside our allowed namespaces (mostly
+      # org.chromium,and org.xwalk), we need to skip it otherwise there can be
```
wuhengzhi: Expect a space before `and`.

rakuco:
@wuhengzhi: thanks for the review, I've updated the patch to address the suggestions.

build/android/merge_jars.py
```diff
@@ -25,20 +29,53 @@
 from util import build_utils
 
 
+
+# The list below contains all entries allowed in JAR files (it can contain
+# wildcards).
+# If a JAR contains an entry that does not match the white list below, it will
+# be rejected and not merged.
+JAR_ENTRY_WHITELIST = (
+  'META-INF',
```
wuhengzhi: `META-INF` -> `META-INF/`, otherwise all of the jars will be rejected.

rakuco:
@wuhengzhi thanks for the attention to the details. I've uploaded a new version.

rakuco:
ping @wuhengzhi

wuhengzhi:
It's an awesome solution to detect the skipped and missed jars.
LGTM, thanks.

PR #3716, opened 2016-05-27
[Android] Fix crash when setting surface type through layout

lincsoon:
XWalkViewBridge will try to get the member mAnimatable from XWalkView.
But if the class type in the layout file is XWalkView's subclass, the
member can't be found and a exception is thrown. This patch will look
for the member by going up the class hierarchy.


lincsoon:
@xzhan96

xzhan96:
LGTM, thanks

PR #3718, opened 2016-05-27
[Lite] support enable-xxx flags defined in GYP_DEFINES

tingshao:
enable-xxx flags can be added to GYP_DEFINES to control
which feature is enabled.
@halton @heke123 @PeterWangIntel


PeterWangIntel:
LGTM

PR #3722, opened 2016-05-27
Handle URLs without seperators in CSP rules properly

pozdnyakov:
This PR fixes handling of URLs without separtors e.g. "frame-src https://crosswalk-project.org"

BUG=XWALK-6906


darktears:
LGTM

PR #3727, opened 2016-05-31
[Android] Fix a crash issue of shared mode

lincsoon:
The old version of XWalkView doesn't have the private member
mAnimatable. When those versions of shared app are trying to link to
newest library, XWalkViewBridge in the library side can't find the
member and will throw an exception which causes the app crash.


lincsoon:
@xzhan96

xzhan96:
LGTM, :+1:

PR #3734, opened 2016-06-02
[Linux] Fix the crash in ColorChooser

XiaosongWei:
The original implementation with CreateWindowWithContext may cause
X receive a bad window parameter and the ColorChooserView is placed
at a fixed position but not bundled with the root window.

> The program 'xwalk' received an X Window System error.
This probably reflects a bug in the program.
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 1767 error_code 3 request_code 3 minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
 
Now we use CreateWindowWithParent to parent the ColorChooserView with
toplevel window to avoid the X error mentioned above and the view is
bundled with the root window and it can be moved together with the root
window.

BUG=XWALK-6958

### DIFF
--- a/runtime/browser/ui/color_chooser_aura.cc
+++ b/runtime/browser/ui/color_chooser_aura.cc
@@ -57,9 +57,8 @@ ColorChooserAura::ColorChooserAura(content::WebContents* web_contents,
                                    SkColor initial_color)
     : web_contents_(web_contents) {
   view_ = new views::ColorChooserView(this, initial_color);
-  widget_ = views::Widget::CreateWindowWithContext(
-      view_, web_contents->GetNativeView());
-  widget_->SetAlwaysOnTop(true);
+  widget_ = views::Widget::CreateWindowWithParent(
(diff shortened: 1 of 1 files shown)


XiaosongWei:
@pozdnyakov @darktears Please take a look, Thanks!

pozdnyakov:
lgtm

XiaosongWei:
@darktears @rakuco PTAL

XiaosongWei:
@rakuco Could you please help to merge? Thanks!

rakuco:
This looks identical to https://codereview.chromium.org/65733002, although the upstream CL does not mention any crashes.
I'm assuming you also want to make the color chooser browser modal, not always-on-top, right?

XiaosongWei:
@rakuco yes, you are right. it would be not always on top.

PR #3735, opened 2016-06-03
M51: Fix browsertests on linux

mrunalk:
We were getting following error on all of the browsertests,
[109919:109919:0602/121714:FATAL:browser_context.cc(350)] Check failed: browser_context->GetUserData(kMojoWasInitialized). Attempting to get the mojo user id for a BrowserContext that was never Initialize()ed
becasue we were not initilizing the BrowserContext. This patch fixes it by
calling BrowserContext::Initialize() with appropriate path.

adapted from,
https://codereview.chromium.org/1741953002

BUG=XWALK-6951


mrunalk:
@darktears PTAL

darktears:
LGTM

PR #3737, opened 2016-06-06
[Android] Fix the crash issue when use the cordova plugin.

wuhengzhi:
In cordova-plugin-crosswalk-webview, the AttributeSet is null, please
look at
https://github.com/crosswalk-project/cordova-plugin-crosswalk-webview/blob/master/platforms/android/src/org/crosswalk/engine/XWalkCordovaView.java#L55
app will crash with null pointer.
Add the conditional for AttributeSet.

BUG=XWALK-6960

(cherry picked from commit 5bb898ac7311eace05a9a67d92ed5f25107fc34c)


wuhengzhi:
@lincsoon Please help merge to Crosswalk-20, thanks.

PR #3740, opened 2016-06-07
Refine destroy order in XWalkContent.java

chuan9:
BUG=XWALK-6840

XWalk would destroy contentViewRenderView first and then cleanupReference which points
other native object such as exitFullScreen, it will use contentViewRenderView
when in full screen. So app would cause crash. The correct behavior is that exit full
screen first and then destroy contentViewRenderView.


rakuco:
Your commit message contains zero information about why this is being done, it is all being explained only in the pull request message. "destroy" is being misspelled as "destory", and the call is actually called `onDestroy`.

fujunwei:
LGTM.

chuan9:
@lincsoon @fujunwei @rakuco Please have a review, thanks!

PR #3745, opened 2016-06-08
crosswalk-19: Build with enable_webvr=0 by default.

rakuco:
**IMPORTANT**
We actually intend to fix this properly. We are cherry-picking
this commit from the crosswalk-18 branch first in order to have a stable
baseline with no WebVR support propagating from Crosswalk 21 to 20 to 19.
After that, we can start working on fixing this in 21 and 20 and, if there
is enough time, 19 too.

*From the original commit message:*

Disable WebVR support on Android by default, since it depends on
`third_party/cardboard-java` which in turn needs
`third_party/android_protobuf`.

The dependency on `android_protobuf` means `xwalk_core_library_java.jar`
bundle Protobuf and cause build conflicts for users depending on Protobuf
themselves (or using other libraries that pull it).

Related BUG=XWALK-6597

### DIFF
--- a/DEPS.xwalk
+++ b/DEPS.xwalk
@@ -17,7 +17,7 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = '2ae20d6a4a9d8362a398e4e0f6a7574de50fc9ab'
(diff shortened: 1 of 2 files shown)


kenchris:
rs=me

PR #3747, opened 2016-06-15
[Android] Expose embedding apis for video fullscreen.

wuhengzhi:
The onShow/HideCustomView() was implemented in Crosswalk by default and
not exposed, it could not be overrided by customer to implement the video
fullscreen.
This patch is to expose onShow/HideCustomView().

BUG=XWALK-6939

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java
@@ -1,0 +1,22 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package org.xwalk.core.internal;
+
+/**
+ * A callback used by the host application to notify
+ * the current page that its custom view has been dismissed.
+ */
+@XWalkAPI(impl = CustomViewCallbackInternal.class, createInternally = true)
+public class CustomViewCallbackHandlerInternal implements CustomViewCallbackInternal {
+
+    /**
+     * Invoked when the host application dismisses the
+     * custom view.
+     * @since 7.0
+     */
+    @XWalkAPI
+    public void onCustomViewHidden() {
+    }
+}
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
@@ -8,17 +8,21 @@ import android.app.Activity;
 import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
+import android.content.pm.ActivityInfo;
 import android.graphics.Bitmap;
 import android.net.Uri;
 import android.os.Build.VERSION;
 import android.os.Build.VERSION_CODES;
 import android.os.Message;
+import android.view.Gravity;
 import android.view.KeyEvent;
 import android.view.View;
+import android.view.ViewGroup;
 import android.view.Window;
 import android.view.WindowManager;
 import android.webkit.ValueCallback;
 import android.widget.EditText;
+import android.widget.FrameLayout;
 
 import org.chromium.base.ApiCompatibilityUtils;
 
@@ -36,6 +40,11 @@ public class XWalkUIClientInternal {
     private boolean mOriginalFullscreen;
     private boolean mOriginalForceNotFullscreen;
     private boolean mIsFullscreen = false;
+    private View mCustomXWalkView;
+    private final int INVALID_ORIENTATION = -2;
+    private int mPreOrientation = INVALID_ORIENTATION;
+    private CustomViewCallbackInternal mCustomViewCallback;
+    private XWalkContentsClient mContentsClient;
 
     /**
      * Initiator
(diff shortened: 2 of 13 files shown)


wuhengzhi:
@fujunwei @axinging @lincsoon @xzhan96 PTAL, thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
```diff
@@ -522,4 +531,100 @@ public void onCancel(DialogInterface dialog) {
...
     }
+
+    void setContentsClient(XWalkContentsClient client) {
+        mContentsClient = client;
+    }
+
+    private Activity addContentView(View view, CustomViewCallbackInternal callback) {
+        Activity activity = mXWalkView.getActivity();
+        if (mCustomXWalkView != null || activity == null) {
+            if (callback != null) callback.onCustomViewHidden();
```
fujunwei: The callback will be Invoked when the host application dismisses the custom view explained in https://developer.android.com/reference/android/webkit/WebChromeClient.CustomViewCallback.html, customer need to exit full screen mode.
The simple code in java/src/org/chromium/android_webview/AwContentViewClient.java
```
WebChromeClient.CustomViewCallback cb = new WebChromeClient.CustomViewCallback() {
            @Override
            public void onCustomViewHidden() {
                if (mCustomView != null) {
                    mAwContents.requestExitFullscreen();
                }
            }
        };
```
But above logic isn't align with it's usage?

runtime/android/core_internal/src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * Invoked when the host application dismisses the
+     * custom view.
+     * @since 7.0
+     */
+    @XWalkAPI
+    public void onCustomViewHidden() {
+    }
+}
```
xzhan96: I think CustomViewCallbackInternal.java is for reflection, right? If yes, why need the same annotation here?

tools/reflection_generator/java_method.py
```diff
@@ -415,11 +416,12 @@ def FormatSingleParam(self, param_type, param_name, param_string_type):
...
-          not java_data.HasInstanceCreateInternallyAnnotation()):
-        return'(%s instanceof %s)' % (param_name, java_data.GetBridgeName())
-      else:
-        return None
+      if is_internal_class:
+        if not java_data.HasInstanceCreateInternallyAnnotation():
+          return'(%s instanceof %s)' % (param_name, java_data.GetBridgeName())
+        else:
+          self._parameter_is_internal = True
+      return None
```
xzhan96: sorry, I don't understand here, what's the different between `CustomViewCallbackHandlerInternal&CustomViewCallbackInternal` and `ClientCertRequestHandlerInternal & ClientCertRequestInternal` that need you change the java_method.py?

xzhan96:
@wuhengzhi is this PR still vaild?

wuhengzhi:
Yes, I will answer your questions later. Please wait a moment, thanks.

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
wuhengzhi: Before adding custom view, we should check the condition.
The callback was exposed in `CustomViewCallbackInternal.java` for user to invoke.

runtime/android/core_internal/src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java
wuhengzhi: Expose it for the default implementation for external use.
User could use `CustomViewCallbackHandler` to create instance without implementing a new handler which inheriting from `CustomViewCallbackInternal'.

xzhan96: I am not sure, but my concern is:
just like `ClientCertRequestHandlerInternal & ClientCertRequestInternal`? CustomViewCallbackInternal is the implementation of CustomViewCallbackInternal, after reflection, customer can use the interface in CustomViewCallback, which actually the same with CustomViewCallbackInternal's, so I don't know why need reflect both of them?

tools/reflection_generator/java_method.py
wuhengzhi: It has no relation with `CustomViewCallbackHandlerInternal&CustomViewCallbackInternal`.
Its for `onShowCustomView(View view, CustomViewCallbackInternal callback)` in `XWalkUIClientInternal.java`, please notice the first parameter, its not `XWalkView`.
From the comment `XWalkViewInternal view => (view instanceof XWalkViewBridge)`, it just checks the view for XWalkView, so need to modify code for `View`.

runtime/android/core_internal/src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java
wuhengzhi: Yes, customer could use `CustomViewCallback` to implement self handler, but customer also could use default handler which we exposed.

xzhan96: I am still worry about it, I do think customer maybe confuse about CustomViewCallbackHandler and CustomViewCallback. @fujunwei @axinging @lincsoon any comment on this point?

wuhengzhi:
ping @lincsoon @axinging

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkUIClientInternal.java
```diff
@@ -522,4 +531,100 @@ public void onCancel(DialogInterface dialog) {
         mDialog.show();
         return false;
     }
+
+    void setContentsClient(XWalkContentsClient client) {
+        mContentsClient = client;
+    }
+
+    private Activity addContentView(View view, CustomViewCallbackInternal callback) {
+        Activity activity = mXWalkView.getActivity();
```
axinging: Will this work on non-Activity environment?

wuhengzhi: I will rebase it, thanks.

lincsoon:
Are the discussions above all settled?

runtime/android/core_internal/src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java
wuhengzhi: @lincsoon  Please  look at above comment.  Do  you mean this discussion?

wuhengzhi: @xzhan96 CustomViewCallbackHandlerInternal should be exposed since it will generate CustomViewCallbackHandlerBridge which will be used in XWalkUIClientBridge.java.

>     @Override
    public void onShowCustomView(View view, CustomViewCallbackInternal callback) {
        onShowCustomView(view, (callback instanceof CustomViewCallbackHandlerBridge ? ((CustomViewCallbackHandlerBridge) callback ) : new CustomViewCallbackHandlerBridge((CustomViewCallbackHandlerInternal) callback)));
    }

PR #3749, opened 2016-06-16
DEPS.xwalk: Roll chromium-crosswalk and v8-crosswalk

mrunalk:
This corrects the commit hashes of chromium-crosswalk and v8-crosswalk
when I rolled Chromium 51.0.2704.84 earlier.


mrunalk:
@darktears, Earlier I rolled in incorrect commit hashes when I pushed M51. This should fix it.

darktears:
LGTM

PR #3738, opened 2016-06-06
[Windows] WiFi Direct API

astojilj:
Initial implementation of WiFi Direct experimental API on Windows.
Supported only on Windows10 and Android.
JS API is the same as in Android.

BUG=XWALK-5752

PR #3753, opened 2016-06-20
Revert "[Windows] WiFi Direct API"

astojilj:
Reverts crosswalk-project/crosswalk#3738


astojilj:
scoped_ptr is removed.
@pozdnyakov could you please check the revert?

pozdnyakov:
rs=me

rakuco:
You should have explained why your change had to be reverted. If a year from now I look at your commit message I have no idea why the wifi direct API code was landed and then reverted, the only thing that looks like an explanation is in a comment in the pull request.

PR #3757, opened 2016-06-21
Remove messaging, contacts and device capabilities SysApps API

tingshao:
Those API extensions are not needed anymore and should be removed.

BUG=XWALK-6875
@halton


tingshao:
In XWalkExtensionsIFrameTest.CrossContextsReferenceShouldNotCrash testcase, call fs sysapp API to verify the cross context reference instead of the removed device capability API.
@halton

halton:
LGTM, @rakuco before I merge this PR, any concerns? CC @baleboy

baleboy:
Basically this is fine, but could you list the APIs that are being removed? I'd like to send a deprecation email with the details.

tingshao:
@baleboy 
Only contacts, messages and device capability are removed.
others are kept, for example, the storage api is used by our realsense extension.

PR #3759, opened 2016-06-22
Update release notes for Crosswalk stable

baleboy:
Updating the release notes to reflect the fact that Crosswalk 19 is stable. The same change will be applied to the 19 and 20 branches.


rakuco:
lgtm!

PR #3762, opened 2016-06-28
[Android] Support the feature of proxy in Crosswalk

fujunwei:
Crosswalk for android will update proxy when a PROXY_CHANGE intent is received,
but it’s need system permission to send the broadcast, there is also a workaround with
reflecting the incompatible api, but it’s too complex to maintain those codes, so
implement the new API in XWalkView to set proxy for images/text resources.

BUG=XWALK-6769


fujunwei:
@lincsoon @xuxin @wuhengzhi PTAL.

lincsoon:
@fujunwei Please rebase this commit and make it pass the try-bot

Kudo:
@fujunwei  What is this PR's status? I would like to see if Crosswalk could support custom proxy settings

PR #3765, opened 2016-06-28
Init IPC attachment broker for extension process

pozdnyakov:
The IPC::AttachmentBroker must be initialized within
any process that uses IPC channel, however this was not
done so far for extension process in Crosswalk.

This patch initializes the attachment brocker for extension
process with an 'AttachmentBrokerPrivileged' instance as 
extension process is privileged.

BUG=XWALK-7021


pozdnyakov:
@rakuco @darktears PTAL

kenchris:
lgtm

rakuco:
thanks. it's interesting to know that this was basically a one-liner in the end...

PR #3767, opened 2016-06-29
[Android] Refactor XWalkView to receive events directly.

wuhengzhi:
In Crosswalk, ContentView and ContentViewRenderView were used to render
page, some events were dispatched from ContentView to XWalkView, such as
touching and clicking. Its inefficiency and in latest version of
Chromium, ContentView was declared as private, so it should be
deprecated.

Now XWalkViewInternal implements all of the related interfaces and
overrides all of the functions which are used in ContentView, so
XWalkViewInternal could be regarded as another ContentView.
ContentViewRenderView and XWalkViewInternal were added to XWalkView
and XWalkViewInternal could receive the events directly.
Also fix some bugs.

BUG=XWALK-6118


wuhengzhi:
@shaoboyan @XiaosongWei @xzhan96 @fujunwei @axinging @lincsoon 
Please help to review. Thanks.

xzhan96:
> Its inefficiency and in **latest version** of Chromium, ContentView was declared as private, so it should be deprecated.

Do you think it's better to clear the version as from Chromium 5X..?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
```diff
@@ -325,6 +340,14 @@ public XWalkViewInternal(Context context) {
...
         mContext = getContext();
 
+        if (getScrollBarStyle() == View.SCROLLBARS_INSIDE_OVERLAY) {
+            setHorizontalScrollBarEnabled(false);
+            setVerticalScrollBarEnabled(false);
+        }
+
+        setFocusable(true);
+        setFocusableInTouchMode(true);
+
```
xzhan96: There is a JIRA [XWALK-5815](https://crosswalk-project.org/jira/browse/XWALK-5815), about the fact that XWalkView is SurfaceView in default that is not focusable and then doesn't broadcast keyevent, so customer can't override OnKeyListener to pass onKey event, but there is a workaround (actually I never successful, but customer was!) in stackoverflow just as the same with above two lines:

> setFocusable(true);
setFocusableInTouchMode(true);
requestFocus();

**My point** is that since you add them here, does that means XWALK-5815 works well with nothing additional workaround?

test/android/core_internal/javatests/src/org/xwalk/core/internal/xwview/test/XWalkViewInternalTestBase.java
```diff
@@ -241,7 +241,30 @@ protected void setUp() throws Exception {
             @Override
             public void run() {
                 mXWalkViewInternal = new XWalkViewInternal(activity, activity);
-                getActivity().addView(mXWalkViewInternal);
+                // Create new FrameLayout as a container to add XWalkViewInternal and
+                // ContentViewRenderView, it is just for test case
+                // "OnShowOnHideCustomViewTest.testOnShowCustomViewAndPlayWithHtmlControl".
+                // The play button should be rendered above the video right in
```
xzhan96: Since it is just for special cases, any more explanation about why add the change here?

test/android/core_internal/javatests/src/org/xwalk/core/internal/xwview/test/XWalkViewInternalTestBase.java
```diff
@@ -241,7 +241,30 @@ protected void setUp() throws Exception {
...
+                // the middle of the custom view, if not, test case will be failed.
+                // But it still has the failure rate, maybe we need to disbale this test case.
```
xzhan96: Do you see any rules about the failure rate? or just random?

wuhengzhi: The rate is indefinitely. I have ran it about 20 times, only one time failed.

test/android/core_internal/javatests/src/org/xwalk/core/internal/xwview/test/XWalkViewInternalTestBase.java
wuhengzhi: `XWalkView` does not display in the middle of the screen without `FrameLayout`.
With following code to force `XWalkView` to display in the middle.
`                         new FrameLayout.LayoutParams(
                                 FrameLayout.LayoutParams.MATCH_PARENT,
                                 FrameLayout.LayoutParams.MATCH_PARENT,
                                 Gravity.CENTER_HORIZONTAL|Gravity.CENTER_VERTICAL)`.

wuhengzhi:
> Do you think it's better to clear the version as from Chromium 5X..? Actually I didn't find any hint about the change in Chromium.

I will add the version, you can find it at https://github.com/crosswalk-project/chromium-crosswalk/commit/f9ad3afca4c5f0621b497c299ac43dcd2cf2b3e1

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewInternal.java
wuhengzhi: Maybe, we can verify it later.

PR #3768, opened 2016-06-30
Add GN build support in Crosswalk.

heke123:
Now it can build Crosswalk for Android and Linux

    Authors:
      peter.h.wang@intel.com,  ke.he@intel.com

    Usage:
      In Chromium src
           python xwalk/gn_xwalk --target-os='android' --is-debug=false
      See more info by run:
           python xwalk/gn_xwalk -h

      By default, the target_cpu is set as x86, the .ninja files are generated in folder is
      'out/Default'. Then
           ninja -C out/Default xwalk_builder

    Explanation of patch:
      1) The neccessary changes in source files are enclosed by macro "GN_BUILDING"
      2) Some utilities scripts for GN are added to "build/android/gn"
         We will clean gyp-special scripts once transferring to GN is succesess
      3) 'BUILD.gn's are ditributing in correspondent sub-folders.
         Followings show the relationship between main targets:

                  xwalk_core_shell_apk (runtime/android/core_shell/BUILD.gn)
               _______________|________________________________
               |                   |                           |
               |                   |         xwalk_core_internal_java (runtime/android/core_internal/BUILD.gn)
               |             xwalk_core_java (runtime/android/core/BUILD.gn)
               |
         libxwalkcore (android/BUILD.gn)
               |
         xwalk_runtime (./BUILD.gn)

                              xwalk (./BUILD.gn)
               _________________|_________________________
               |                                          |
         xwalk_runtime (./BUILD.gn)                   xwalk_main.cc
        _______|___________________________________________________
        |                     |                                    |
     Chromium's target    extensions (extensions/BUILD.gn)   sysapps (sysapps/BUILD.gn)

    TODO:
      1) Building for Windows.
      2) Features for Android: Add LZMA support.
      3) Features for Linux, such as nacl, file picker etc.

    BUG=XWALK-3674


heke123:
@rakuco  I found the failed reason is "gn.py: Could not find gn executable at: /home/crwuser/build-infrastructure/slave/crosswalk-android-x86/build/buildtools/linux64/gn".
My understanding is the build server should ignore all the BUILD.gn files before we do some GN configuration on it.  But it seems that our build server already knows to build by GN once it finds the BUILD.gn files. 
Could you please help to give a hint? thanks very much.

heke123:
@PeterWangIntel  @halton  @mrunalk  Thanks.

heke123:
@rakuco  the build server complains **"gn.py: Could not find gn executable at: /home/crwuser/build-infrastructure/slave/crosswalk-android-x86/build/buildtools/linux64/gn"**.
I found that if we run gn command when the current working dir is "src/xwalk",  it complains the similar error. The reason is in depot_tools/gn.py it returns the wrong bin_path of gn executable. 
Gn.py tries to find the bin_path by .gclient_entries, the contents of .gclient_entries is **'src/xwalk': 'https://github.com/crosswalk-project/crosswalk.git'**, it mistakenly regards src/xwalk as a primariy path.
I think if we set the current working dir as "src" instead of "src/xwalk" before run gn in build server,  everything should be ok.  Otherwise we have to change the script of depot_tools.
Could you please help to give a check? thanks.

halton:
@heke123  as PR #3779 get merged, please force update your branch to re-trigger the trybot.

heke123:
@rakuco  @mrunalk   Please have a review, Thanks!  
I think we need to switch the trybot build script into GN too.

heke123:
@halton  @PeterWangIntel  Thanks.

rakuco:
From a very quick look:
* It makes much more sense to move the lists of files to a `.gypi` file so they can be shared between GN and gyp while we do not migrate everything. Just moving the existing lists to `.gypi` files (with no GN involved) would be a very good first commit.
* There's no explanation as to why `GN_BUILDING` and the other `if defined()` checks are needed.

PeterWangIntel:
There are 2 reasons to use GN_BUILDING:
(1) The default folder to generate files changed when use GN, so we have to change the including path of header files.
      Take "xwalk_sysapps_resources.h"  as example, GYP generate it in "out/Debug/gen/grit" by default, by GN generate it in "out/Debug/gen/xwalk/sysapps/grit".
      All the "GN_BUILDING" enclosing header files is caused by that reason.
(2) Some default build flag of GN are different to GYP that causes build error, such as "compare unsigned int with signed". 
     We'll investigate these trivial flags to clean that sort of "GN_BUILDING"

heke123:
@rakuco  It is good idea to make GN and GYP to shared the same .gypi which contains file lists, but the situation is a little bit complex.
For example, in GYP src files can be set by "java_in_dir",  but in GN we can only list the file names in "java_files".   
And, in GN the new generated java sources files(like reflection bridge and wrapper layer) are added into a srcjar first, then add this srcjar into the "srcjar_deps" of "Android_library" target.  "Android_library" complains error if we add the new generated java sources file directly in its "java_files".
I'm sorry we directly submit a huge PR instead of several small ones, I understand it is really tough to review. Let's discuss how to cope with this.

PR #3771, opened 2016-07-01
[Android] Expose set/getCacheMode and set/getBlockNetworkLoads in XWa…

xzhan96:
…lkSettings

XWalk already has the implementation of them, and because they share the same
test case, there expose them at the same commit.

BUG=XWALK-6832

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkSettingsInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkSettingsInternal.java
@@ -36,6 +36,41 @@ public class XWalkSettingsInternal {
         TEXT_AUTOSIZING
     }
 
+    /**
+     * Default cache usage mode. If the navigation type doesn't impose any
+     * specific behavior, use cached resources when they are available
+     * and not expired, otherwise load resources from the network.
+     * Use with {@link #setCacheMode}.
+     * @since 7.0
+     */
+    @XWalkAPI
+    public static final int LOAD_DEFAULT = -1;
+
+    /**
+     * Use cached resources when they are available, even if they have expired.
+     * Otherwise load resources from the network.
+     * Use with {@link #setCacheMode}.
+     * @since 7.0
+     */
+    @XWalkAPI
+    public static final int LOAD_CACHE_ELSE_NETWORK = 1;
+
+    /**
+     * Don't use the cache, load from the network.
+     * Use with {@link #setCacheMode}.
+     * @since 7.0
+     */
+    @XWalkAPI
+    public static final int LOAD_NO_CACHE = 2;
+
+    /**
+     * Don't use the network, load from the cache.
+     * Use with {@link #setCacheMode}.
+     * @since 7.0
+     */
+    @XWalkAPI
+    public static final int LOAD_CACHE_ONLY = 3;
+
     private static final String TAG = "XWalkSettings";
 
     // This class must be created on the UI thread. Afterwards, it can be
@@ -250,8 +285,19 @@ public class XWalkSettingsInternal {
     }
 
     /**
-     * See {@link android.webkit.WebSettings#setCacheMode}.
+     * Overrides the way the cache is used. The way the cache is used is based
+     * on the navigation type. For a normal page load, the cache is checked
+     * and content is re-validated as needed. When navigating back, content is
+     * not revalidated, instead the content is just retrieved from the cache.
+     * This method allows the client to override this behavior by specifying
+     * one of {@link #LOAD_DEFAULT},
+     * {@link #LOAD_CACHE_ELSE_NETWORK}, {@link #LOAD_NO_CACHE} or
+     * {@link #LOAD_CACHE_ONLY}. The default value is {@link #LOAD_DEFAULT}.
+     *
+     * @param mode the mode to use
+     * @since 7.0
      */
+    @XWalkAPI
     public void setCacheMode(int mode) {
         synchronized (mXWalkSettingsLock) {
             if (mCacheMode != mode) {
@@ -261,8 +307,13 @@ public class XWalkSettingsInternal {
     }
 
     /**
-     * See {@link android.webkit.WebSettings#getCacheMode}.
+     * Gets the current setting for overriding the cache mode.
+     *
+     * @return the current setting for overriding the cache mode
+     * @see #setCacheMode
+     * @since 7.0
      */
+    @XWalkAPI
     public int getCacheMode() {
         synchronized (mXWalkSettingsLock) {
             return mCacheMode;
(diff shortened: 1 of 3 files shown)


xzhan96:
@lincsoon @fujunwei @hengzhi @axinging please help to review and merge, thanks

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
```diff
@@ -17,10 +19,80 @@
...
+        setCacheMode(XWalkSettings.LOAD_NO_CACHE);
+        loadUrlSync(url);
+        assertEquals(2, mWebServer.getRequestCount(htmlPath));
+        loadUrlSync(url);
+        assertEquals(3, mWebServer.getRequestCount(htmlPath));
+
+        setCacheMode(XWalkSettings.LOAD_CACHE_ONLY);
+        loadUrlSync(url);
+        assertEquals(3, mWebServer.getRequestCount(htmlPath));
+        loadUrlSync(url);
```
wuhengzhi: The second `loadUrlSync` is duplicated?

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkSettingsInternal.java
```diff
@@ -261,17 +307,38 @@ public void setCacheMode(int mode) {
...
+    @XWalkAPI
     public int getCacheMode() {
         synchronized (mXWalkSettingsLock) {
             return mCacheMode;
         }
     }
 
     /**
-     * See {@link android.webkit.WebSettings#setBlockNetworkLoads}.
+     * Sets whether the WebView should not load resources from the network.
```
wuhengzhi: `WebView` -> 'XWalkView'.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
```diff
@@ -17,10 +19,80 @@
...
+        assertEquals(3, mWebServer.getRequestCount(htmlPath));
+
+        loadUrlSyncAndExpectError(urlNotInCache);
+        assertEquals(0, mWebServer.getRequestCount(htmlNotInCachePath));
+    }
+
+    @MediumTest
+    @Feature({"Settings"})
+    // As our implementation of network loads blocking uses the same net::URLRequest settings, make
+    // sure that setting cache mode doesn't accidentally enable network loads.  The reference
```
wuhengzhi: Extra space before `The`.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
```diff
@@ -17,10 +19,80 @@
...
+    @Override
+    public void tearDown() throws Exception {
+        mWebServer.shutdown();
+        super.tearDown();
+    }
+
+    @MediumTest
+    @Feature({"Settings"})
+    public void testCacheMode() throws Throwable {
+        final String htmlPath = "/testCacheMode.html";
```
fujunwei: Did the file of testing commit?

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
```diff
@@ -17,10 +19,80 @@
...
+        final String url = mWebServer.setResponse(htmlPath, "response", null);
+        final String htmlNotInCachePath = "/testCacheMode-not-in-cache.html";
+        final String urlNotInCache = mWebServer.setResponse(htmlNotInCachePath, "", null);
+
+        clearCacheOnUiThread(true);
+        assertEquals(XWalkSettings.LOAD_DEFAULT, getCacheMode());
+
+        setCacheMode(XWalkSettings.LOAD_CACHE_ELSE_NETWORK);
+        loadUrlSync(url);
+        assertEquals(1, mWebServer.getRequestCount(htmlPath));
```
fujunwei: What's the meaning of "1, 2,3"?

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
```diff
@@ -17,10 +19,80 @@
...
+
+        setCacheMode(XWalkSettings.LOAD_CACHE_ELSE_NETWORK);
+        loadUrlSyncAndExpectError(url);
+        assertEquals(0, mWebServer.getRequestCount(htmlPath));
+
+        setCacheMode(XWalkSettings.LOAD_NO_CACHE);
+        loadUrlSyncAndExpectError(url);
+        assertEquals(0, mWebServer.getRequestCount(htmlPath));
+
+        setCacheMode(XWalkSettings.LOAD_CACHE_ONLY);
```
fujunwei: The function of `clearCacheOnUiThread` cleared the cache, do the test case of  LOAD_CACHE_ONLY need  which don't load resources from network?

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
xzhan96: No, it's necessary.
LOAD_CACHE_ELSE_NETWORK means "use cached resources when they are available, even if they have expired. Otherwise load resources from the network."
So after the 1st loadUrlSync(page cached), we should verify the 2nd loadUrlSync is from cache instead of network.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
xzhan96: Thanks

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
xzhan96: All the the testing files such as "testCacheMode.html" is non-existent, it has effect because the set up of `mWebServer.setResponse(htmlPath, "response", null)`. Actually I also need study the Chromium test framework, can't explain more clear.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
wuhengzhi: But the current cache mode was set to `LOAD_CACHE_ONLY` instead of `LOAD_CACHE_ELSE_NETWORK`.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
xzhan96: This number is depends on how many request to TestWebServer.
For example, clear all the previous cache by `clearCacheOnUiThread`, then set the cache mode to LOAD_CACHE_ELSE_NETWORK, which means `use cached resources when they are available, even if they have expired. Otherwise load resources from the network.`, and then after the loadUrlSync, **one time request** to web server is needed because **no cache currently**.  And then, **the page is cached**, so after the 2nd loadUrlSync, as the previous **cache mode setting**, there is **no need to request** webserver, so the mWebServer.getRequestCount should also return 1, just as the same before 2nd loadUrlSync.
Others and so on..

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
xzhan96: Just like previous 2 times loadUrlSync and 2 times assertEqual judge after each time cache mode setting, we should ensure all the settings are always work instead of just work one time.

test/android/core/javatests/src/org/xwalk/core/xwview/test/SettingsTest.java
xzhan96: `clearCacheOnUiThread` is just clear all the caches in other test, it's better to verify all the settings here.

xzhan96:
@lincsoon please help to merge, thanks

xzhan96:
@lincsoon please help to merge, thanks

PR #3775, opened 2016-07-01
crosswalk-20: DEPS.xwalk: Explicitly set `managed` to True in .gclient-xwalk.

rakuco:
https://codereview.chromium.org/2099153003 has recently switched the
default value of the `managed` option to False instead of True.

Since we did not explicitly set that option in `.gclient-xwalk` before,
it means that for the last few days existing `src/` checkouts were not
being automatically updated to match `chromium_crosswalk_rev` any
longer, which was particularly bad for the bots, which were not testing
the commits they should.

(cherry picked from commit e9f19a987c604f30b46420fcbdb679e5f977fb98)


pozdnyakov:
lgtm

PR #3776, opened 2016-07-02
Fix the video play issue after M51 rebase.

wuhengzhi:
This issue was caused by https://codereview.chromium.org/1577433004
The condition of 'OS=="android"' was removed, this leads to
'proprietary_codecs' was always 0, then mp4 and other codecs could not
be added to related 'MediaFormatMappings'.

'proprietary_codecs' was defined in the condition in src/build/common.gypi,
so in src/xwalk/common.gypi, 'proprietary_codecs' should also be moved
to condition to keep same scope level.

BUG=XWALK-7032

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -114,6 +114,12 @@
         ['OS=="win"', {
           'use_rssdk%': 1,
         }],
+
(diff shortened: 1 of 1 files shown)


wuhengzhi:
@rakuco @lincsoon PTAL, thanks.

PR #3777, opened 2016-07-04
[Android] Fix the potential vulnerability issue.

wuhengzhi:
For SSL errors, if user allows it, native layer will keep the host and
the error num, the future communications with same host and error will
not trigger the onReceivedSslError callback. Even if the certificate is
invalid, all of requests will also be passed, its vulnerable for some
serious SSL error.
Especially with MITM, the sensitive messages will be stolen, e.g.
passwords, credit cards.
Deny the invalid requests for some serious SSL error to avoid this issue.

For keeping the host and error num, please refer to
https://codereview.chromium.org/794023002
We also have the same implementation.

BUG=XWALK-6986

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/SslUtil.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/SslUtil.java
@@ -24,17 +24,7 @@ class SslUtil {
      */
     public static SslError sslErrorFromNetErrorCode(int error, SslCertificate cert, String url) {
         assert (error >= NetError.ERR_CERT_END && error <= NetError.ERR_CERT_COMMON_NAME_INVALID);
-        switch(error) {
-            case NetError.ERR_CERT_COMMON_NAME_INVALID:
-                return new SslError(SslError.SSL_IDMISMATCH, cert, url);
-            case NetError.ERR_CERT_DATE_INVALID:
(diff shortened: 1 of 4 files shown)


wuhengzhi:
@lincsoon @rakuco @baleboy @xzhan96 @axinging @fujunwei PTAL, thanks.

baleboy:
@terriko can you take a look?

vincentbernat:
This completely break the ability for an application to handle SSL errors, notably self-signed certificates or custom certificate handling. See https://crosswalk-project.org/jira/browse/XWALK-7429. On the browser, people are exposed to the same problem: once they accept an invalid certificate, anything can happen: certificate may change, certificate may expire. I don't see why the restrictions should be more important inside the webview.

vincentbernat:
The vulneraibility could have been fixed by changing the default implementation of `onReceivedSslError()` to:

```java
Toast.makeText(mXWalkView.getContext(), R.string.ssl_error_deny_request,
                    Toast.LENGTH_SHORT).show();
callback.onReceiveValue(false);
```

This would match the behavior of Android's own webview and still enable the user to override this with its own code and display more meaningful messages.

rakuco:
@vincentbernat: thanks for the investigation. Would you be interested in sending a pull request with the proposed changes?

vincentbernat:
Yes, I can do that. Give me a few days.

PR #3781, opened 2016-07-06
build: Generalize the |proprietary_codecs| check from 7db2189.

rakuco:
Commit 7db2189 ("Fix the video play issue after M51 rebase") correctly
mentioned the reason for moving the `proprietary_codecs` assignment to a
`conditions` block, but needlessly checked for the `OS` variable instead
of just adding a check that is always true to better simulate what we
were doing before (an unconditional assignment).

Since gyp's evaluation rules are very confusing, also add more comments
to `common.gypi` explaining what is really going on.

BUG=XWALK-7032

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -9,6 +9,19 @@
 # called. This file is always included immediately after src/build/common.gypi
 # itself so that variables originally set there can have their values
 # overridden here.
+#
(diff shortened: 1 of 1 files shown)


rakuco:
@wuhengzhi @lincsoon

wuhengzhi:
LGTM, thanks.

PR #3782, opened 2016-07-06
[Linux][Windows] Implement navigation history of links.

darktears:
So they can change color appropriately if you already navigated that URL.
Mostly borrowing the work already done for Android.

BUG=XWALK-6935

(cherry picked from commit 98d737fb480d8f718702df6cb00d37cf1050f76f)


darktears:
@rakuco PTAL

PR #3786, opened 2016-07-07
[Windows][Linux] Fix Geolocation crashing after the user accept the

darktears:
prompt.

Bug is a regression after M51 and incorrect patch of
8d0c7a657433574edc397e2dbebe8dcc690e45c1 which removes the functions
CreateRequestContext and CreateRequestContextForStoragePartition from
XWalkContentBrowserClient as they moved to BrowserContext (so
XWalkBrowserContext). The problem is that they were setting the
url_request_context_getter_ member before calling XWalkBrowserContext.
The URL request context is then used afterwards inside xwalk to query
things like geolocation. With that accidental removal it was always null
so it was crashing/asserting later in the geolocation code. Fix is to
setup a getter from the browser context to access the URL request context
when creating the AccessTokenStore.

BUG=XWALK-7067

### DIFF
--- a/runtime/browser/xwalk_browser_context.h
+++ b/runtime/browser/xwalk_browser_context.h
@@ -20,16 +20,13 @@
 #include "components/visitedlink/browser/visitedlink_delegate.h"
 #include "content/public/browser/browser_context.h"
 #include "content/public/browser/content_browser_client.h"
+#include "xwalk/runtime/browser/runtime_url_request_context_getter.h"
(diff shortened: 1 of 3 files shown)


darktears:
@rakuco @pozdnyakov PTAL

rakuco:
@mrunalk FYI &

rakuco:
le lgtm :fr:

mrunalk:
@darktears Looking at this PR makes more sense. Thanks for nailing this down and fixing it. I will try to be extra careful when rebasing such changes in future.

PR #3791, opened 2016-07-08
[Android] Reset the default value of setDomStorageEnabled

xzhan96:
As a frequently used feature, it is also enabled in Chrome in default.
And change the value may cause potential risk, so reset it to original.

BUG=XWALK-7082


xzhan96:
@lincsoon I just wanted to make the default value the same with WebView, but it will cause some test cases fail, so reset it. Please help to merge.

PR #3795, opened 2016-07-11
[Android] Correct some mistakes in XWalkSettings API annotation

xzhan96:
Correct some mistakes in XWalkSettings API annotation.


xzhan96:
@lincsoon

PR #3797, opened 2016-07-11
M52: correcting SHA1 in DEPS.xwalk after reverting SIMD.js in v8 

msisov:


DEPS.xwalk
```diff
@@ -17,8 +17,8 @@
 # Edit these when rolling DEPS.xwalk.
 # -----------------------------------
 
-chromium_crosswalk_rev = 'd216c2f7d8f2a57ffe0b12290c9bf3f0fbe23630'
-v8_crosswalk_rev = 'c944c4577bd341cadd426633683becab280c8c12'
+chromium_crosswalk_rev = 'd14f6d5c51be7d13de6208721df65930d9345019'
```
rakuco: The change to `chromium_crosswalk_rev` looks wrong, you're pointing at an old hash.

msisov: yeah, right. I've forgotten Mrunal has reverted WebCL. Sorry, here is corrected DEPS.xwalk file.

rakuco: It still looks wrong. `next` currently points to 1a3e4b853512415b4333e7e7e012b9a8ff3ec98e ("M52 : Fix build on Windows"). If this is only for the V8 hashes, you should only need to change `v8_crosswalk_rev`.

mrunalk:
I have just pushed the correct hashes so I don't think we need this patch anymore.

PR #3799, opened 2016-07-12
[Android] Reset the default value of setDomStorageEnabled

xzhan96:
As a frequently used feature, it is also enabled in Chrome in default.
And change the value may cause potential risk, so reset it to original.

BUG=XWALK-7082

(cherry picked from commit 55834b2)


xzhan96:
@rakuco please help to merge into Crosswalk 21, thanks

xzhan96:
@rakuco 
these two commits are important, for the first one `dc444e3` is to fix the P1 issue XWALK-7082, and second one `b37c5ed` to correct some mistakes in API annotation. Please help to merge, thanks.

PR #3802, opened 2016-07-12
[Windows][Linux] DevTools: Fix Maximum call stack size exceeded error

darktears:
while retrieving CPU profile.

Do not pass object literals into dispatchMessage API. Always turn it into
a string.


darktears:
@rakuco

rakuco:
le lgtm :soccer:

PR #3806, opened 2016-07-13
[Windows] Autostart application if no paramaters are passed to xwalk.

darktears:
Automatically try to find the manifest.json relative to the xwalk.exe path
in approot/ (which is the place for application files created using
crosswalk-app-tools). If it's not found then crosswalk doesn't start.


darktears:
@pozdnyakov PTAL it goes with https://github.com/crosswalk-project/crosswalk-app-tools/pull/144

darktears:
@pozdnyakov landed the app-tool part.

darktears:
@pozdnyakov ping

rakuco:
I think Misha is on holidays. rs=me

PR #3819, opened 2016-07-25
Add gn build support for crosswalk-linux

heke123:
As Chromium will deprecate GYP build system since M54, we need to enable
GN build support for Crosswalk as soon as possible.
   
This PR only implements most part of Crosswalk-Linux.
See comments in xwalk/build/linux.gni on how to build.
    
TODO: Add web_notification, installer support

BUG=XWALK-7180

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -1,0 +1,489 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/features.gni")
+import("//build/config/ui.gni")
+import("//xwalk/build/version.gni")
+import("//xwalk/build/xwalk.gni")
+
+group("xwalk_builder") {
+  # In GN, When a target is marked "testonly = true", it must only be depended
+  # on by other test-only targets. In crosswalk project we don't need to care
+  # which targets are testonly. In order to group all targets into "xwalk_builder"
+  # we just set testonly=true.
+  testonly = true
+  deps = []
+  if (is_linux) {
+    deps += [
+      ":xwalk",
+      ":xwalk_all_tests",
+    ]
+  }
+}
+
+executable("xwalk") {
+  sources = [
+    "runtime/app/xwalk_main.cc",
+  ]
+  deps = [
+    ":xwalk_runtime",
+  ]
+}
+
+source_set("xwalk_runtime") {
+  sources = [
+    "//extensions/common/constants.cc",
+    "//extensions/common/constants.h",
+    "//extensions/common/url_pattern.cc",
+    "//extensions/common/url_pattern.h",
+    "runtime/app/android/xwalk_main_delegate_android.cc",
+    "runtime/app/android/xwalk_main_delegate_android.h",
+    "runtime/app/xwalk_main_delegate.cc",
+    "runtime/app/xwalk_main_delegate.h",
+    "runtime/browser/android/cookie_manager.cc",
+    "runtime/browser/android/cookie_manager.h",
+    "runtime/browser/android/find_helper.cc",
+    "runtime/browser/android/find_helper.h",
+    "runtime/browser/android/net/android_protocol_handler.cc",
+    "runtime/browser/android/net/android_protocol_handler.h",
+    "runtime/browser/android/net/android_stream_reader_url_request_job.cc",
+    "runtime/browser/android/net/android_stream_reader_url_request_job.h",
+    "runtime/browser/android/net/init_native_callback.h",
+    "runtime/browser/android/net/input_stream.h",
+    "runtime/browser/android/net/input_stream_impl.cc",
+    "runtime/browser/android/net/input_stream_impl.h",
+    "runtime/browser/android/net/input_stream_reader.cc",
+    "runtime/browser/android/net/input_stream_reader.h",
+    "runtime/browser/android/net/url_constants.cc",
+    "runtime/browser/android/net/url_constants.h",
+    "runtime/browser/android/net/xwalk_cookie_store_wrapper.cc",
+    "runtime/browser/android/net/xwalk_cookie_store_wrapper.h",
+    "runtime/browser/android/net/xwalk_url_request_job_factory.cc",
+    "runtime/browser/android/net/xwalk_url_request_job_factory.h",
+    "runtime/browser/android/net_disk_cache_remover.cc",
+    "runtime/browser/android/net_disk_cache_remover.h",
+    "runtime/browser/android/renderer_host/xwalk_render_view_host_ext.cc",
+    "runtime/browser/android/renderer_host/xwalk_render_view_host_ext.h",
+    "runtime/browser/android/scoped_allow_wait_for_legacy_web_view_api.h",
+    "runtime/browser/android/state_serializer.cc",
+    "runtime/browser/android/state_serializer.h",
+    "runtime/browser/android/xwalk_autofill_client_android.cc",
+    "runtime/browser/android/xwalk_autofill_client_android.h",
+    "runtime/browser/android/xwalk_content.cc",
+    "runtime/browser/android/xwalk_content.h",
+    "runtime/browser/android/xwalk_content_lifecycle_notifier.cc",
+    "runtime/browser/android/xwalk_content_lifecycle_notifier.h",
+    "runtime/browser/android/xwalk_contents_client_bridge.cc",
+    "runtime/browser/android/xwalk_contents_client_bridge.h",
+    "runtime/browser/android/xwalk_contents_client_bridge_base.cc",
+    "runtime/browser/android/xwalk_contents_client_bridge_base.h",
+    "runtime/browser/android/xwalk_contents_io_thread_client.h",
+    "runtime/browser/android/xwalk_contents_io_thread_client_impl.cc",
+    "runtime/browser/android/xwalk_contents_io_thread_client_impl.h",
+    "runtime/browser/android/xwalk_cookie_access_policy.cc",
+    "runtime/browser/android/xwalk_cookie_access_policy.h",
+    "runtime/browser/android/xwalk_dev_tools_server.cc",
+    "runtime/browser/android/xwalk_dev_tools_server.h",
+    "runtime/browser/android/xwalk_download_resource_throttle.cc",
+    "runtime/browser/android/xwalk_download_resource_throttle.h",
+    "runtime/browser/android/xwalk_http_auth_handler.cc",
+    "runtime/browser/android/xwalk_http_auth_handler.h",
+    "runtime/browser/android/xwalk_http_auth_handler_base.cc",
+    "runtime/browser/android/xwalk_http_auth_handler_base.h",
+    "runtime/browser/android/xwalk_icon_helper.cc",
+    "runtime/browser/android/xwalk_icon_helper.h",
+    "runtime/browser/android/xwalk_login_delegate.cc",
+    "runtime/browser/android/xwalk_login_delegate.h",
+    "runtime/browser/android/xwalk_path_helper.cc",
+    "runtime/browser/android/xwalk_path_helper.h",
+    "runtime/browser/android/xwalk_presentation_host.cc",
+    "runtime/browser/android/xwalk_presentation_host.h",
+    "runtime/browser/android/xwalk_request_interceptor.cc",
+    "runtime/browser/android/xwalk_request_interceptor.h",
+    "runtime/browser/android/xwalk_settings.cc",
+    "runtime/browser/android/xwalk_view_delegate.cc",
+    "runtime/browser/android/xwalk_view_delegate.h",
+    "runtime/browser/android/xwalk_web_contents_delegate.cc",
+    "runtime/browser/android/xwalk_web_contents_delegate.h",
+    "runtime/browser/android/xwalk_web_contents_view_delegate.cc",
+    "runtime/browser/android/xwalk_web_contents_view_delegate.h",
+    "runtime/browser/android/xwalk_web_resource_response.cc",
+    "runtime/browser/android/xwalk_web_resource_response.h",
+    "runtime/browser/android/xwalk_web_resource_response_impl.cc",
+    "runtime/browser/android/xwalk_web_resource_response_impl.h",
+    "runtime/browser/application_component.cc",
+    "runtime/browser/application_component.h",
+    "runtime/browser/devtools/remote_debugging_server.cc",
+    "runtime/browser/devtools/remote_debugging_server.h",
+    "runtime/browser/devtools/xwalk_devtools_manager_delegate.cc",
+    "runtime/browser/devtools/xwalk_devtools_manager_delegate.h",
+    "runtime/browser/geolocation/xwalk_access_token_store.cc",
+    "runtime/browser/geolocation/xwalk_access_token_store.h",
+    "runtime/browser/image_util.cc",
+    "runtime/browser/image_util.h",
+    "runtime/browser/media/media_capture_devices_dispatcher.cc",
+    "runtime/browser/media/media_capture_devices_dispatcher.h",
+    "runtime/browser/runtime.cc",
+    "runtime/browser/runtime.h",
+    "runtime/browser/runtime_download_manager_delegate.cc",
+    "runtime/browser/runtime_download_manager_delegate.h",
+    "runtime/browser/runtime_file_select_helper.cc",
+    "runtime/browser/runtime_file_select_helper.h",
+    "runtime/browser/runtime_geolocation_permission_context.cc",
+    "runtime/browser/runtime_geolocation_permission_context.h",
+    "runtime/browser/runtime_javascript_dialog_manager.cc",
+    "runtime/browser/runtime_javascript_dialog_manager.h",
+    "runtime/browser/runtime_network_delegate.cc",
+    "runtime/browser/runtime_network_delegate.h",
+    "runtime/browser/runtime_notification_permission_context.cc",
+    "runtime/browser/runtime_notification_permission_context.h",
+    "runtime/browser/runtime_platform_util.h",
+    "runtime/browser/runtime_platform_util_android.cc",
+    "runtime/browser/runtime_platform_util_linux.cc",
+    "runtime/browser/runtime_platform_util_mac.mm",
+    "runtime/browser/runtime_platform_util_win.cc",
+    "runtime/browser/runtime_quota_permission_context.cc",
+    "runtime/browser/runtime_quota_permission_context.h",
+    "runtime/browser/runtime_resource_dispatcher_host_delegate.cc",
+    "runtime/browser/runtime_resource_dispatcher_host_delegate.h",
+    "runtime/browser/runtime_resource_dispatcher_host_delegate_android.cc",
+    "runtime/browser/runtime_resource_dispatcher_host_delegate_android.h",
+    "runtime/browser/runtime_select_file_policy.cc",
+    "runtime/browser/runtime_select_file_policy.h",
+    "runtime/browser/runtime_ui_delegate.cc",
+    "runtime/browser/runtime_ui_delegate.h",
+    "runtime/browser/runtime_url_request_context_getter.cc",
+    "runtime/browser/runtime_url_request_context_getter.h",
+    "runtime/browser/speech/speech_recognition_manager_delegate.cc",
+    "runtime/browser/speech/speech_recognition_manager_delegate.h",
+    "runtime/browser/ssl_error_page.cc",
+    "runtime/browser/ssl_error_page.h",
+    "runtime/browser/storage_component.cc",
+    "runtime/browser/storage_component.h",
+    "runtime/browser/sysapps_component.cc",
+    "runtime/browser/sysapps_component.h",
+    "runtime/browser/ui/color_chooser.cc",
+    "runtime/browser/ui/color_chooser.h",
+    "runtime/browser/ui/color_chooser_android.cc",
+    "runtime/browser/ui/color_chooser_mac.cc",
+    "runtime/browser/ui/desktop/exclusive_access_bubble.cc",
+    "runtime/browser/ui/desktop/exclusive_access_bubble.h",
+    "runtime/browser/ui/desktop/exclusive_access_bubble_views_context.h",
+    "runtime/browser/ui/native_app_window.cc",
+    "runtime/browser/ui/native_app_window.h",
+    "runtime/browser/ui/native_app_window_android.cc",
+    "runtime/browser/ui/native_app_window_mac.h",
+    "runtime/browser/ui/native_app_window_mac.mm",
+    "runtime/browser/xwalk_app_extension_bridge.cc",
+    "runtime/browser/xwalk_app_extension_bridge.h",
+    "runtime/browser/xwalk_application_mac.h",
+    "runtime/browser/xwalk_application_mac.mm",
+    "runtime/browser/xwalk_autofill_client.cc",
+    "runtime/browser/xwalk_autofill_client.h",
+    "runtime/browser/xwalk_autofill_manager.cc",
+    "runtime/browser/xwalk_autofill_manager.h",
+    "runtime/browser/xwalk_browser_context.cc",
+    "runtime/browser/xwalk_browser_context.h",
+    "runtime/browser/xwalk_browser_main_parts.cc",
+    "runtime/browser/xwalk_browser_main_parts.h",
+    "runtime/browser/xwalk_browser_main_parts_android.cc",
+    "runtime/browser/xwalk_browser_main_parts_android.h",
+    "runtime/browser/xwalk_browser_main_parts_mac.h",
+    "runtime/browser/xwalk_browser_main_parts_mac.mm",
+    "runtime/browser/xwalk_component.h",
+    "runtime/browser/xwalk_content_browser_client.cc",
+    "runtime/browser/xwalk_content_browser_client.h",
+    "runtime/browser/xwalk_content_settings.cc",
+    "runtime/browser/xwalk_content_settings.h",
+    "runtime/browser/xwalk_form_database_service.cc",
+    "runtime/browser/xwalk_form_database_service.h",
+    "runtime/browser/xwalk_permission_manager.cc",
+    "runtime/browser/xwalk_permission_manager.h",
+    "runtime/browser/xwalk_platform_notification_service.cc",
+    "runtime/browser/xwalk_platform_notification_service.h",
+    "runtime/browser/xwalk_pref_store.cc",
+    "runtime/browser/xwalk_pref_store.h",
+    "runtime/browser/xwalk_presentation_service_delegate.cc",
+    "runtime/browser/xwalk_presentation_service_delegate.h",
+    "runtime/browser/xwalk_presentation_service_delegate_android.cc",
+    "runtime/browser/xwalk_presentation_service_delegate_android.h",
+    "runtime/browser/xwalk_presentation_service_delegate_win.cc",
+    "runtime/browser/xwalk_presentation_service_delegate_win.h",
+    "runtime/browser/xwalk_presentation_service_helper.cc",
+    "runtime/browser/xwalk_presentation_service_helper.h",
+    "runtime/browser/xwalk_presentation_service_helper_android.cc",
+    "runtime/browser/xwalk_presentation_service_helper_android.h",
+    "runtime/browser/xwalk_presentation_service_helper_win.cc",
+    "runtime/browser/xwalk_presentation_service_helper_win.h",
+    "runtime/browser/xwalk_render_message_filter.cc",
+    "runtime/browser/xwalk_render_message_filter.h",
+    "runtime/browser/xwalk_runner.cc",
+    "runtime/browser/xwalk_runner.h",
+    "runtime/browser/xwalk_runner_win.cc",
+    "runtime/browser/xwalk_runner_win.h",
+    "runtime/browser/xwalk_ssl_host_state_delegate.cc",
+    "runtime/browser/xwalk_ssl_host_state_delegate.h",
+    "runtime/common/android/xwalk_globals_android.cc",
+    "runtime/common/android/xwalk_globals_android.h",
+    "runtime/common/android/xwalk_hit_test_data.cc",
+    "runtime/common/android/xwalk_hit_test_data.h",
+    "runtime/common/android/xwalk_message_generator.cc",
+    "runtime/common/android/xwalk_message_generator.h",
+    "runtime/common/android/xwalk_render_view_messages.cc",
+    "runtime/common/android/xwalk_render_view_messages.h",
+    "runtime/common/logging_xwalk.cc",
+    "runtime/common/logging_xwalk.h",
+    "runtime/common/paths_mac.h",
+    "runtime/common/paths_mac.mm",
+    "runtime/common/xwalk_common_message_generator.cc",
+    "runtime/common/xwalk_common_message_generator.h",
+    "runtime/common/xwalk_common_messages.cc",
+    "runtime/common/xwalk_common_messages.h",
+    "runtime/common/xwalk_content_client.cc",
+    "runtime/common/xwalk_content_client.h",
+    "runtime/common/xwalk_localized_error.cc",
+    "runtime/common/xwalk_localized_error.h",
+    "runtime/common/xwalk_paths.cc",
+    "runtime/common/xwalk_paths.h",
+    "runtime/common/xwalk_resource_delegate.cc",
+    "runtime/common/xwalk_resource_delegate.h",
+    "runtime/common/xwalk_runtime_features.cc",
+    "runtime/common/xwalk_runtime_features.h",
+    "runtime/common/xwalk_switches.cc",
+    "runtime/common/xwalk_switches.h",
+    "runtime/common/xwalk_system_locale.cc",
+    "runtime/common/xwalk_system_locale.h",
+    "runtime/renderer/android/xwalk_permission_client.cc",
+    "runtime/renderer/android/xwalk_permission_client.h",
+    "runtime/renderer/android/xwalk_render_thread_observer.cc",
+    "runtime/renderer/android/xwalk_render_thread_observer.h",
+    "runtime/renderer/android/xwalk_render_view_ext.cc",
+    "runtime/renderer/android/xwalk_render_view_ext.h",
+    "runtime/renderer/isolated_file_system.cc",
+    "runtime/renderer/isolated_file_system.h",
+    "runtime/renderer/xwalk_content_renderer_client.cc",
+    "runtime/renderer/xwalk_content_renderer_client.h",
+  ]
+  if (!is_android) {
+    sources += [
+      "runtime/browser/devtools/xwalk_devtools_frontend.cc",
+      "runtime/browser/devtools/xwalk_devtools_frontend.h",
+      "runtime/browser/runtime_ui_delegate_desktop.cc",
+      "runtime/browser/runtime_ui_delegate_desktop.h",
+      "runtime/browser/ui/desktop/xwalk_autofill_popup_controller.cc",
+      "runtime/browser/ui/desktop/xwalk_autofill_popup_controller.h",
+      "runtime/browser/ui/desktop/xwalk_autofill_popup_view.cc",
+      "runtime/browser/ui/desktop/xwalk_autofill_popup_view.h",
+      "runtime/browser/ui/desktop/xwalk_permission_dialog_manager.cc",
+      "runtime/browser/ui/desktop/xwalk_permission_dialog_manager.h",
+      "runtime/browser/ui/desktop/xwalk_permission_modal_dialog.cc",
+      "runtime/browser/ui/desktop/xwalk_permission_modal_dialog.h",
+      "runtime/browser/ui/desktop/xwalk_popup_controller.cc",
+      "runtime/browser/ui/desktop/xwalk_popup_controller.h",
+      "runtime/browser/ui/native_app_window_desktop.cc",
+      "runtime/browser/ui/native_app_window_desktop.h",
+      "runtime/browser/xwalk_autofill_client_desktop.cc",
+      "runtime/browser/xwalk_autofill_client_desktop.h",
+      "runtime/renderer/xwalk_render_thread_observer_generic.cc",
+      "runtime/renderer/xwalk_render_thread_observer_generic.h",
+    ]
+  }
+  if (is_linux && !is_component_build && is_component_ffmpeg) {
+    # Set RPath for executable xwalk, xwalk_browsertest and xwalk_unittest.
+    public_configs = [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
+  }
+
+  include_dirs = [ "$root_gen_dir/xwalk" ]
+  defines = [
+    "CHROME_VERSION=\"$chrome_version\"",
+    "XWALK_VERSION=\"$xwalk_version\"",
+  ]
+  deps = [
+    ":generate_upstream_blink_version",
+    "//base",
+    "//base:base_static",
+    "//base:i18n",
+    "//base/third_party/dynamic_annotations",
+    "//cc",
+    "//components/autofill/content/browser",
+    "//components/autofill/content/renderer",
+    "//components/autofill/core/browser",
+    "//components/cdm/renderer",
+    "//components/devtools_http_handler",
+    "//components/resources:components_resources",
+    "//components/strings",
+    "//components/url_formatter",
+    "//components/user_prefs",
+    "//components/visitedlink/browser",
+    "//components/visitedlink/renderer",
+    "//content",
+    "//content/public/app:both",
+    "//content/public/browser",
+    "//content/public/child",
+    "//content/public/common",
+    "//content/public/utility",
+    "//gin",
+    "//ipc",
+    "//media",
+    "//net",
+    "//net:net_resources",
+    "//skia",
+    "//storage/browser",
+    "//storage/common",
+    "//third_party/WebKit/public:blink",
+    "//third_party/boringssl",
+    "//ui/base",
+    "//ui/gl",
+    "//ui/shell_dialogs",
+    "//ui/snapshot",
+    "//url",
+    "//v8",
+    "//xwalk/application:xwalk_application_lib",
+    "//xwalk/application:xwalk_application_resources",
+    "//xwalk/experimental",
+    "//xwalk/extensions",
+    "//xwalk/resources:generate_xwalk_resources",
+    "//xwalk/resources:xwalk_pak",
+    "//xwalk/resources:xwalk_resources_100_percent_pak",
+    "//xwalk/resources:xwalk_resources_200_percent_pak",
+    "//xwalk/resources:xwalk_resources_300_percent_pak",
+    "//xwalk/resources:xwalk_strings",
+    "//xwalk/sysapps",
+    "//xwalk/sysapps:xwalk_sysapps_resources",
+  ]
+
+  if (is_linux) {
+    deps += [ "//xwalk/resources:xwalk_locales" ]
+  }
+
+  # TODO(heke123): Add WebNotification for Linux
+  # WebNotification needs libgtk2ui and libnotify.
+
+  if (disable_bundled_extensions) {
+    defines += [ "DISABLE_BUNDLED_EXTENSIONS" ]
+  }
+
+  if (enable_nacl) {
+    if (is_linux) {
+      sources += [
+        "runtime/browser/nacl_host/nacl_browser_delegate_impl.cc",
+        "runtime/browser/nacl_host/nacl_browser_delegate_impl.h",
+      ]
+    }
+    deps += [
+      "//components/nacl/browser",
+      "//components/nacl/common",
+      "//components/nacl/loader",
+      "//components/nacl/loader:nacl_helper",
+      "//components/nacl/loader/sandbox_linux",
+      "//components/nacl/renderer",
+      "//components/nacl/renderer/plugin:nacl_trusted_plugin",
+      "//native_client/src/trusted/service_runtime/linux:bootstrap",
+    ]
+  }
+
+  if (enable_plugins) {
+    sources += [
+      "//chrome/renderer/pepper/pepper_flash_drm_renderer_host.cc",
+      "//chrome/renderer/pepper/pepper_flash_drm_renderer_host.h",
+      "//chrome/renderer/pepper/pepper_flash_font_file_host.cc",
+      "//chrome/renderer/pepper/pepper_flash_font_file_host.h",
+      "//chrome/renderer/pepper/pepper_flash_fullscreen_host.cc",
+      "//chrome/renderer/pepper/pepper_flash_fullscreen_host.h",
+      "//chrome/renderer/pepper/pepper_flash_menu_host.cc",
+      "//chrome/renderer/pepper/pepper_flash_menu_host.h",
+      "//chrome/renderer/pepper/pepper_flash_renderer_host.cc",
+      "//chrome/renderer/pepper/pepper_flash_renderer_host.h",
+      "runtime/browser/renderer_host/pepper/xwalk_browser_pepper_host_factory.cc",
+      "runtime/browser/renderer_host/pepper/xwalk_browser_pepper_host_factory.h",
+      "runtime/renderer/pepper/pepper_helper.cc",
+      "runtime/renderer/pepper/pepper_helper.h",
+      "runtime/renderer/pepper/pepper_uma_host.cc",
+      "runtime/renderer/pepper/pepper_uma_host.h",
+      "runtime/renderer/pepper/xwalk_renderer_pepper_host_factory.cc",
+      "runtime/renderer/pepper/xwalk_renderer_pepper_host_factory.h",
+    ]
+    deps += [
+      "//components/pdf/renderer",
+      "//ppapi/host",
+      "//ppapi/proxy",
+      "//ppapi/proxy:ipc",
+      "//ppapi/shared_impl",
+    ]
+  }
+
+  if (toolkit_views) {
+    sources += [
+      "runtime/browser/ui/desktop/download_views.cc",
+      "runtime/browser/ui/desktop/download_views.h",
+      "runtime/browser/ui/desktop/exclusive_access_bubble_views.cc",
+      "runtime/browser/ui/desktop/exclusive_access_bubble_views.h",
+      "runtime/browser/ui/desktop/xwalk_permission_modal_dialog_views.cc",
+      "runtime/browser/ui/desktop/xwalk_permission_modal_dialog_views.h",
+      "runtime/browser/ui/native_app_window_views.cc",
+      "runtime/browser/ui/native_app_window_views.h",
+      "runtime/browser/ui/top_view_layout_views.cc",
+      "runtime/browser/ui/top_view_layout_views.h",
+      "runtime/browser/ui/xwalk_javascript_native_dialog_factory.h",
+      "runtime/browser/ui/xwalk_javascript_native_dialog_factory_views.cc",
+      "runtime/browser/ui/xwalk_views_delegate.cc",
+      "runtime/browser/ui/xwalk_views_delegate.h",
+    ]
+    deps += [
+      "//components/app_modal",
+      "//components/constrained_window",
+      "//components/guest_view/browser",
+      "//components/ui/zoom",
+      "//ui/events",
+      "//ui/resources",
+      "//ui/strings:ui_strings",
+      "//ui/views",
+      "//ui/views/controls/webview",
+    ]
+  }
+
+  if (use_aura) {
+    sources += [
+      "runtime/browser/runtime_platform_util_aura.cc",
+      "runtime/browser/ui/color_chooser_aura.cc",
+    ]
+    deps += [
+      "//ui/aura",
+      "//ui/base/ime",
+      "//ui/wm",
+    ]
+  }
+}
+
+# While we could just call lastchange.py here and generate the header
+# directly, it would only work if there is a git checkout (ie. it does
+# not work with a tarball, for example).
+action("generate_upstream_blink_version") {
+  output_file = "$root_gen_dir/blink_upstream_version.h"
+  template = "//xwalk/runtime/browser/blink_upstream_version.h.in"
+  upstream = "//xwalk/build/UPSTREAM.blink"
+
+  outputs = [
+    "$output_file",
+  ]
+  script = "//build/util/version.py"
+  args = [
+    "-f",
+    rebase_path(upstream, root_out_dir),
+    rebase_path(template, root_out_dir),
+    rebase_path(output_file, root_out_dir),
+  ]
+}
+
+group("xwalk_all_tests") {
+  testonly = true
+  deps = [
+    "//xwalk/extensions/test:xwalk_extensions_browsertest",
+    "//xwalk/extensions/test:xwalk_extensions_unittest",
+    "//xwalk/sysapps:xwalk_sysapps_browsertest",
+    "//xwalk/sysapps:xwalk_sysapps_unittest",
+    "//xwalk/test:xwalk_browsertest",
+    "//xwalk/test:xwalk_unittest",
+  ]
+}
--- a/application/common/BUILD.gn
+++ b/application/common/BUILD.gn
@@ -1,0 +1,51 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+source_set("xwalk_application_common_lib") {
+  sources = [
+    "application_data.cc",
+    "application_data.h",
+    "application_file_util.cc",
+    "application_file_util.h",
+    "application_manifest_constants.cc",
+    "application_manifest_constants.h",
+    "application_resource.cc",
+    "application_resource.h",
+    "constants.cc",
+    "constants.h",
+    "id_util.cc",
+    "id_util.h",
+    "manifest.cc",
+    "manifest.h",
+    "manifest_handler.cc",
+    "manifest_handler.h",
+    "manifest_handlers/csp_handler.cc",
+    "manifest_handlers/csp_handler.h",
+    "manifest_handlers/permissions_handler.cc",
+    "manifest_handlers/permissions_handler.h",
+    "manifest_handlers/warp_handler.cc",
+    "manifest_handlers/warp_handler.h",
+    "manifest_handlers/widget_handler.cc",
+    "manifest_handlers/widget_handler.h",
+    "package/package.cc",
+    "package/package.h",
+    "package/wgt_package.cc",
+    "package/wgt_package.h",
+    "package/xpk_package.cc",
+    "package/xpk_package.h",
+    "permission_policy_manager.cc",
+    "permission_policy_manager.h",
+    "permission_types.h",
+  ]
+  deps = [
+    "//base",
+    "//base:i18n",
+    "//crypto",
+    "//net",
+    "//sql",
+    "//third_party/libxml",
+    "//third_party/zlib:zip",
+    "//url",
+  ]
+}
--- a/build/common.gni
+++ b/build/common.gni
@@ -1,0 +1,18 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file contains common build flags for building Crosswalk. Users should
+# NOT directly import this file into their build folder.
+# See //xwalk/build/linux.gni for how to import flags for Crosswalk-Linux Build
+
+root_extra_deps = [ "//xwalk:xwalk_builder" ]
+
+# From src/build/config/features.gni
+# Disable WebVR support. The code is still experimental and and ends up
+# pulling additional dependencies into our JARs (XWALK-6597).
+enable_webvr = false
+
+# From src/build/config/features.gni
+# Whether to include support for proprietary codecs.
+proprietary_codecs = true
--- a/build/xwalk.gni
+++ b/build/xwalk.gni
@@ -1,0 +1,10 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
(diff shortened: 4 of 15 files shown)


heke123:
Actually the gn_xwalk is a python wrapper to pass parameters to "gn gen".
In Crosswalk we use build/common.gypi to controll flags of chromium. But in GN system, all the controll flags are distributed in many *.gni files,  and modules import the *.gni file directly.
For the convenience of control flags and avoid to modify different *.gni files too much, we use the gn_xwalk to control all flags.

heke123:
I split the patch into small PRs to make it easy to review.  This PR only contains the linux part but without the test implementation.  I'll submit test implementation for linux in next PR.
Actually In my local machine all the unit tests & browser tests have passed (for linux build).

I modified some .cc files directly because there are some un-used variables which cause warnings. It is safe to remove them even for GYP build system.

heke123:
@rakuco   @halton  @PeterWangIntel    Please have a look, thanks.

halton:
@PeterWangIntel please give a peer review on this PR.

heke123:
@halton  I create a new JIRA XWALK-7180: "Add gn build support for crosswalk-linux".
The git log has been changed as: BUG=XWALK-7180.

halton:
@heke123 great! It is good behavior to divide complex task into pieces, please keep doing that.

@rakuco could you prioritize your time of code reviewing?

rakuco:
This is a good start, but there's some work to be done:
* The changes to the C++ files must be done separately -- preferably in a separate pull request, at the very least in a separate commit in this one.
* If this pull request is about GN on Linux, I don't expect to see anything related to Windows or Android in it. At the moment there's a mixture of Linux, Windows and Android support in the GN files that only makes everything look more confusing.
* The `gn_xwalk` approach is not good. It moves a lot of logic out of the build system and into the Python script, where all overrides are scattered around in `add_option()` calls without any description for users. Not only that, but it also implements a ton of non-Linux stuff that should not be part of this pull request. At the moment, the best course of action is to create a `build/args/xwalk.gni` with all the variable overrides and the logic to change them. Users should then do something like `echo 'import("//xwalk/build/args/xwalk.gni")' > out/args.gn` and then use `gn args` to add other options if necessary. It also has the benefit that any changes to `xwalk.gni` automatically cause GN to be run. See [this upstream bug](https://bugs.chromium.org/p/chromium/issues/detail?id=588513) for a larger discussion.
* Please take a look at the GN style guide, in particular the section about formatting and only listing source files once. At the moment this patch is doing the opposite (which is the gyp approach), and listing all available source files and removing some from the list depending on the operating system.
* Some files are being referenced with relative paths ("../extensions/common/constants.cc") for no reason.
* Now is a good time to break down those huge, monolithic targets in gyp into smaller ones. For example, there's no reason to declare the xwalk_runtime target in the top-level `BUILD.gn`.
* Most static library declarations should probably be source sets instead.

Overall, I get the feeling that you guys just did a word-by-word translation of what's currently in gyp. GN also has a different approach to things that does not seem to have been fully understood. In case you have not read them, I strongly suggest you take a look at the following documents in addition to the language reference;
* [GN Style Guide](https://chromium.googlesource.com/chromium/src/+/master/tools/gn/docs/style_guide.md)
* [GYP->GN conversion cookbook](https://chromium.googlesource.com/chromium/src/+/master/tools/gn/docs/cookbook.md)

rakuco:
This presentation is also quite useful: https://docs.google.com/presentation/d/15Zwb53JcncHfEwHpnG_PoIbbzQ3GQi_cpujYwbpcbZo/edit?usp=sharing

rakuco:
> Users should then do something like echo 'import("//xwalk/build/args/xwalk.gni")' > out/args.gn and then use gn args to add other options if necessary.

Which is also what some Chromium upstream bots do when they need to change several flags.

PeterWangIntel:
@rakuco 
Thank you very much for your comments, but one point I think we need to discuss more. 
We were intended to make a "word by word translation" when put hands to this task. 
The reason was pretty straightforward, nobody clearly know the details of all the existent modules (or features) : ozone, nacl, extension, sysapps, webUI, stubs to support Mac, and so on.
So I believe the feasible road-map is, firstly, to implement a reliable GN building and  publish it to all developers as early as possible, then we can make it more GN-style, or maybe, when the owners of modules (or features)  have opportunity to use GN to build XWalk, they will have the best idea to improve the building files for their part.

Agree, we need a balance between usable and beautiful for the initial patch . But by my experience in practice, sometimes "word to word translation" can be a good tip to help resolving secret bugs during transition.

PeterWangIntel:
@rakuco 
And one of the reasons of 'gn_xwalk' is to deal with legacy problems. Taking 'xwalk_version' as example, the 'xwalk.gpy' use a tricky syntax to catch the screen output of 'build/util/version.py' to give it a value. But there is no equal syntax for GN. Using script in 'gn_xwalk'  is a simple and direct way to resolve.

heke123:
@rakuco  Thanks very much for your review comments, they are very helpful and I'll make change accordingly. Like you mentioned, I'll do below changes one by one.
1) I'll repair all the format/style things.  Clean it to "linux only" code.
2) static_library to source_set.  No problem.
3) I'll remove the gn_xwalk approach, and implement in the way of "echo 'import("//xwalk/build/args/xwalk.gni")' > out/args.gn".  
4) As Peter points out:  in GYP the legacy code is: 
```
'variables': {
   'xwalk_version': '<!(python ../build/util/version.py -f VERSION -t "@MAJOR@.@MINOR@.@BUILD@.@PATCH@")', 
}
```
In this way GYP can execute python script and get the return value directly. But in GN, we have to use the "exec_script" to return the stdout of the script "version.py".( the only way I found in GN so far)  Although in Chromium they don't encourage the usage of "exec_script",  I'll try this way.  **@rakuco  Any comments on this?** 

Finally, About the **break down "xwalk_runtime" target**,  we referred both the GYP and GN implementation of content_shell which is very similar as our crosswalk. In BUILD.gn of content_shell, we found it is modified based on the content_shell.gypi, instead of breaking down or re-organizing.  But I agree it is very good idea to re-think the structure of build files, actually in our BUILD.gn for Android, we did re-organize the huge xwalk_android.gypi into lots of small BUILD.gn files for separate APKs.  But I don't think the target xwalk_runtime apply to this situation. I prefer we don't break down the xwalk_runtime , **@rakuco, How do you about this?**

Thanks all very much!

rakuco:
> In this way GYP can execute python script and get the return value directly. But in GN, we have to use the "exec_script" to return the stdout of the script "version.py".( the only way I found in GN so far) Although in Chromium they don't encourage the usage of "exec_script", I'll try this way. @rakuco Any comments on this?

It's very easy to change this from synchronously executing a Python script to obtain those values and add them as defines in the `xwalk_runtime` target to generating headers with those values instead. This applies to both gyp and GN and makes things cleaner for both build systems. Chromium already does something similar with Blink. See `src/build/util/BUILD.gn` for the GN version and `src/build/util/version.gypi`  along with the locations that include it for the gyp version.

> Finally, About the break down "xwalk_runtime" target, we referred both the GYP and GN implementation of content_shell which is very similar as our crosswalk. In BUILD.gn of content_shell, we found it is modified based on the content_shell.gypi, instead of breaking down or re-organizing. But I agree it is very good idea to re-think the structure of build files, actually in our BUILD.gn for Android, we did re-organize the huge xwalk_android.gypi into lots of small BUILD.gn files for separate APKs. But I don't think the target xwalk_runtime apply to this situation. I prefer we don't break down the xwalk_runtime , @rakuco, How do you about this?

Don't get so hung up on `xwalk_runtime`: it was simply an example of all the big targets defined at the wrong level that I was talking about; we are basically copying the same mistakes made with the gyp version of the build system. However, I guess it does make sense to change this only once we get rid of gyp to avoid requiring people to edit build system files in different locations in the future.

heke123:
@rakuco   Thanks very much!

heke123:
@rakuco  In this patch I made some modifications:
1) I added a config: -Wno-unused-const-variables,  so in this patch I don't need to change the C++ files which has warnings.  I'll remove it as soon as the un-used variables are cleaned in our source code.  It should in another PR.
2) Remove other platform related things like "is_android" "is_win".
3) Change static_library into source_set.
4) Remove gn_xwalk approach,  add all crosswalk build flags in xwalk.gni,   Now the "echo  import(xwalk.gni)" and "gn gen" work well as expected.
5) GN style clean.

Like we discussed above, I keeped the xwalk_runtime and didn't break it down.  I think so far it is ok to keep it like content_shell's BUILD.gn.  I totally agree that let's setup GN build system first, in next step we can see how to re-construct it.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/features.gni")
+import("//build/config/sanitizers/sanitizers.gni")
```
rakuco: You don't seem to be using any of the flags defined here.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+import("//build/config/ui.gni")
+import("//build_overrides/v8.gni")
```
rakuco: Ditto.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+import("//media/media_options.gni")
```
rakuco: Ditto.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+import("//third_party/widevine/cdm/widevine.gni")
```
rakuco: Ditto.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+import("//third_party/ffmpeg/ffmpeg_options.gni")
```
rakuco: Ditto.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+import("//xwalk/build/version.gni")
+
+config("xwalk_runtime_config") {
```
rakuco: This is a very generic name. In any case, we don't really support Precise anymore so I'll just remove this from gyp and you can drop it from the GN code too.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+  # TODO(rakuco): Remove this once we stop supporting Ubuntu Precise or
+  # default to use_sysroot==1. This is only required because Precise's dbus
+  # 1.4.18 causes
+  # <https://bugs.chromium.org/p/chromium/issues/detail?id=263960#c6>.
+  if (is_clang && !use_sysroot) {
+    cflags = [ "-Wno-reserved-user-defined-literal" ]
+  }
+}
+
+source_set("xwalk_runtime") {
```
rakuco: The entire ordering of the entries in this target does not match GN's style guide's recommendations.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+  public_configs = [ "//build/config:feature_flags" ]
```
rakuco: Why are you setting it here? This is done automatically in `src/build/config/BUILDCONFIG.gn`.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+  configs += [
+    # TODO(jschuh): crbug.com/167187 fix size_t to int truncations.
+    "//build/config/compiler:no_size_t_to_int_warning",
```
rakuco: This is Windows-specific.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+    "//xwalk/experimental:experimental",
+    "//xwalk/extensions:xwalk_extensions",
+    "//xwalk/resources:xwalk_pak",
+    "//xwalk/resources:xwalk_resources",
+    "//xwalk/resources:xwalk_resources_100_percent_pak",
+    "//xwalk/resources:xwalk_resources_200_percent_pak",
+    "//xwalk/resources:xwalk_resources_300_percent_pak",
+    "//xwalk/sysapps:sysapps",
+  ]
+  if (is_linux) {
```
rakuco: It makes more sense to put this at the bottom and group it with the other `is_linux` block you already have there.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+    deps += [ "//xwalk/resources:xwalk_strings" ]
+  }
+
+  include_dirs = [
+    "..",
```
rakuco: Again, from the "GYP->GN conversion cookbook":

> Note that many of the includes are trying to add the root “src” directory to the include path. This is always present in GN so you can remove these.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+    "$root_gen_dir",
```
rakuco: This is already done in `src/build/config/compiler/BUILD.gn`.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+  sources = [
+    "//extensions/common/constants.cc",
+    "//extensions/common/constants.h",
+    "//extensions/common/url_pattern.cc",
+    "//extensions/common/url_pattern.h",
+    "runtime/app/android/xwalk_main_delegate_android.cc",
+    "runtime/app/android/xwalk_main_delegate_android.h",
+    "runtime/app/xwalk_main_delegate.cc",
+    "runtime/app/xwalk_main_delegate.h",
+    "runtime/browser/android/cookie_manager.cc",
```
rakuco: I wasn't expecting to find Android and Windows-specific files here.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+    rebase_path(output_file, root_out_dir),
+  ]
+
+  outputs = [
+    "$output_file",
+  ]
+}
+
+#TODO: Remove this when there is no sign-comparation in xwalk code.
+config("no_sign_compare_warning") {
```
rakuco: Actually fixing this in the code takes only a few minutes, you should do that in another pull request instead of adding hacks here.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+  if (is_clang) {
+    cflags = [ "-Wno-sign-compare" ]
+  }
+}
+
+#TODO: Remove this when there is no unused const variables in xwalk code.
+config("no_unused_const_variable_warning") {
```
rakuco: Ditto.

In both cases, this is actually just a symptom of GN building with stricter compiler flags (`chromium_code` in gyp) by default. This is a bigger problem for Android (XWALK-6276), but fixing the errors that affect the Linux build should be really straight-forward.

BUILD.gn
```diff
@@ -0,0 +1,530 @@
...
+  if (is_clang) {
+    cflags = [ "-Wno-unused-const-variable" ]
+  }
+}
+
+executable("xwalk") {
+  deps = [
+    ":xwalk_runtime",
+  ]
+  include_dirs = [ ".." ]
```
rakuco: Same thing about this being unnecessary.

application/common/BUILD.gn
```diff
@@ -0,0 +1,58 @@
...
+    "package/wgt_package.cc",
+    "package/wgt_package.h",
+    "package/xpk_package.cc",
+    "package/xpk_package.h",
+    "permission_policy_manager.cc",
+    "permission_policy_manager.h",
+    "permission_types.h",
+  ]
+
+  include_dirs = [ "//xwalk/application" ]
```
rakuco: Not needed.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+# found in the LICENSE file.
+
+# This file contains Crosswalk-feature-related build flags. Users should
+# import it into their build folder by run:
+#   echo 'import("//xwalk/build/xwalk.gni")' > out/your_folder/args.gn'
+# Then run:
+#   gn gen out/your_folder
+
+# Set it false when building content_shell.
+declare_args() {
```
rakuco: It looks like you didn't read my comment in crosswalk-project/chromium-crosswalk#370 about using `build/secondary` in Chromium to get rid of the `is_build_crosswalk` hack.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  is_build_crosswalk = true
+}
+
+if (is_build_crosswalk) {
+  target_os = "linux"
```
rakuco: This 3-line block does not make sense. The `target_*` variables should be set by GN and the toolchain code only, there is no need to change them here. As for `is_debug`, I don't see a reason to change the default for Crosswalk. If upstream enables debug builds by default, so should we.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  target_cpu = "x64"
+  is_debug = false
+
+  # From src/build/config/features.gni
+  # TODO: Enable nacl for Linux.
```
rakuco: What's currently preventing it from being enabled? Just saying it needs to be enabled without saying why it's not working whereas the gyp version does is unhelpful.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  enable_nacl = false
+
+  # From src/build/config/features.gni
+  enable_plugins = true
```
rakuco: Why is this being changed here if `build/common.gypi` does not touch it?

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+
+  # From src/build/config/features.gni
+  enable_webvr = false
+
+  # From src/third_party/widevine/cdm/widevine.gni
```
rakuco: There's no need to set this here since commit b5f0d1330f484e6c3b3c8ceeed5c94dd8689f684, the default is `false`. I'll remove it from the gyp version.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  enable_widevine = false
+
+  # From src/build/config/compiler/BUILD.gn
+  # If we want stack unwind support for backtrace()
+  exclude_unwind_tables = false
+
+  # From src/third_party/ffmpeg/ffmpeg_generated.gni
```
rakuco: Wrong file, this variable is not defined there.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  ffmpeg_branding = "Chromium"
```
rakuco: For variables whose value differs from the upstream default only for the non-Linux case (e.g. in gyp we only change `ffmpeg_branding` on Android), I think it makes more sense not to include them in this commit. It makes it much easier to understand where they came from if they are added with the rest of the code to support Android, Windows and other targets later.

This applies to other variables, such as `use_rssdk`, `exclude_unwind_tables` and `v8_use_external_startup_data`.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+
```
rakuco: You did not include `ffmpeg_component`'s GN counterpart here.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  # From src/third_party/icu/config.gni
+  icu_use_data_file = true
```
rakuco: Why? The default is already `true` and it's not changed in `common.gypi`.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+
+  # From src/build/config/features.gni
+  proprietary_codecs = true
+
+  # From src/build/config/ui.gni
```
rakuco: Why are you changing this here?

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  toolkit_views = true
+
+  # From src/build/config/ui.gni
+  use_aura = true
```
rakuco: And this too.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+
+  # From //build/config/sysroot.gni.
+  # Android and Windows do not use the sysroot, and we cannot use it on
+  # Linux because we use libnotify that is not present there.
+  use_sysroot = false
+
+  # From src/build_overrides/v8.gni
+  v8_use_external_startup_data = false
+
+  # From src/v8/BUILD.gn
```
rakuco: I also don't get why you are changing this if it's not changed in gyp.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+  v8_use_snapshot = false
+
+  # Crosswalk specific flags:
```
rakuco: I'd rather move the Crosswalk-specific options to a separate file, so that GN files that require access to these settings explicitly include it instead. It makes it a lot easier to understand what's overriding default upstream values and what are new variables.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+    disable_bundled_extensions = false
+
+    # TODO: RealSense
+    # Whether to use the RealSense SDK (libpxc) video capture.
+    # use_rssdk = false
+
+    # Whether to use a WebUI-based file picker.
+    use_webui_file_picker = false
+
+    verify_xwalk_apk = false
```
rakuco: This is lacking a description, but, most importantly, it's unrelated to Linux.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+
+    # Name of Crosswalk Maven artifacts used to generate their
```
rakuco: These are also unrelated to Linux support.

build/xwalk.gni
```diff
@@ -0,0 +1,86 @@
...
+    # respective POM files.
+    xwalk_core_library_artifact_id = "xwalk_core_library_canary"
+    xwalk_shared_library_artifact_id = "xwalk_shared_library_canary"
+
+    xwalk_product_name = "XWalk"
```
rakuco: This only seems to be used by OS X code, so it could be dropped.

experimental/BUILD.gn
```diff
@@ -0,0 +1,66 @@
...
+    "native_file_system/virtual_root_provider_android.cc",
+    "native_file_system/virtual_root_provider_linux.cc",
+    "native_file_system/virtual_root_provider_mac.cc",
+    "native_file_system/virtual_root_provider_win.cc",
+  ]
+
+  include_dirs = [
+    "$root_gen_dir",
+    "$root_gen_dir/xwalk",
+    "$root_gen_dir/xwalk/experimental",
```
rakuco: This is the only entry that's really needed.

resources/BUILD.gn
```diff
@@ -0,0 +1,176 @@
...
+  resource_ids = "//xwalk/resources/resource_ids"
+  output_dir = "$root_gen_dir/xwalk/locales"
+  outputs = [
+    "$root_gen_dir/xwalk/grit/xwalk_strings.h",
+    "$root_gen_dir/xwalk/locales/en-US.pak",
+    "$root_gen_dir/xwalk/locales/zh-CN.pak",
+  ]
+}
+
+copy("xwalk_strings") {
```
rakuco: This does not work for me:

```
ERROR at //tools/grit/repack.gni:24:3: Duplicate output file.
  action(target_name) {
  ^--------------------
Two or more targets generate the same output:
  locales/en-US.pak

This is can often be fixed by changing one of the target names, or by 
setting an output_name on one of them.

Collisions:
  //chrome:repack_locales_pack_en-US
  //xwalk/resources:xwalk_strings

See //xwalk/resources/BUILD.gn:166:1: Collision.
copy("xwalk_strings") {
^----------------------
```

sysapps/BUILD.gn
```diff
@@ -0,0 +1,166 @@
...
+      "//components/storage_monitor:storage_monitor",
+      "//media:media",
+      "//third_party/ffmpeg:ffmpeg",
+    ]
+  }
+
+  include_dirs = [
+    "$root_gen_dir",
+    "$root_gen_dir/xwalk",
+    "$root_gen_dir/xwalk/sysapps",
```
rakuco: Only this is needed.

rakuco:
Why are the unit and browser tests not being built? It's hard to approve this pull request if I can't even easily test if the generated binaries work as expected.

BUILD.gn
heke123: I did consider to remove those Android/Windows files, but finally just leave them alone.  The reason is that in //build/config/BUILDCONFIG.gn,  there is a "sources_assignment_filter",  it will filter out all platform-specific files automatically. To avoid mistakes, I just keep all the source files here.  
If you think it is not good, I can remove them in this linux PR. But I think we'll add them back sooner or later.  so could we just keep the them?

build/xwalk.gni
heke123: `is_debug` should be same as upstream.
My understanding:  User should set `target_os/target_cpu/is_debug` by` "gn args"`.  Now we use this `"echo import(xwalk.gni)"` approach, User just need to edit this xwalk.gni to modify the values of this 3 lines. 
Another reason is, some flags have different values for different platforms.  for example:
```
 ['OS=="android"', {
     'release_unwind_tables%': 0,
}]
```
By add defines of `target_os`,  I can later write as: 
`if(target_os == Android) { release_unwind_tables = false }`

heke123: Maybe to ask user set `target_os/target_cpu/is_debug` in command line is better.  But I still need to add conditions like `if(target_os == Android)` even the `target_os`  comes from user's command line args.

build/xwalk.gni
heke123: Sorry I didn't get your points.  My understanding:  GN will always tries to check the BUILD.gn file in src/ directory.
Only if it's not found, the build/secondary/BUILD.gn will be checked.  In our case, the src/BUILD.gn is always there..
Do you mean to add some options in command line like "gn gen out/Debug --root=/home/build --dotfile=/home/my_gn_file.gn"?
Another question is, should we enable build content_shell in the environment of both Crosswalk and Chromium-Crosswalk are downloaded?

build/xwalk.gni
heke123: Sorry, there is some android flags.  My initial purpose to split out PR for Linux is to make it easier to review. So I didn't delete those android flags clean.  Anyway, you are right, it should have Linux related only in this PR.  I'll make it clean.

resources/BUILD.gn
heke123: I delete the out/ folder and build again. no error found.  But I did meet this before: in xwalk.gyp:
```
  ['OS=="linux" and use_ozone!=1', {                                                                    
            'defines': ['USE_GTK_UI'],
            'dependencies': [
              '../chrome/browser/ui/libgtk2ui/libgtk2ui.gyp:gtk2ui',
            ],
  }]
```
If add the `libgtk2ui` in your deps, you will find the` //chrome/BUILD.gn` is finally imported, then it also try to generate the output file `"locales/en-US.pak"`. That's why it complains collisions.
Peter checked the "USE_GTK_UI" usage in crosswalk,  we think it should not be used anymore. Without build this libgtk2ui, there is not collision error.
Actually I'm not understand this background of `USE_GTK_UI` and `libgtk2ui`,   @PeterWangIntel  Any comments on this?

BUILD.gn
rakuco: The main reason why this makes more sense to me is that landing them together with all the other nuts and bolts for supporting another target makes it easier to understand, revert and reason about: in one commit you can see everything being added together as a coherent piece.
I don't have a strong objection to keeping these lines though, so either way is fine.

build/xwalk.gni
rakuco: > Sorry I didn't get your points. My understanding: GN will always tries to check the BUILD.gn file in src/ directory.
> Only if it's not found, the build/secondary/xwalk/BUILD.gn will be checked. But in our case, the src/BUILD.gn is always there no matter you checkout "chromium-crosswalk only" or "chromium-crosswalk plus crosswalk." Could you please give more explanations on this?

Again, see GN's documentation: [here](https://chromium.googlesource.com/chromium/src/tools/gn/+/master/docs/cookbook.md#Common-target-conversion) and [here](https://chromium.googlesource.com/chromium/src/tools/gn/+/master/docs/language.md#Labels). There's also [the comment in .gn itself](https://github.com/crosswalk-project/chromium-crosswalk/blob/crosswalk-22/51.0.2704.105/.gn#L8) explaining what `secondary_source` does.

Quoting from the Language and Operation document:

> When loading a build file, if it doesn‘t exist in the given location relative to the source root, GN will look in the secondary tree in build/secondary. The structure of this tree mirrors the main repository and is a way to add build files for directories that may be pulled from other repositories where we can’t easily check in BUILD files. The secondary tree is a fallback rather than an override, so a file in the normal location always takes precedence.

I don't understand why you are talking about `src/BUILD.gn`. Either `src/xwalk/BUILD.gn` exists and is read from `src/BUILD.gn` or it does not exist (the content shell builds), in which case `src/BUILD.gn` reads `src/build/secondary/xwalk/BUILD.gn`.

With all the above said, another, even simpler approach, is to just change `root_extra_deps` in `src/xwalk/build/xwalk.gni` -- this variable was added to `src/BUILD.gn` precisely for this purpose.

> Another question is, should we enable build content_shell in the environment of both Crosswalk and Chromium-Crosswalk are downloaded?

It must be possible to build content_shell regardless of whether `src/xwalk` is present or not, just like it already happens with gyp.

build/xwalk.gni
rakuco: > My understanding: User should set target_os/target_cpu/is_debug by"gn args". Now we use this "echo import(xwalk.gni)" approach, User just need to edit this xwalk.gni to modify the values of this 3 lines.

Requiring users to modify `xwalk.gni` makes **zero** sense. Not only does this mean a file that is part of the checkout will need to be modified by each user, but it will also prevent multiple build directories from having different build flags.

If one wants to change `is_debug`, it's just a matter of adding it to `args.gn` via `gn args`, the `import` clause is irrelevant in this context:
```
import("//xwalk/build/xwalk.gni")
is_debug = false
```

The problem with the conditionals is real though, and I wasn't aware of it. For now, what comes to mind is having all the common settings in, say, `xwalk.gni` or `common.gni` and then separate files such as `linux.gni` and `windows.gni` with additional settings. `linux.gni` imports `xwalk.gni` and adds some settings after that, and users import `linux.gni` in their `args.gn` instead.

However, this approach is not very flexible, so I'll ask the gn-dev mailing list about this issue.

build/xwalk.gni
heke123: Finally I found what causes the mis-understanding:
I assume when running `"gn gen"` for crosswalk,  all the upstream original targets in `src/BUILD.gn` (like `group("ALL")`) should never be triggered,  no ninja files should be generated for those upstream targets.  Only those targets DEPsed by xwalk will be generated into ninja files.
In the approach of `"build/secondary"` or `"root_extra_deps"`,  the `group("All")` will be generated into ninja files although they are not compiled.
I guess maybe it's why you meet the error of 
```
  "Collisions:
    //chrome:repack_locales_pack_en-US
    //xwalk/resources:xwalk_strings"
```
It is because the chrome targets conflict with crosswalk targets,  they both try to output `locales/en-US.pak` which is used for locales in resource_bundle module.

heke123: I understand we should try to not touch the src/BUILD.gn.  Actually there is only one flag is added in BUILD.gn which is very easy to rebase. On another hand, To some extent, Crosswalk and Chrome are quite similar apps both based on content_layer, I'm not sure if there are more underlying GN target conflicts. Fix those conflicts will also bring new rebase burden.
I'm not very sure about this part, please help to give some suggestions,  Thanks very much.

heke123:
No problem, All the test are passed in my Local machine, I split them out just to make the patch smaller.  It's good to add the test part in this PR.

build/xwalk.gni
rakuco: > I understand we should try to not touch the src/BUILD.gn. Actually there is only one flag added in BUILD.gn which is very easy to rebase.

Are you referring to the `building_crosswalk` flag in the chromium-crosswalk pull request? If so, the pull request itself shows that's not the case: at the moment the entire file is being rewritten there because of formatting, and not fixing the formatting means the style checks run in each pull request will always fail.

> I'm not sure if there are more underlying GN target conflicts

This is actually pretty easy to find out: if any conflicts arise `gn gen` will fail to run just like it did now.

> On another hand, To some extent, Crosswalk and Chrome are quite similar apps both based on content_layer [...]. Fix those conflicts will also bring new rebase burden.
> I'm not very sure about this part, please help to give some suggestions, Thanks very much

You are correct that this is different compared to what we currently do with gyp, where `gyp_xwalk` makes gyp start from` xwalk/xwalk.gyp` and ignore the chromium gyp files such as `all.gyp`. That does makes things easier for us, but I don't see an easy way of achieving the same thing with GN, possibly by design: we'd either need to heavily patch `src/BUILD.gn` or add a `.gn` to `src/xwalk` and require people to use `src/xwalk` as the GN root. However, I don't think the latter would even work because if `src/xwalk` it won't be able to reference anything above it in `src`.

So in practice the only ideas I can think of are:
* Patching the Chromium GN code to stop packing the locales (not really feasible unless a flag such as `building_crosswalk` is introduced so not a very good idea)
* `git mv`'ing `src/BUILD.gn` to something else and replacing it with a smaller file that only introduces a flag like `building_crosswalk` and imports either the original `src/BUILD.gn` or `xwalk/BUILD.gn`.
* Not copying the locales to `locales/` and use `xwalk/locales/` instead. The release building code may copy those files manually to `locales`, or our code could be changed to look for the locales in `xwalk/locales` instead.

The 3rd option requires more investigation, but is the cleanest approach. Option 2 is a hack that may be usable for a little while, and option 1 is out of question so far to me.

build/xwalk.gni
heke123: @rakuco  How about I start to implement this part by the approach of " the `linux.gni `imports `common.gni`", If we find a better way, I'll re-implement then.

build/xwalk.gni
heke123: @rakuco   I submitted a PR based on the 3rd option.  The gyp file is also changed to avoid break our regular build. Please help to have a look.
I think I can continue the GN work with the approach of "3rd option" plus "root_extra_deps".  Thanks so much for your help!

build/xwalk.gni
rakuco: Yes, that's basically what's been suggested so far here: https://groups.google.com/a/chromium.org/d/msg/gn-dev/DJns9h39Hng/GlighcU5BwAJ

heke123:
@rakuco  About the flag:
```
# From src/third_party/ffmpeg/ffmpeg_options.gni
# Set true to build ffmpeg as a shared library.
is_component_ffmpeg = true
```
In file  `src/third_party/ffmpeg/ffmpeg_options.gni`, line 36, I found:  `"is_component_ffmpeg = is_component_build"`.  By default the flag "is_component_build" is false, Everything works fine when ffmpeg is static linked.
If we override `is_component_ffmpeg = true` in our xwalk.gni,  while keeps` "is_component_build = false"`, then I found:  Although the "libffmpeg.so" is built out successfully, but runtime error pop out: `"./xwalk: error while loading shared libraries: libffmpeg.so: cannot open shared object file: No such file or directory"`
I also tried on content_shell (chrome version 50.0.2654.0), the same error is found.  So I guess it is a upstream issue that we cannot set `"is_component_ffmpeg=true"` while` "is_component_build = false"`?

If we set `"is_component_build = true"`,  then everything works fine.

Do you have any comments on this?  Thanks.

heke123:
Docs says: `The component build is currently the default for debug non-iOS builds (it doesn’t work for iOS). You can force it on for release builds using the GN build arg`
The default value of `is_debug` is true. so the default value of `is_component_build` is true.
Anyway, the problem of "we cannot set `is_component_ffmpeg=true `while `is_component_build = false` " still exists.  Maybe we can ignore this issue if we accept always set `is_component_build` as true for both debug and release build .

rakuco:
I've spoken to @heke123 about the ffmpeg issue: I was unable to reproduce it on M52 with a content_shell build; he's going to double-check it on his side, and after that I guess we're ready for another review round.

heke123:
@rakuco  I did "gclient sync" on my codes and will check the "ffmpeg" issue again.  Please have a look on the updated PR.  Browser test and unit test are also included.  Thanks very much.

heke123:
By the way, is this page too messy? should I create a clean page to start a new round of review?

rakuco:
Thanks; I took a quick look and the remaining issues are small. I'll also try out the ffmpeg issue you mentioned.

> By the way, is this page too messy? should I create a clean page to start a new round of review?

This is fine, it just shows we've been discussing this code a lot.

rakuco:
OK, I've taken a look at the ffmpeg issue, and it's related to RPATH in the executables.

gyp always passed `-Wl,-rpath=$ORIGIN/.` to the compiler when building executables, which meant the resulting ELF binaries had an `RPATH` field pointing to the directory where the binary is at, so `libffmpeg.so` was always found regardless of where the binaries were invoked from.

GN takes a different, opt-in approach, where the `//build/config/gcc:rpath_for_built_shared_libraries` config must be manually added to a target for this to happen (it happens automatically when `is_component_build=true`). The comment above the config's definition says it's not on by default because of possible issues with Chromium's setuid binaries such as `chrome_sandbox`.

We should probably re-evaluate if it still makes sense for us to build with `is_component_ffmpeg` on by default on Linux, but, until then, you can do something like this:

```diff
diff --git a/BUILD.gn b/BUILD.gn
index 26ec808..eab8e54 100644
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -28,6 +28,10 @@ executable("xwalk") {
   deps = [
     ":xwalk_runtime",
   ]
+
+  if (is_linux && !is_component_build && is_component_ffmpeg) {
+    configs += [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
+  }
 }
 
 source_set("xwalk_runtime") {
 ```

Do note, however, that the dependency on ffmpeg comes via `xwalk_runtime` depending on `media`, so the tests are also affected. It might make sense to move that block to `xwalk_runtime` instead, and for tests I suggest you `git grep` for `rpath_for_built_shared_libraries` in `src` to see how upstream takes care of test binaries.

File: build/common.gni [hunk 37]
```diff
@@ -0,0 +1,52 @@
...
+# found in the LICENSE file.
+
+# This file contains common build flags for building Crosswalk. Users should
+# NOT directly import this file into their build folder.
+# See //xwalk/build/linux.gni for how to import flags for Crosswalk-Linux Build
+
+root_extra_deps = [ "//xwalk:xwalk_builder" ]
+
+# From src/build/config/features.gni
+# TODO: Enable nacl for Linux.
```
rakuco: Is it enabled or not after all? What's needed to enable it?

build/common.gni
```diff
@@ -0,0 +1,52 @@
...
+# Set true to build ffmpeg as a shared library.
+#is_component_ffmpeg = true
+
+# From src/build/config/features.gni
+proprietary_codecs = true
+
+# From //build/config/sysroot.gni.
+# Android and Windows do not use the sysroot, and we cannot use it on
+# Linux because we use libnotify that is not present there.
+use_sysroot = false
```
rakuco: Duplicate of what's done in `linux.gni`.

BUILD.gn
```diff
@@ -0,0 +1,495 @@
...
+    "//xwalk/resources:xwalk_resources_300_percent_pak",
+    "//xwalk/sysapps",
+  ]
+
+  if (is_linux) {
+    deps += [ "//xwalk/resources:xwalk_strings" ]
+  }
+
+  # TODO: Add WebNotification for Linux
+  # WebNotification needs libgtk2ui and libnotify.
```
rakuco: This must be solved before we can officially switch to GN.

extensions/test/BUILD.gn
```diff
@@ -0,0 +1,219 @@
...
+
+# Cannot specify destination folder in shared_library so we have to copy them.
+copy("cp_echo_extension") {
+  deps = [
+    ":echo_extension",
+    ":echo_extension_messaging_2",
+  ]
+  sources = [
+    "$root_out_dir/libecho_extension.so",
+    "$root_out_dir/libecho_extension.so.TOC",
```
rakuco: You should check what `TOC` files are before doing this, they're not needed at all.

extensions/test/BUILD.gn
```diff
@@ -0,0 +1,219 @@
...
+  ]
+}
+
+shared_library("echo_extension_messaging_2") {
+  sources = [
+    "echo_extension_messaging_2.c",
+  ]
+}
+
+# Cannot specify destination folder in shared_library so we have to copy them.
```
rakuco: `output_dir = "/some/path"`?

extensions/test/BUILD.gn
```diff
@@ -0,0 +1,219 @@
...
+  args = [
+    "$root_path/test/test.idl",
+    "--root=$root_path",
+    "--destdir=$gen_path/xwalk/extensions/",
+    "--namespace=xwalk::jsapi::test",
+    "--generator=cpp",
+  ]
+}
+
+shared_library("echo_extension") {
```
rakuco: For this kind of internal target I'd add a `visibility = [":*"]`. `git grep` for similar uses in `src` to see what I'm talking about.

resources/BUILD.gn
```diff
@@ -0,0 +1,163 @@
...
+    "$root_gen_dir/xwalk/resources/grit/xwalk_resources.h",
+    "$root_gen_dir/xwalk/resources/grit/xwalk_resources_map.cc",
+    "$root_gen_dir/xwalk/resources/grit/xwalk_resources_map.h",
+  ]
+  deps = [
+    ":generate_xwalk_resources",
+  ]
+}
+
+grit("generate_xwalk_strings") {
```
rakuco: `generate_xwalk_strings` and `xwalk_strings` don't look great. Part of what we spent so much time discussing in #3849 was how it was a bad idea to hardcode the locale names. And here you are doing it twice (and you'd need to do it again in the GN version of `xwalk_win_zip.gypi` for the same reasons we discussed at length in the other pull request).

Unfortunately the GN version of the `grit` action requires a list of outputs so we can't fully escape this.

However, you should study more closely what Chromium itself does: take a look at how the `chrome_repack_locales` action works, see how Chromium avoids duplication by keeping a list of locales in `build/config/locales.gni` and try to do something similar here.

resources/BUILD.gn
```diff
@@ -0,0 +1,163 @@
...
+  outputs = [
+    "$root_gen_dir/xwalk/grit/xwalk_strings.h",
```
rakuco: If you're setting `output_dir` you should need to add the same prefix here again.

resources/BUILD.gn
```diff
@@ -0,0 +1,163 @@
...
+    "$root_gen_dir/xwalk/locales/en-US.pak",
+    "$root_gen_dir/xwalk/locales/zh-CN.pak",
+  ]
+  source = "//xwalk/runtime/resources/xwalk_strings.grd"
+  resource_ids = "//xwalk/resources/resource_ids"
+  output_dir = "$root_gen_dir/xwalk/locales"
```
rakuco: If you're generating a file in `grit/` this value is wrong.

extensions/BUILD.gn
```diff
@@ -0,0 +1,93 @@
...
+}
+
+copy("xwalk_extensions_resources") {
+  outputs = [
+    "$root_gen_dir/xwalk/grit/{{source_file_part}}",
+  ]
+  sources = [
+    "$root_gen_dir/xwalk/extensions/grit/xwalk_extensions_resources.h",
+  ]
+  public_deps = [
```
rakuco: Why a `public_dep`? If another target depends on `xwalk_extensions_resources` and `xwalk_extensions_resources` depends on `generate_xwalk_extensions_resources` then there's no need to make the latter a public dependency of the former.

extensions/BUILD.gn
```diff
@@ -0,0 +1,93 @@
...
+    "//base",
+    "//content",
+    "//ipc",
+    "//third_party/WebKit/public:blink_headers",
+    "//url",
+    "//v8",
+  ]
+}
+
+grit("generate_xwalk_extensions_resources") {
```
rakuco: You're copying the gyp version so it isn't really your fault, but I'm pretty sure this whole generate-then-copy dance can be avoided if you just set the right parameters to put the files in their final locations in this action. Can you add a `# TODO(heke123): Get rid of this grit+copy combination by putting the files in the right places in grit` to this and other locations with the same combo?

BUILD.gn
heke123: Yes, I plan to solve it in another PR next to this one.
The unsolved problem is in xwalk/build/system.gyp we use pkg-config on libnotify.  I tried to follow "src/config/linux/pkg_config.gni" to write a new one, but failed.
I'll continue investigate on this later.

build/common.gni
heke123: It is done now.  Previously I found even in gyp it cannot build with `disable_nacl = 0`,  In lastest code everything works fine.

build/common.gni
```diff
@@ -0,0 +1,52 @@
...
+
+# From src/build_overrides/v8.gni
+v8_use_external_startup_data = true
+
+# Crosswalk specific flags:
+declare_args() {
```
heke123: Yes, done.

extensions/BUILD.gn
heke123: The purpose of `copy` is only to move the` .h` file into `gen/xwalk/grit`.  It is so ugly that I removed most of them in `sysapps/  extensions/.`  Instead I changed the `include_dir` of their source_set.
I totally agree the clean way is to correct the file places in grit.  No problem, I'll add TODO and clean it in future.

extensions/test/BUILD.gn
heke123: It works, all the `cp_libxxx` targets are not needed anymore.
So strange last time GN complains error when I add the `output_dir`.  Anyway, it works now :D

resources/BUILD.gn
heke123: I found the `generate_xwalk_strings` is much like targets in `chrome/app/BUILD.gn`.  The problem here is what `xwalk_win_zip.gypi`(will be converted to gn format) wants in its "deps" list is a **directory**, but **outputs** of the `grit` or `copy` is a `file-list`. So I think the key is `how to set the directory as an item of outputs` in a target.
I had a look on chrome_repack_locales, I understand how it defines its template,  but I didn't find any similar thing helpful to our xwalk_string problem :(
Could you please help to give more hints? Thanks!

heke123: by the way, I will remove line43, the "$root_gen_dir/xwalk/locales/en-US.pak" should not be packed into xwalk.pak.

heke123:
In chrome or content, I found they add` rpath_for_built_shared_libraries` directly when needs.
I found our test executable target which has no dependency on xwalk_runtime also pops out errors without this `rpath_for_built_shared_libraries` config.
How about we add this config for every executable(*browsertest,  *unittest)?

BUILD.gn
rakuco: Looking at uses of `pkg_config.gni` instead of `pkg_config.gni` itself is a lot more helpful. `build/config/linux/atk/BUILD.gn`, for example.

resources/BUILD.gn
rakuco: Sorry, I meant you should _not_ need to add the same prefix here again.

resources/BUILD.gn
heke123: GN automatically detects the value for us, which guarantees the generated files path are correct.
Problem here is that I try to force the grit to output a folder,  it prints warnings as below:
```
Expected output files:
/home/heke/master/crosswalk-src/src/out/linus_release/locales/grit/xwalk_strings.h
**/home/heke/master/crosswalk-src/src/out/linus_release/locales/xwalk**
/home/heke/master/crosswalk-src/src/out/linus_release/locales/xwalk/en-US.pak
/home/heke/master/crosswalk-src/src/out/linus_release/locales/xwalk/zh-CN.pak
Actual output files:
/home/heke/master/crosswalk-src/src/out/linus_release/locales/grit/xwalk_strings.h
/home/heke/master/crosswalk-src/src/out/linus_release/locales/xwalk/en-US.pak
/home/heke/master/crosswalk-src/src/out/linus_release/locales/xwalk/zh-CN.pak
Missing output files:

Extra output files:
/home/heke/master/crosswalk-src/src/out/linus_release/locales/xwalk
```

resources/BUILD.gn
rakuco: > I found the generate_xwalk_strings is much like targets in chrome/app/BUILD.gn.

First of all, did you understand the relationship between all the `*_strings` targets in `chrome/app/BUILD.gn` and the calls to `chrome_repack_locales` in `chrome/BUILD.gn`? Each `*_strings` target processes a different .grd file and produces several pak files out of it. The `chrome_repack_locales("repack_locales_pack")` call then goes through each locale and merges all those .pak files generated in the other targets into a single pak per locale.

> The problem here is what xwalk_win_zip.gypi(will be converted to gn format) wants in its "deps" list is a directory, but outputs of the grit or copy is a file-list. So I think the key is how to set the directory as an item of outputs in a target.
> I had a look on chrome_repack_locales, I understand how it defines its template, but I didn't find any similar thing helpful to our xwalk_string problem :(

`xwalk_win_zip.gypi` currently uses a directory as input because it's possible and easy with gyp. If it doesn't make sense with GN, we just use a list of files instead (remember what I said about not doing a word-by-word translation of the gyp files weeks ago?).

My whole point with my first comment in this line is that, according to what you are doing so far, there would be hardcoded lists of locales in at least 3 separate locations: `generate_xwalk_strings`, `copy_xwalk_strings` and the GN version of `xwalk_win_zip.gypi`. This is very error-prone and can be avoided if you keep one single list of locales somewhere (just like Chromium does with `build/config/locales.gni`) and use only that list all targets. And my suggestion to check `chrome_repack_locale` comes from the fact that that template takes care of setting the correct final output location of the pak files; even though we likely don't need to call grit's `repack.py` you can still take some ideas from that template code that can help you here.

resources/BUILD.gn
rakuco:
```
grit("generate_xwalk_strings") {
  outputs = [
    "../grit/xwalk_strings.h",
    "en-US.pak",
    "zh-CN.pak",
  ]
  source = "//xwalk/runtime/resources/xwalk_strings.grd"
  resource_ids = "//xwalk/resources/resource_ids"
  output_dir = "$root_gen_dir/xwalk/locales"
}
```

works just fine for me (in other words, `output_dir` per se is not wrong, but its combination with fully-qualified paths in `outputs` doesn't make much sense, which does back to my comment above this one and the other about using a single locales list).

sysapps/BUILD.gn
```diff
@@ -0,0 +1,195 @@
...
+}
+
+copy("xwalk_sysapps_resources") {
+  outputs = [
+    "$root_gen_dir/xwalk/grit/{{source_file_part}}",
+  ]
+  sources = [
+    "$root_gen_dir/xwalk/sysapps/grit/xwalk_sysapps_resources.h",
+  ]
+  public_deps = [
```
rakuco: My comment about `public_deps` in `extensions/BUILD.gn` also applies here.

heke123:
@rakuco  Sorry to response you late because I took the annual leave. (and will be absent tomorrow)
I updated the PR based on discussion above, please have a look. Thanks!

test/BUILD.gn
```diff
@@ -0,0 +1,77 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# import flag toolkit_views from ui.gni
```
rakuco: This comment is quite redundant, it makes sense to remove it.

BUILD.gn
```diff
@@ -0,0 +1,507 @@
...
+    "runtime/renderer/xwalk_render_thread_observer_generic.h",
+  ]
+  if (is_linux && !is_component_build && is_component_ffmpeg) {
+    # Set RPath for executable xwalk, xwalk_browsertest and xwalk_unittest.
+    public_configs = [ "//build/config/gcc:rpath_for_built_shared_libraries" ]
+  }
+
+  # TODO(heke123): Generate header files into gen/xwalk/grit by putting the
+  # files in the right places in grit. Clean below include_dirs.
+  include_dirs = [
```
rakuco: I don't think this is necessary as you're also doing this in other files (`xwalk_application_resources_include_dir` in `application/BUILD.gn`, for example).

BUILD.gn
```diff
@@ -0,0 +1,507 @@
...
+    "$root_gen_dir/xwalk",
+    "$root_gen_dir/xwalk/application",
+    "$root_gen_dir/xwalk/resources",
+    "$root_gen_dir/xwalk/sysapps",
+  ]
+  defines = [
+    "XWALK_VERSION=\"$xwalk_version\"",
```
rakuco: Minor nitpick: this should be alphabetically sorted.

BUILD.gn
```diff
@@ -0,0 +1,507 @@
...
+    "//xwalk/resources:xwalk_resources_300_percent_pak",
+    "//xwalk/resources:xwalk_strings",
+    "//xwalk/sysapps",
+  ]
+
+  if (is_linux) {
+    deps += [ "//xwalk/resources:xwalk_locales" ]
+  }
+
+  # TODO: Add WebNotification for Linux
```
rakuco: The usual style from Chromium (which we follow) is that TODOs and FIXMEs should always have someone responsible for it, so this should be "TODO(heke123): ...".

BUILD.gn
```diff
@@ -0,0 +1,507 @@
...
+      "runtime/browser/ui/webui/xwalk_web_ui_controller_factory.h",
+      "runtime/browser/ui/xwalk_web_dialog_view.cc",
+      "runtime/common/url_constants.cc",
+      "runtime/common/url_constants.h",
+    ]
+    deps += [ "//content/app/resources" ]
+  }
+}
+
+action("generate_upstream_blink_version") {
```
rakuco: The comment above this action's gyp version ("While we could just call lastchange.py [...]") is pretty important, so it should be added here too.

build/common.gni
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file contains common build flags for building Crosswalk. Users should
+# NOT directly import this file into their build folder.
+# See //xwalk/build/linux.gni for how to import flags for Crosswalk-Linux Build
+
+# Crosswalk specific build flags are defined in xwalk.gni. Import them.
```
rakuco: I don't see why you need to import this file, you're not using any of the values defined there. The BUILD.gn and .gni files using those values should be the ones importing it.

build/common.gni
```diff
@@ -0,0 +1,31 @@
...
+import("//xwalk/build/xwalk.gni")
+
+root_extra_deps = [ "//xwalk:xwalk_builder" ]
+
+# From src/build/config/features.gni
```
rakuco: You're enabling this on more platforms than necessary. Look at how it's defined in `build/config/features.gni`: you basically don't need to set `enable_nacl` here at all (disabling it in eventual win or mac gni files should do the job).

build/common.gni
```diff
@@ -0,0 +1,31 @@
...
+enable_nacl = true
+
+# From src/build/config/features.gni
```
rakuco: Its **very** important not to discard the comments from the gyp version. If I just look at this assignment I have no idea why it is being made.

build/common.gni
```diff
@@ -0,0 +1,31 @@
...
+enable_webvr = false
+
+# From src/build/config/compiler/BUILD.gn
+# If we want stack unwind support for backtrace()
+exclude_unwind_tables = false
```
rakuco: This is only `false` for Android.

build/common.gni
```diff
@@ -0,0 +1,31 @@
...
+
+# From src/third_party/ffmpeg/ffmpeg_options.gni
+ffmpeg_branding = "Chromium"
```
rakuco: You don't need to set this one, it was only changed in gyp based on `mediacodecs_EULA`, which was marked for removal.

build/common.gni
```diff
@@ -0,0 +1,31 @@
...
+
+# From src/build/config/features.gni
+proprietary_codecs = true
+
+# From src/build_overrides/v8.gni
+v8_use_external_startup_data = true
```
rakuco: The default is already true (and we're only changing it for Android), so you can remove it from here.

rakuco:
I've left some inline comments. We're practically done here: you basically need to take care of documenting the variables we're setting in `build`, not setting some to their default values etc.
Don't forget that the chromium-crosswalk side needs to be taken care of first though (it needs to land, you need to roll `DEPS.xwalk` etc).

BUILD.gn
heke123: Without this include_dirs it complains:
```
../../xwalk/runtime/renderer/xwalk_content_renderer_client.cc:18:10: fatal error: 'grit/xwalk_application_resources.h' file not found
#include "grit/xwalk_application_resources.h"
```
I think we can temporally leave them alone as TODO for now, I'll clean all the TODOs later.

build/common.gni
heke123: In GYP the original flag is `release_unwind_tables = 1`,  in GN it is "exclude"_unwind_tables, so the value is `false`.

rakuco: Ooops, I hate it when they change the meaning of a variable like that :-)

Still, the effect is the same:
* in gyp `release_unwind_tables` is on by default, and disabled for official Chrome builds (and also on Android for us)
* in gn, `exclude_unwind_tables` is off by default, enabled for official Chrome builds and we want it to be enabled on Android (i.e. you only want to change this value to `true` in a future `android.gni`).

BUILD.gn
rakuco: I'm raising this issue because it's really easy to fix and at the moment you have the worst of both worlds: `config`s for most of those directories in other `BUILD.gn` files + duplication of the same lines here.

* `xwalk_application_resources_include_dir` takes care of adding `$root_gen_dir/xwalk/application`
* `xwalk_sysapps_resources_include_dir` takes care of adding `$root_gen_dir/xwalk/sysapps`

This leaves us with `$root_gen_dir/xwalk/resources`, which can be solved by a config entry in `resources/BUILD.gn` just like the others you already have, and `$root_gen_dir/xwalk`, which is the only one that makes sense here (with a TODO for getting rid of it replacing the includes with proper, specialized ones).

build/common.gni
heke123: Yes, thanks.

BUILD.gn
heke123: I have to add those configs like` xwalk_application_resources_include_dir` here if we delete the `include_dirs` items to make the build pass. do you think it's ok?

rakuco: I was hoping you'd get that config for free by depending on another resource target that depends on it (e.g. other targets depend on `xwalk_application_resources`, not `xwalk_application_resources_include_dir` directly).

heke123: That needs to change `public_deps` into  `all_dependent_configs` to make the config visible.

extensions/test/BUILD.gn
```diff
@@ -0,0 +1,157 @@
...
+
+shared_library("echo_extension_messaging_2") {
+  visibility = [ ":*" ]
+  sources = [
+    "echo_extension_messaging_2.c",
+  ]
+  output_dir = "$root_out_dir/tests/extension/echo_extension"
+}
+
+shared_library("bad_extension") {
```
rakuco: We were talking about how `xwalk_extensions_browsertest` was broken in Debug mode earlier today because it was failing with errors such as

    error while loading shared libraries: libbad_extension.so: cannot open shared object file: No such file or directory

This come from the fact that you're using `shared_library` for these targets, which means the targets that _depend_ on them will link against them directly. In other words, linking the `xwalk_extensions_browsertest` binary will try to link all these extension .so's against it, which is definitely not what we want (the Release case works because GN passes `--as-needed` to the linker, which removes unneeded shared library dependencies automatically).

You actually need to change all these targets to `loadable_module` instead.

BUILD.gn
```diff
@@ -0,0 +1,507 @@
...
+    "runtime/browser/ssl_error_page.cc",
+    "runtime/browser/ssl_error_page.h",
+    "runtime/browser/storage_component.cc",
+    "runtime/browser/storage_component.h",
+    "runtime/browser/sysapps_component.cc",
+    "runtime/browser/sysapps_component.h",
+    "runtime/browser/ui/color_chooser.cc",
+    "runtime/browser/ui/color_chooser.h",
+    "runtime/browser/ui/color_chooser_android.cc",
+    "runtime/browser/ui/color_chooser_aura.cc",
```
rakuco: One thing I've just noticed is that, contrary to gyp, GN's file exclusion rules do not have entries for views and aura, which means all those files ending in `*_views.{cc,h}` and `*_aura.{cc,h}` need to be moved to the respective `use_aura`/`toolkit_views` blocks here.

BUILD.gn
```diff
@@ -0,0 +1,507 @@
...
+    "runtime/browser/android/xwalk_web_contents_view_delegate.h",
+    "runtime/browser/android/xwalk_web_resource_response.cc",
+    "runtime/browser/android/xwalk_web_resource_response.h",
+    "runtime/browser/android/xwalk_web_resource_response_impl.cc",
+    "runtime/browser/android/xwalk_web_resource_response_impl.h",
+    "runtime/browser/application_component.cc",
+    "runtime/browser/application_component.h",
+    "runtime/browser/devtools/remote_debugging_server.cc",
+    "runtime/browser/devtools/remote_debugging_server.h",
+    "runtime/browser/devtools/xwalk_devtools_frontend.cc",
```
rakuco: `xwalk.gyp` needs to manually remove a bunch of files from the list depending on whether `OS=="android"`. Since the GN way is to actually _add_ stuff to the list based on variables, I think you need to create an `if (!is_android)` block below and put these entries there.

heke123: No problem, I'll add them back. Actually they were there, but previously they are removed because we don't want any android flag in this patch.  no problem, done.

BUILD.gn
heke123: One thing I'm not sure:  do all files ending in `*_views.{cc,h}` perfectly match with `toolkit_views`?

rakuco: Yes, or at least they should (they were being automatically excluded in gyp anyway).

rakuco: Note that we're only talking about `*_views(_browsertest|_unittest).{cpp,h}`, so things like `foo_views_generic.cc` does not match and should not be moved.

heke123:
@rakuco   PR updated. Please have a look, thanks!

heke123:
@rakuco   webui_file_picker removed.

rakuco:
lgtm :+1: :tada: :balloon: :cool:

heke123:
Big Big thanks for your patient and great help!!!

PR #3820, opened 2016-07-26
[Android] Fix the crash bug of setOnTouchListener

xzhan96:
The background is XWalkView has been refactored in the commit 9808ed4.

Here correct the endless loop error which will cause crash.

BUG=XWALK-7099

(cherry picked from commit 48c14352f56ab5fbd94d46c6ff414a516590495b)


xzhan96:
@rakuco please help to merge, thanks

PR #3821, opened 2016-07-26
[Android] Decouple XWalkView from the activity

lincsoon:
Originally, XWalkView is designed to be tightly bound to the activity
that creates and holds it, and to listen the state change and activity
result, which will cause following problems:

- With the change of the container activity's life cycle, XWalkView will
  pause all timers and other components like videos when activity
  paused, which is not suited to the usage in background.
- Single XWalkView can't be shared across multiple activities.
- XWalkView must be created in an activity, it can't be used for a
  service.
- A XWalkView object will be destroyed after its container activity
  died and can't be used anymore.

To solve the problems above, we decided to decouple XWalkView from the
activity and give the developer more freedom to handle it. This patch
does following things:

- XWalkView can be created and held in any context, including a service.
- XWalkInitializer and XWalkUpdate don't have to be created in an
  activity, they support to initialize Crosswalk environment in any context.
- XWalkExternalExtensionManager doesn't handle the activity result anymore.
- The implementation of the file chooser in XWalkView is separated to
  a individual class called XWalkFileChooser. Because it needs to receive
  an activity, it can't be integrated into internal library as default
  file chooser. So move it from org.xwalk.core.internal to org.xwalk.core
  as a new API, the developer has to invoke it manually in
  XWalkUIClient.openFileChooser.
- Remove built-in extension "WifiDirect". WifiDirect extension needs an
  activity parameter, meanwhile BuiltinXWalkExtensions handles all of the
  loaded Crosswalk extensions in a static member. So the activity passed
  to WifiDirect extension will be held permanently, which may cause memory
  leak.
- Use the application context instead of the activity in
  XWalkPresentationHot to avoid memory leak.

BUG=XWALK-7002

### DIFF
--- a/runtime/android/core/src/org/xwalk/core/XWalkEnvironment.java
+++ b/runtime/android/core/src/org/xwalk/core/XWalkEnvironment.java
@@ -8,6 +8,7 @@ import android.content.Context;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.PackageManager.NameNotFoundException;
+import android.content.SharedPreferences;
 import android.os.Build;
 import android.util.Log;
 
@@ -26,9 +27,11 @@ class XWalkEnvironment {
     private static final String META_XWALK_APK_URL = "xwalk_apk_url";
     private static final String META_XWALK_VERIFY = "xwalk_verify";
 
+    private static final String PRIVATE_DATA_DIRECTORY_SUFFIX = "xwalkcore";
+    private static final String XWALK_CORE_EXTRACTED_DIR = "extracted_xwalkcore";
+    private static final String OPTIMIZED_DEX_DIR = "dex";
     private static final String PACKAGE_RE = "[a-z]+\\.[a-z0-9]+\\.[a-z0-9]+.*";
 
-    private static Context sInitializationContext;
     private static Context sApplicationContext;
 
     private static String sDeviceAbi;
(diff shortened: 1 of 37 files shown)


wuhengzhi:
In commit comment, ```XWalkPresentationHot``` -> ```XWalkPresentationHost```.

> A XWalkView object will be destroyed after its container activity died and can't be used anymore.

With this patch, XWalkView won't be destroyed if container activity died?

runtime/android/core/src/org/xwalk/core/XWalkEnvironment.java
```diff
@@ -41,23 +44,32 @@
...
+    public static SharedPreferences getSharedPreferences() {
+        return sApplicationContext.getSharedPreferences("libxwalkcore", Context.MODE_PRIVATE);
     }
 
-    public static Context getApplicationContext() {
-        return sApplicationContext;
+    public static String getPrivateDataDir() {
+        return sApplicationContext.getDir(PRIVATE_DATA_DIRECTORY_SUFFIX,
+                Context.MODE_PRIVATE).getAbsolutePath();
+
```
wuhengzhi: extra line.

runtime/android/core/src/org/xwalk/core/XWalkEnvironment.java
```diff
@@ -41,23 +44,32 @@
...
+    }
+
+    public static String getExtractedCoreDir() {
+        return sApplicationContext.getDir(XWALK_CORE_EXTRACTED_DIR,
+                Context.MODE_PRIVATE).getAbsolutePath();
+
```
wuhengzhi: ditto.

runtime/android/core/src/org/xwalk/core/XWalkFileChooser.java
```diff
@@ -0,0 +1,288 @@
...
+        return true;
+    }
+
+    /**
+     * Pass through activity result to XWalkViewInternal.
+     * See <a href="http://developer.android.com/reference/android/app/Activity.html">
+     * android.app.Activity.onActivityResult()</a>.
+     * @param requestCode passed from android.app.Activity.onActivityResult().
+     * @param resultCode passed from android.app.Activity.onActivityResult().
+     * @param data passed from android.app.Activity.onActivityResult().
```
wuhengzhi: For the param explanation, we can follow it from [onActivityResult](https://developer.android.com/reference/android/app/Activity.html#onActivityResult(int, int, android.content.Intent))

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkViewDelegate.java
```diff
@@ -181,6 +183,15 @@ private static void init(final Context context) throws IOException {
         ResourceExtractor.get(context);
 
         startBrowserProcess(context);
+
+        if (!CommandLine.getInstance().hasSwitch("disable-xwalk-extensions")) {
+            BuiltinXWalkExtensions.load(context);
+        } else {
+            XWalkPreferencesInternal.setValue(XWalkPreferencesInternal.ENABLE_EXTENSIONS, false);
+        }
```
wuhengzhi: Move these codes to ```startBrowserProcess()```?

PR #3824, opened 2016-07-27
Backport bug fixes and critical features to Crosswalk-21

lincsoon:


rakuco:
You could have listed the related JIRA tickets in the pull request message so this PR would be referenced in them. Now you'll have to do that manually for us to keep track of what is being committed into which branch.

lincsoon:
Thanks for your correction. I'll send another one.

PR #3825, opened 2016-07-27
[Windows] Presentation API : Fix bug where no screens would be listed as available.

darktears:
When creating the monitors list we need to make sure that they are
initialized first as not in use. In order to do that we're removing the
user defined DisplayInfo constructor and thanks to C++11 the field in_use
will be set to false. Before the field was not initialized by default
and was being affected a random value (which then resolve as true) and
all monitors becomes in use and the API never let you choose one.

BUG=XWALK-7098

### DIFF
--- a/runtime/browser/xwalk_presentation_service_helper.cc
+++ b/runtime/browser/xwalk_presentation_service_helper.cc
@@ -11,8 +11,6 @@
 
 namespace xwalk {
 
-DisplayInfo::DisplayInfo() {}
-
 std::unique_ptr<DisplayInfoManagerService> DisplayInfoManagerService::Create() {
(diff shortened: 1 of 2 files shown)


darktears:
@kenny-y PTAL. Android most likely need the same fix.

@rakuco

rakuco:
I think this is the wrong place to fix this. `DisplayInfo info = {}` should have been enough to initialize all members to zero, the problem is that `DisplayInfo` has a user-defined empty constructor that prevents this from happening. Either that needs to be removed (and we become more C++11-y) or the constructor needs to initialize all members to sane values.

rakuco:
lgtm

PR #3829, opened 2016-07-28
[Linux] Fix build after 630ba8e

darktears:
Chromium style requires us to define an explicit out-of-line constructor
so we can't use the C++11 trick. Let's just use old school initialization.


rakuco:
lgtm

PR #3830, opened 2016-07-29
Enable audio playback with enable-unified-media-pipeline flag.

wuhengzhi:
The 'ffmpeg_branding' was defined in 'third_party/ffmpeg/ffmpeg.gyp',
defaut value is 'chromium'. It was overridden in 'xwalk/build/common.gypi',
but the overridden value did not work since this variable coming from
different file was not at the same scope level.
Move 'ffmpeg_branding' to 'variables.conditions' to keep the same scope
level instead of 'variables.variables'.

BUG=XWALK-6874

### DIFF
--- a/build/common.gypi
+++ b/build/common.gypi
@@ -32,6 +32,14 @@
   # overridden here. The default values should match the ones in the original
   # .gyp files.
   'variables': {
+    'conditions': [
+      ['1==1', {
+        # From src/third_party/ffmpeg/ffmpeg.gyp.
+        # Whether to build the Chromium or Google Chrome version of FFmpeg (the
+        # latter contains additional codecs).
+        'ffmpeg_branding%': 'Chrome',
+      }],
+    ],
     # Putting a variables dict inside another variables dict looks kind of
     # weird. This is caused by the way GYP variable expansion and evaluation
     # works: one cannot set a variable in a scope and use it in a conditional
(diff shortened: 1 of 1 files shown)


wuhengzhi:
@lincsoon @rakuco PTAL, thanks.

lincsoon:
LGTM. @rakuco Please have a look.

rakuco:
This patch cannot be merged as-is. You'd be changing the default branding on all platforms and causing us to start shipping proprietary codecs out of the blue (in addition to not propagating the variables to all scope levels and half-removing `mediacodecs_EULA`). This change needs a larger discussion.

darktears:
Not LGTM. We can't do that.

wuhengzhi:
@rakuco @darktears Should we just enable it for android platform?
Currently, ```ffmpeg_branding``` was set to ```Chromium``` in a different scope level in ```xwalk/build/common.gypi```(please see ```third_party/ffmpeg/ffmpeg.gyp```), so it has no effect to override the upstream's setting.

tehmaestro:
Hi, I'm experiencing the following issue: https://crosswalk-project.org/jira/projects/XWALK/issues/XWALK-7307?filter=allopenissues . Is it related to this pull? Are there any suggestions on how to work around this?

rakuco:
(for future reference, the discussion is continuing in XWALK-7307)

PR #3831, opened 2016-08-01
Stop changing |enable_widevine| in gyp.

rakuco:
Changing it used to make sense because we used to enable it on Windows
by default before reverting that change. We're now always setting it to
the same value it has by default upstream.

Related BUG=XWALK-6207


rakuco:
@darktears @heke123

darktears:
LGTM

heke123:
LGTM

PR #3834, opened 2016-08-01
[Windows][Linux] Fix crash when clicking the Resources tab in the

darktears:
Inspector.

It turns out that the Renderer was crashing so the inspector would stop
working. The Resources tab lists for example the application cache, as
well as the IndexedDb or the ServiceWorker. Chromium now requires these
values to be coming from a secured origin (such as HTTPS) but in the case
of Crosswalk it was failing (and thus sending a bad IPC message which
would trigger the shutdown of the renderer) because we use the app://
protocol. This scheme is trusted so we can safely add it as such.

BUG=XWALK-7194

### DIFF
--- a/runtime/common/xwalk_content_client.cc
+++ b/runtime/common/xwalk_content_client.cc
@@ -214,6 +214,12 @@ void XWalkContentClient::AddAdditionalSchemes(
   savable_schemes->push_back(application::kApplicationScheme);
 }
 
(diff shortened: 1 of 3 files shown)


rakuco:
lgtm, but it'd be good to have a test if possible.

darktears:
Test added @rakuco, PTAL.

PR #3788, opened 2016-07-08
M52: Initialize MaterialDesignController in XWalkMainDelegate::PreSan…

mrunalk:
…dboxStartup

Recently Material design related changes were landed in upstream because of that
we need to initialize MaterialDesignController in our code. Without this some of
the browsertests were failing as tracked by XWALK-7073.

adapt to,
https://codereview.chromium.org/1878973002

PR #3835, opened 2016-08-01
Roll Chromium 52.0.2743.82

mrunalk:
This has the latest M52 stable from Chromium.
Big thanks to Maksim Sisov, Alexis Menard, Hengzhi Wu and Chuan Liu
for their help with this rebase.

The list of upstream changes we adapted to is as follows,
https://codereview.chromium.org/1920793002
https://codereview.chromium.org/1966733002
https://codereview.chromium.org/1883073003
https://codereview.chromium.org/1673103002
https://codereview.chromium.org/1727063002
https://codereview.chromium.org/1924703002
https://codereview.chromium.org/1864583006
https://codereview.chromium.org/1892893002
https://codereview.chromium.org/1727063002
https://codereview.chromium.org/1892483002
https://codereview.chromium.org/1953803002
https://codereview.chromium.org/1931043002
https://codereview.chromium.org/1895143003
https://codereview.chromium.org/1898153002
https://codereview.chromium.org/1898633004
https://codereview.chromium.org/1522823002
https://codereview.chromium.org/1864583006
https://codereview.chromium.org/1727063002
https://codereview.chromium.org/1873293002
https://codereview.chromium.org/1898633004
https://codereview.chromium.org/1878973002
https://codereview.chromium.org/1557733002
https://codereview.chromium.org/1844243002
https://codereview.chromium.org/1919043003
https://codereview.chromium.org/1785953002

BUG=XWALK-7022


mrunalk:
`Error: Command 'git merge-base HEAD origin' returned non-zero exit status 1 in /home/crwuser/build-infrastructure/slave/crosswalk-android-x86/build/src/third_party/libjingle/source/talk`

I have seen this error on my side as well and the way I fixed it was remove the 'src/third_party/libjingle/source' directory and rerun gclient sync. The directory has been removed upstream.

mrunalk:
Updated the buildtools_rev and rebased on top of the latest.

rakuco:
> I have seen this error on my side as well and the way I fixed it was remove the 'src/third_party/libjingle/source' directory and rerun gclient sync. The directory has been removed upstream.

I've had some time to investigate this today, and it's caused by gclient and how that repository's branches work. While manually removing that directory solves the problem, it doesn't scale for our users, so we either need to fix depot_tools or add a script to remove the directory (plan B). I've just filed https://bugs.chromium.org/p/chromium/issues/detail?id=633962 to discuss this with upstream.

rakuco:
> or add a script to remove the directory (plan B)

An example of what it would like is `fix-speex-checkout.py`, which was once introduced for a similar reason: https://github.com/crosswalk-project/crosswalk/pull/3109/files

rakuco:
My upstream patch https://codereview.chromium.org/2215673002/ has landed, I think the runhooks step will now work if you resubmit your PR.

mrunalk:
Awesome ! Never thought the problem would be in gclient. Great Job @rakuco :+1:

rakuco:
lgtm once the bots go green

darktears:
LGTM

runtime/renderer/xwalk_content_renderer_client.cc
```diff
@@ -312,11 +312,4 @@ void XWalkContentRendererClient::GetNavigationErrorStrings(
   }
 }
 
-void XWalkContentRendererClient::AddKeySystems(
```
rakuco: Does this mean Widevine support is no longer enabled on Android?

runtime/browser/xwalk_browser_context.cc
```diff
@@ -349,6 +315,17 @@ net::URLRequestContextGetter*
 #endif
 }
 
+net::URLRequestContextGetter* XWalkBrowserContext::CreateMediaRequestContext() {
```
rakuco: Are you sure the code removed above isn't necessary here?

runtime/app/xwalk_main_delegate.cc
```diff
@@ -76,6 +77,10 @@ void XWalkMainDelegate::PreSandboxStartup() {
       command_line->GetSwitchValueASCII(switches::kProcessType);
   InitLogging(process_type);
 #endif
+
+#if !defined(OS_ANDROID)
```
rakuco: I dont'see this ifdef in Chromium code; what happens if this function is called on Android?

runtime/browser/xwalk_browser_context.cc
mrunalk: I think so as only GetRequestContext() was used once in xwalk_devtools_frontend.cc:234 which has been replaced by `content::BrowserContext::GetDefaultStoragePartition(web_contents()->GetBrowserContext())->GetURLRequestContext()`

runtime/app/xwalk_main_delegate.cc
mrunalk: @rakuco Please refer to, https://github.com/crosswalk-project/crosswalk/pull/3788

runtime/renderer/xwalk_content_renderer_client.cc
mrunalk: I have added a commit to add Widevine support through AddSupportedKeySystems()

runtime/browser/xwalk_browser_context.cc
rakuco: Removing the code in `XWalkBrowserContext::GetMediaRequestContextForStoragePartition` for the non-Android case can potentially break commit 507013f772a8ab688aec2c68e671ab0c8e449038, can't it? To me it's similar to https://codereview.chromium.org/1892893002 for `chrome/browser/profiles/profile_impl.cc`. The Android code path may be fine with returning `nullptr`, though even in that case I'm not entirely sure.

runtime/renderer/xwalk_content_renderer_client.cc
rakuco: Thanks, please squash everything together for merging the PR.

runtime/browser/xwalk_browser_context.cc
mrunalk: Can you please be specific in what it is breaking? I don't understand what that commit is trying to do. I've just tried to adapt to the upstream changes as much as possible.

rakuco: With https://codereview.chromium.org/1892893002 in mind:
* `GetMediaRequestContextForStoragePartition()` in `android_webview/` did not do anything in the past, so `AwBrowserContext::CreateMediaRequestContextForStoragePartition()` is now returning `NULL`.
* `chrome/`'s `ProfileImpl::GetMediaRequestContextForStoragePartition()`, on the other hand, had an actual implementation, which was kept in `ProfileImpl::CreateMediaRequestContextForStoragePartition()`.
* Analogously, since `XWalkBrowserContext::GetMediaRequestContextForStoragePartition()` was not empty, I don't see why `XWalkBrowserContext::CreateMediaRequestContextForStoragePartition()` should.

mrunalk: Let's narrow down our problem area. We have 2 new functions(I'm not considering, GetMediaRequestContextForRenderProcess() as it's totally been removed from everywhere),
1. CreateMediaRequestContext()
2. CreateMediaRequestContextForStoragePartition()
 which seem like are new versions of 
a. GetMediaRequestContext()
b. GetMediaRequestContextForStoragePartition()
Out of that 'a' was returning just GetRequestContext() anyway so it makes sense to have been replaced by url_request_getter_.get().
Regarding 2 replacing b I see we are losing some logic there but if you look at shell_browser_context, CastBrowserContext, HeadlessBrowserContext, AwBrowserContext.. are all returning null_ptr except ProfileImpl and OffTheRecordProfileImp so they are returning nullptr if it is android or not. I am not sure if we should follow ProfileImpl or OffTheRecordImpl. There could be a potential issue with 2 but let's not spend too much time on what could be in there from Crosswalk point of view since neither of us know what should go in there. If something breaks we can always fix it later.

rakuco: All those Context classes you mentioned also had either an empty implementation or just called GetURLRequestContext() before in GetMediaRequestContextForStoragePartition(). ProfileImpl is the only one that had a different implementation, which was kept in the new code. I maintain that it's safer to keep the code than to remove it just because you're not sure if it's necessary -- less code being touched means a smaller surface of things that can break.

mrunalk: > I maintain that it's safer to keep the code than to remove it just because you're not sure if it's necessary -- less code being touched means a smaller surface of things that can break.

I don't understand what you want here. Are you suggesting we leave those 4 getter functions in there? We can do that only if we remove overrides on them. But if you remove overrides that will be dead code used by no one. Or are you suggesting to move the code in GetMediaRequestContextForStoragePartition() to CreateMediaRequestContextForStoragePartition()?

darktears: No of course he did not meant to remove the override. He meant to keep the the code that was in GetMediaRequestContextForStoragePartition and put it in CreateMediaRequestContextForStoragePartition just like upstream did with ProfileImpl.

mrunalk: Ok, just updated the CreateMediaRequestContextForStoragePartition(). PTAL.

rakuco: That's finally what I meant \o/

PR #3836, opened 2016-08-02
Eliminate comparison of integers of different signs, unused const variables in xwalk/application.

heke123:
Eliminate comparison of integers of different signs, unused const variables in xwalk/application.
    
    With more strict compile options like "-Wunused-const-variable" and
    "-Wsign-compare", errors are found during building:
    ../../xwalk/application/common/application_file_util.cc:173:21:
    error: comparison of integers of different signs: 'int' and
    'unsigned long' [-Werror,-Wsign-compare]
    ../../xwalk/application/common/application_file_util.cc:45:12:
    error: unused variable 'kContentKey' [-Werror,-Wunused-const-variable]
    ../../xwalk/application/common/package/xpk_package.cc:18:15:
    error: unused variable 'kSignatureAlgorithm' [-Werror,-Wunused-const-variable]
    ../../xwalk/application/extension/application_widget_storage.cc:47:12:
    error: unused variable 'kSelectTableLength' [-Werror,-Wunused-const-variable]
    
    Remove Un-used const variables, change int into size_t type.
    
    BUG=XWALK-7201


heke123:
@rakuco,  Please have a look.   Thanks.

application/common/application_file_util.cc
```diff
@@ -170,7 +168,8 @@ inline bool IsElementSupportSpanAndDir(xmlNode* root) {
 }
 
 bool IsSingletonElement(const std::string& name) {
-  for (int i = 0; i < arraysize(kSingletonElements); ++i)
```
rakuco: Huh? It looks like you didn't understand either what the `arraysize()` macro does or what is wrong with the original code. Just change `i`'s type to `size_t`.

sysapps/raw_socket/udp_socket_object.cc
```diff
@@ -273,7 +273,7 @@ void UDPSocketObject::OnSend(int status) {
 
   if (ret == net::ERR_IO_PENDING) {
     has_write_pending_ = true;
-  } else if (ret == write_buffer_size_) {
+  } else if (ret == base::checked_cast<int>(write_buffer_size_)) {
```
rakuco: This code can be rewritten without a cast. See #3838

rakuco:
The code is OK apart from the inline comments I added, the explanation in the commit/pull request message is bad:
* Why does GN have stricter build options?
* How does that affect the build (ie. how is it failing)?
* Why are certain variables no longer needed (ie. which commit stopped referencing them)?

Personally, I would not even mention GN here since these are bug fixes regardless of whether we are moving to GN or not. Just mention the errors you are getting and when the unused variables became unused. See 5ef3bba25287fd98c7d91d720e9471b0a8b9c46e and 1cc332261988b329b2fa468c03fcbdfd1a0536c7 for commits listing build failures, and 13df79034d421fbcc4853c120632c0414da5a2ea for a commit that shows a build failure and explain where it comes from.

heke123:
@rakuco  Thanks for your comments, please help to have a look.

rakuco:
I'd fix each issue separately and mention which commit stopped using those unused variables, but lgtm

PR #3838, opened 2016-08-02
sysapps: Rewrite the check for UDPSocket::SendTo's return value.

rakuco:
Drop the check for whether it has returned the number of bytes we are
trying to send (which includes a comparison between a size_t and an
int) since it does not make much sense. We only need to contemplate 3
different cases:

1. SendTo returns ERR_IO_PENDING, which means it is still transmitting
   all the data.
2. SendTo returns another negative value, so we just bail out.
3. SendTo returns a positive value, which means everything was
   transmitted, otherwise ERR_IO_PENDING would have been returned.

### DIFF
--- a/sysapps/raw_socket/udp_socket_object.cc
+++ b/sysapps/raw_socket/udp_socket_object.cc
@@ -271,15 +271,17 @@ void UDPSocketObject::OnSend(int status) {
       addresses_[0],
       base::Bind(&UDPSocketObject::OnWrite, base::Unretained(this)));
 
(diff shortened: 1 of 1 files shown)


rakuco:
@darktears @heke123

darktears:
LGTM

heke123:
LGTM
I don't think base::checked_cast is unacceptable, Line 270 has another base::checked_cast.

rakuco:
Nobody said it's unacceptable, it's just unnecessary if the code can be rewritten in a more clear manner.

PR #3839, opened 2016-08-03
Just use id instead of calling host->GetID() again

mrunalk:
simple fix


rakuco:
What is it supposed to fix? Is there a related ticket?

mrunalk:
It's just a simple optimization I noticed while working on Print feature.

rakuco:
While there's nothing wrong with the idea, the new code will break the build if NaCl is disabled. For example, I do my Linux builds with `component=shared_library disable_nacl=1` and get this with your patch:

```
../../xwalk/runtime/browser/xwalk_content_browser_client.cc:171:7: error: unused variable 'id' [-Werror,-Wunused-variable]
  int id = host->GetID();
```

mrunalk:
Agreed, not worth it then.

PR #3842, opened 2016-08-08
Change locale path from "locales" to "locales/xwalk"

heke123:
Both Crosswalk and Chrome they place the en-US.pak into folder
"locales" in build of Linux and windows. Potientially there is
conflicts, so it's better to set Crosswalk's locale-path as
"locales/xwalk".
Add a "xwalk_resource_delegate" class which indicates the right
locale pack file to load for Crosswalk.

### DIFF
--- a/runtime/app/xwalk_main_delegate.cc
+++ b/runtime/app/xwalk_main_delegate.cc
@@ -19,6 +19,7 @@
 #include "xwalk/runtime/common/logging_xwalk.h"
 #include "xwalk/runtime/common/paths_mac.h"
 #include "xwalk/runtime/common/xwalk_paths.h"
+#include "xwalk/runtime/common/xwalk_resource_delegate.h"
 #include "xwalk/runtime/renderer/xwalk_content_renderer_client.h"
 
 #if !defined(DISABLE_NACL) && defined(OS_LINUX)
@@ -116,7 +117,6 @@ void XWalkMainDelegate::ZygoteStarting(
 
 #endif  // defined(OS_POSIX) && !defined(OS_ANDROID)
 
-// static
 void XWalkMainDelegate::InitializeResourceBundle() {
   base::FilePath pak_file;
   base::FilePath pak_dir;
@@ -129,8 +129,10 @@ void XWalkMainDelegate::InitializeResourceBundle() {
 #endif
 
 #if !defined(OS_ANDROID)
+  resource_delegate_.reset(new XWalkResourceDelegate());
   ui::ResourceBundle::InitSharedInstanceWithLocale(
-      "en-US", nullptr, ui::ResourceBundle::DO_NOT_LOAD_COMMON_RESOURCES);
+      "en-US", resource_delegate_.get(),
+      ui::ResourceBundle::DO_NOT_LOAD_COMMON_RESOURCES);
   pak_file = pak_dir.Append(FILE_PATH_LITERAL("xwalk.pak"));
   ResourceBundle::GetSharedInstance().AddDataPackFromPath(
       pak_file, ui::SCALE_FACTOR_NONE);
--- a/runtime/app/xwalk_main_delegate.h
+++ b/runtime/app/xwalk_main_delegate.h
@@ -15,6 +15,7 @@
 namespace xwalk {
 
 class XWalkRunner;
+class XWalkResourceDelegate;
 
 class XWalkMainDelegate : public content::ContentMainDelegate {
  public:
@@ -35,12 +36,13 @@ class XWalkMainDelegate : public content::ContentMainDelegate {
   content::ContentBrowserClient* CreateContentBrowserClient() override;
   content::ContentRendererClient* CreateContentRendererClient() override;
 
-  static void InitializeResourceBundle();
-
  private:
+  void InitializeResourceBundle();
+
   std::unique_ptr<XWalkRunner> xwalk_runner_;
   std::unique_ptr<content::ContentRendererClient> renderer_client_;
   std::unique_ptr<content::ContentClient> content_client_;
+  std::unique_ptr<XWalkResourceDelegate> resource_delegate_;
 
   DISALLOW_COPY_AND_ASSIGN(XWalkMainDelegate);
 };
--- a/runtime/common/xwalk_resource_delegate.cc
+++ b/runtime/common/xwalk_resource_delegate.cc
@@ -1,0 +1,61 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "xwalk/runtime/common/xwalk_resource_delegate.h"
+
+#include "base/files/file_path.h"
+#include "base/path_service.h"
+#include "ui/gfx/image/image.h"
+
+namespace xwalk {
+
+XWalkResourceDelegate::XWalkResourceDelegate() {}
+
+XWalkResourceDelegate::~XWalkResourceDelegate() {}
+
+base::FilePath XWalkResourceDelegate::GetPathForResourcePack(
+    const base::FilePath& pack_path,
+    ui::ScaleFactor scale_factor) {
+  return pack_path;
+}
+
+base::FilePath XWalkResourceDelegate::GetPathForLocalePack(
+    const base::FilePath& pack_path,
+    const std::string& locale) {
+  base::FilePath product_dir;
+  if (!PathService::Get(base::DIR_MODULE, &product_dir)) {
+    NOTREACHED();
+  }
+  return product_dir.
+      Append(FILE_PATH_LITERAL("locales")).
+      Append(FILE_PATH_LITERAL("xwalk")).
+      AppendASCII(locale + ".pak");
+}
+
+gfx::Image XWalkResourceDelegate::GetImageNamed(int resource_id) {
+  return gfx::Image();
+}
+
+gfx::Image XWalkResourceDelegate::GetNativeImageNamed(int resource_id) {
+  return gfx::Image();
+}
+
+base::RefCountedStaticMemory* XWalkResourceDelegate::LoadDataResourceBytes(
+    int resource_id,
+    ui::ScaleFactor scale_factor) {
+  return nullptr;
+}
+
+bool XWalkResourceDelegate::GetRawDataResource(int resource_id,
+                                               ui::ScaleFactor scale_factor,
+                                               base::StringPiece* value) {
+  return false;
+}
+
+bool XWalkResourceDelegate::GetLocalizedString(int message_id,
+                                               base::string16* value) {
+  return false;
+}
+
+}  // namespace xwalk
--- a/runtime/common/xwalk_resource_delegate.h
+++ b/runtime/common/xwalk_resource_delegate.h
@@ -1,0 +1,57 @@
+// Copyright (c) 2016 Intel Corporation. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef XWALK_RUNTIME_COMMON_XWALK_RESOURCE_DELEGATE_H_
+#define XWALK_RUNTIME_COMMON_XWALK_RESOURCE_DELEGATE_H_
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "base/macros.h"
+#include "base/memory/ref_counted_memory.h"
+#include "ui/base/resource/resource_bundle.h"
+
+namespace base {
+class FilePath;
+}
+
+namespace gfx {
+class Image;
+}
+
+namespace xwalk {
+
+// In Chrome, the locales path is set as "locales/" which potentially conflicts
+// with it is in Crosswalk. So the purpose of XWalkResourceDelegate is to set
+// the locale pack path of Crosswalk as "locales/xwalk/".
+class XWalkResourceDelegate : public ui::ResourceBundle::Delegate {
+ public:
+  XWalkResourceDelegate();
+  ~XWalkResourceDelegate() override;
+
+  // ui:ResourceBundle::Delegate implementation:
+  base::FilePath GetPathForResourcePack(
+      const base::FilePath& pack_path,
+      ui::ScaleFactor scale_factor) override;
+  base::FilePath GetPathForLocalePack(
+      const base::FilePath& pack_path,
+      const std::string& locale) override;
+  gfx::Image GetImageNamed(int resource_id) override;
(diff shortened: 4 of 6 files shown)


heke123:
@rakuco,  I changed the locales path for GYP build first.  If it's ok, I can continue the GN part. Please help to have a look, Thanks.

rakuco:
See my comment in crosswalk-project/chromium-crosswalk#371: it can be dropped provided you implement a Delegate in this pull request. You also need to improve your `git grep` skills, I can clearly see several other locations that assume the locales are in `locales/`.

heke123:
@rakuco  Thanks for your advise.  PR is updated, totally same as how chromecast does.

application/common/application_resource.cc
```diff
@@ -14,7 +14,7 @@ namespace xwalk {
 namespace application {
 namespace {
 const base::FilePath::StringType WGT_LOCALE_DIRECTORY =
-    FILE_PATH_LITERAL("locales");
+    FILE_PATH_LITERAL("locales/xwalk");
```
rakuco: You're supposed to check the code you are changing instead of blindingly replacing everything. This looks like the implementation of the wgt spec, so it has nothing to do with this pull request.

runtime/app/xwalk_main_delegate.cc
```diff
@@ -129,8 +129,11 @@ void XWalkMainDelegate::InitializeResourceBundle() {
 #endif
 
 #if !defined(OS_ANDROID)
+  resource_delegate_.reset(new XWalkResourceDelegate());
   ui::ResourceBundle::InitSharedInstanceWithLocale(
-      "en-US", nullptr, ui::ResourceBundle::DO_NOT_LOAD_COMMON_RESOURCES);
+      "en-US", resource_delegate_.get(),
+      ui::ResourceBundle::DO_NOT_LOAD_COMMON_RESOURCES);
+
```
rakuco: This empty line is unnecessary.

runtime/common/xwalk_resource_delegate.cc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+bool XWalkResourceDelegate::GetRawDataResource(int resource_id,
+                                              ui::ScaleFactor scale_factor,
+                                              base::StringPiece* value) {
+  return false;
+}
+
+bool XWalkResourceDelegate::GetLocalizedString(int message_id,
+                                              base::string16* value) {
```
rakuco: Wrong indentation.

runtime/common/xwalk_resource_delegate.cc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+base::RefCountedStaticMemory* XWalkResourceDelegate::LoadDataResourceBytes(
+    int resource_id,
+    ui::ScaleFactor scale_factor) {
+  return NULL;
+}
+
+bool XWalkResourceDelegate::GetRawDataResource(int resource_id,
+                                              ui::ScaleFactor scale_factor,
```
rakuco: Wrong indentation.

runtime/common/xwalk_resource_delegate.cc
```diff
@@ -0,0 +1,79 @@
...
+                                              base::StringPiece* value) {
```
rakuco: Wrong indentation.

tools/installer/common/installer.include
```diff
@@ -142,9 +142,9 @@ stage_install_common() {
   fi
 
   # l10n paks
-  cp -a "${BUILDDIR}/locales" "${STAGEDIR}/${INSTALLDIR}/"
-  find "${STAGEDIR}/${INSTALLDIR}/locales" -type f -exec chmod 644 '{}' \;
-  find "${STAGEDIR}/${INSTALLDIR}/locales" -type d -exec chmod 755 '{}' \;
+  cp -a "${BUILDDIR}/locales/xwalk" "${STAGEDIR}/${INSTALLDIR}/"
```
rakuco: You're likely creating a directory called `${STAGEDIR}/${INSTALLDIR}/xwalk` now, not `${STAGEDIR}/${INSTALLDIR}/locales/xwalk`.

runtime/common/xwalk_resource_delegate.h
```diff
@@ -0,0 +1,59 @@
...
+class FilePath;
+}
+
+namespace gfx {
+class Image;
+}
+
+namespace xwalk {
+
+// A singleton resource bundle delegate. Primary purpose is to indicate the
```
rakuco: The explanation here needs to be much more meaningful. Put yourself in the shoes of someone who will look at this code 2 years from now: it is still unclear why the delegate is needed (isn't there code in `ui` doing that automatically? what's wrong with locales in `locales/`?)

runtime/app/xwalk_main_delegate.h
```diff
@@ -35,12 +36,13 @@ class XWalkMainDelegate : public content::ContentMainDelegate {
   content::ContentBrowserClient* CreateContentBrowserClient() override;
   content::ContentRendererClient* CreateContentRendererClient() override;
 
-  static void InitializeResourceBundle();
```
rakuco: Why?

heke123: The member `"resource_delegate_"` is added and referenced in member-function `InitializeResourceBundle()`.  So it cannot be static unless we define` "resource_delegate_"` as static too.  So I refered the similar usage from chromecast/app/cast_main_delegate.h

tools/installer/common/installer.include
heke123: Yes, mistake here. I think we don't need to change this file.  Its original logic works fine with "/locales/xwalk".

heke123:
@rakuco   Bother you to have a look again. Thanks.

runtime/common/xwalk_resource_delegate.cc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+XWalkResourceDelegate::XWalkResourceDelegate() {
+  DCHECK(!g_instance) << "Cannot initialize resource bundle delegate twice.";
+  g_instance = this;
+}
+
+XWalkResourceDelegate::~XWalkResourceDelegate() {
+  DCHECK_EQ(g_instance, this);
+  g_instance = NULL;
```
wuhengzhi: ditto.

runtime/common/xwalk_resource_delegate.cc
```diff
@@ -0,0 +1,79 @@
...
+}
+
+gfx::Image XWalkResourceDelegate::GetNativeImageNamed(int resource_id) {
+  return gfx::Image();
+}
+
+base::RefCountedStaticMemory* XWalkResourceDelegate::LoadDataResourceBytes(
+    int resource_id,
+    ui::ScaleFactor scale_factor) {
+  return NULL;
```
wuhengzhi: ditto.

runtime/app/xwalk_main_delegate.cc
```diff
@@ -18,6 +18,7 @@
 #include "xwalk/runtime/browser/xwalk_runner.h"
 #include "xwalk/runtime/common/logging_xwalk.h"
 #include "xwalk/runtime/common/paths_mac.h"
+#include "xwalk/runtime/common/xwalk_resource_delegate.h"
```
wuhengzhi: alpha sort?

runtime/common/xwalk_resource_delegate.h
```diff
@@ -0,0 +1,60 @@
...
+}
+
+namespace xwalk {
+
+// In Chrome,the locales path is set as "locales/" which potentially conflicts
+// with it is in Crosswalk. So purpose of XWalkResourceDelegate is to set the
+// locale pack path of Crosswalk as "locales/xwalk/".
+class XWalkResourceDelegate : public ui::ResourceBundle::Delegate {
+ public:
+  // Returns the singleton of delegate. It doesn't create an instance.
```
rakuco: I don't see this being called anywhere.

runtime/common/xwalk_resource_delegate.h
```diff
@@ -0,0 +1,60 @@
...
+class FilePath;
+}
+
+namespace gfx {
+class Image;
+}
+
+namespace xwalk {
+
+// In Chrome,the locales path is set as "locales/" which potentially conflicts
```
rakuco: you're missing a space before "the locales".

runtime/common/xwalk_resource_delegate.h
```diff
@@ -0,0 +1,60 @@
...
+// with it is in Crosswalk. So purpose of XWalkResourceDelegate is to set the
```
rakuco: "the purpose"

xwalk.gyp
```diff
@@ -632,8 +634,11 @@
       ],
       'copies': [
```
rakuco: Are you sure this is still needed (ie. changing the output directory above in `xwalk_strings` is not enough)?

runtime/common/xwalk_resource_delegate.cc
```diff
@@ -0,0 +1,79 @@
...
+
+#include "base/files/file_path.h"
+#include "base/path_service.h"
+#include "ui/gfx/image/image.h"
+
+namespace xwalk {
+
+namespace {
+
+XWalkResourceDelegate* g_instance = NULL;
```
heke123: @wuhengzhi   I found in xwalk/runtime/  the usage of "nullptr" and "NULL" is half and half.  Is there  any policy to follow when to use nullptr and NULL?
It is a good point, thanks for pointing out this.

runtime/app/xwalk_main_delegate.cc
heke123: yes, my mistake. should be alpha sort.

runtime/common/xwalk_resource_delegate.h
heke123: Yes you are right, it is not referenced anywhere.  When I "copy" this part code from the src/chromecast, I noticed in cast_resource_delegate of the chromecast, the getinstance() is defined but not referenced either.  
My understanding is:  Since it is designed as a singleton pattern by using the g_instance member,  It is good to provide interface to get this instance. Although it is true we don't reference it so far. That's why I didn't remove it.

xwalk.gyp
heke123: In above `generate_xwalk_string` action, .pak files are generated into subfolder of "gen" first. Then in target "copies" we decide what is the copy destination.
I think it is ok to folllow the old logic,  no matter what the new destination is.

heke123:
@rakuco  @wuhengzhi   I updated the PR based on comments but didn't modify the (1)getinstance() interface and (2)gen_xwalk_string logic in xwalk.gyp.  
Thanks very much.

runtime/common/xwalk_resource_delegate.h
rakuco: Not only are you introducing a dead method that is never called (and thus the singleton pattern is never used), but the `DCHECK`s in the constructor and destructor means Crosswalk will crash every time in debug mode since `g_instance` will be null.

xwalk.gyp
rakuco: The pak files are being generated into `<(SHARED_INTERMEDIATE_DIR)/xwalk/locales` only because the value of `grit_out_dir` is set to that; what I'm asking is if there's any issue with setting it to `<(PRODUCT_DIR)/locales/xwalk` instead, so that you don't need to copy anything afterwards.

runtime/common/xwalk_resource_delegate.cc
rakuco: > Is there any policy to follow when to use nullptr and NULL?

See "Null Pointer Constant" in https://chromium-cpp.appspot.com/

runtime/common/xwalk_resource_delegate.h
heke123: In constructor it is `DCHECK(!g_instance) << "Cannot .... twice"` so it won't crash unless we create two instances. But I agree it is not good that we provide singleton interface but never use it in code.
Do you think it is enough just to remove the `GetInstance()` dead method? Or how about totally remove the singleton part to make it a simple class?

rakuco: Oops, I misread the logic and it's not going to crash indeed. In any case, I think the entire singleton logic can be removed since it's not used.

xwalk.gyp
heke123: Understand now. I noticed if we change the `grit_out_dir`, not only the `*.pak` files are generated in new destination, but also the new generated folder `grit` which contains` xwalk_string.h`.
So let's just keep the old logic to make it easy.

runtime/common/xwalk_resource_delegate.h
heke123: No problem, I'll remove the singleton.

heke123:
@rakuco  Singleton removed. Please have a review. thanks.

rakuco:
lgtm, thank you for working on this

rakuco:
Since this broke the Crosswalk Windows build, I'll have to revert the PR for now.

rakuco:
Done in #3848

rakuco:
@heke123: fixing this is fairly easy.

What's happening here is that gyp is smart enough to know that when you reference a directory such as `<(PRODUCT_DIR)/locales` as an input for an action and another action (`xwalk_strings`) is known to produce that same directory as an output then the former should depend on the latter.

This is why the `generate_crosswalk_win_zip` action does not have any explicit dependency on `xwalk_strings` or any of the targets that generate the files it depends on (`icudtl.dat`, `osmesa.dll` etc).

The problem with the original patch in this pull request is that the `xwalk_strings` target still generates the locales in `gen/xwalk/locales` while you're now manually copying each file to `locales/xwalk` instead of copying the entire directory, so gyp cannot know that it should depend on that target.

Basically, if you set `grit_out_dir` here to `locales/xwalk` instead of `xwalk/locales` you can then change your `copies` section to use `<(PRODUCT_DIR)/locales/` as the destination and keep the list of `files` unchanged (which has the additional benefit that you do not need to hardcode all the locales you want to copy).

heke123:
@rakuco  If we set `grit_out_dir` to `locales/xwalk`, the `grit/xwalk_strings.h` will be generated into `locales/xwalk `too. Then we have to find some way to move `grit/xwalk_strings.h` back to `gen/xwalk/grit/` OR we have to change the `include_dirs` of some targets who includes `xwalk_strings.h`

So my solution is to change the `"outputs"` into `'<(PRODUCT_DIR)/locales/xwalk'` to make `generate_crosswalk_win_zip` aware the new locale path. Do you think it's ok? PR updated. please have a look.  Big thanks for your help!

PR #3843, opened 2016-08-08
Add copy-constructors to some classes to fix the sysroot build.

rakuco:
The two commits in this pull request are required to fix the Linux build
when `use_sysroot` is enabled (which we use for our release builds)
after moving to M52.

Something in the old libstdc++ classes in the Debian Wheezy sysroot end
up causing Chromium's chromium-style clang plugin to complain; when
`use_sysroot` is disabled (as is the case by default and what our
Buildbot slaves build), the libstdc++ version used is much more recent
and does not trigger Chromium's plugin.

BUG=XWALK-7022


darktears:
LGTM

PR #3844, opened 2016-08-12
Eliminate comparisons between size_t and int, different signs in unittest code

heke123:
With "-Wsign-compare" build option, errors are found:
error: comparison of integers of different signs: 'const unsigned long'
and 'const int' [-Werror,-Wsign-compare]
error: comparison of integers of different signs: 'const int' and
'const unsigned int' [-Werror,-Wsign-compare]


heke123:
@rakuco  Please have a review, Thanks.

rakuco:
lgtm, thank you!

PR #3848, opened 2016-08-15
Revert "Change locale-path from "locales" to "locales/xwalk""

rakuco:
This reverts commit bfe53fddb980821fa23569913d9e5689043da0ff, which
broke the Crosswalk Windows build:

```
ninja: error: 'locales/xwalk', needed by 'crosswalk_win.zip', missing and no known rule to make it
```


rakuco:
All bots are green, and this fixes a build failure in a Tier-1 platform, so merging.

PR #3852, opened 2016-08-17
[Android] Refine the default values of mUseWideViewport and mLoadWith…

xzhan96:
…OverviewMode

XWalk enable the wide viewport quirk in 2e9640b and implemented corresponding APIs
in ebfc438, and the default values of mUseWideViewport and mLoadWithOverviewMode were
set to false to keep the same with Android WebView.

However, XWalk has the different usage in H5 game development, it no need to require
developer set those valuse via APIs, and it's better to keep the same with Chromium
instead of WebView, so set the default values to true will have benefit and convenience.

BUG=XWALK-6995
BUG=XWALK-6963
BUG=XWALK-7051


xzhan96:
@lincsoon @wuhengzhi please help to review, thanks

PR #3855, opened 2016-08-18
M53 rebasing.

msisov:
Please review


msisov:
sorry, did not have time on Friday to continue rebasing.

extensions/common/xwalk_external_extension.h
```diff
@@ -39,8 +40,9 @@ class XWalkExternalExtension : public XWalkExtension {
 
   bool Initialize();
 
-  void set_runtime_variables(const base::ValueMap& runtime_variables) {
-    runtime_variables_ = runtime_variables;
+  void set_runtime_variables(const base::DictionaryValue::Storage& runtime_variables) {
+    runtime_variables_.emplace_hint(runtime_variables_.end(), runtime_variables.begin()->first,
```
mrunalk: Don't understand why you need to use emplace_hint here. Don't find any comparable use in the upstream patch you referenced in the commit message.

extensions/common/xwalk_extension_server.cc
```diff
@@ -360,10 +360,9 @@ void XWalkExtensionServer::Invalidate() {
 
 namespace {
 base::FilePath::StringType GetNativeLibraryPattern() {
-  const base::string16 library_pattern = base::GetNativeLibraryName(
-      base::UTF8ToUTF16("*"));
+  base::string16 library_pattern = base::ASCIIToUTF16(base::GetNativeLibraryName("*"));
```
mrunalk: base::GetNativeLibraryName() already returns as UTF16 string so don't understand why you need to convert from ASCII to UTF16 here.

runtime/common/xwalk_content_client.h
```diff
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include "base/compiler_specific.h"
+#include "base/memory/ref_counted_memory.h"
```
mrunalk: Use forward declaration for RefCountedStaticMemory instead of #include here. Refer to Chromium C++ guidelines.

extensions/common/xwalk_external_extension.h
msisov: It was a reference to the ValueMap. Second is unique_ptr now and it is not possible to use operator= here.

msisov: sorry, I hurried. I guess swap will be better here.

extensions/common/xwalk_extension_server.cc
msisov: it returns std::string. Which is UTF8, isn't it?

runtime/common/xwalk_content_client.h
msisov: Ops, that was a wrong return type

extensions/common/xwalk_extension_server.cc
```diff
@@ -360,12 +360,11 @@ void XWalkExtensionServer::Invalidate() {
 
 namespace {
 base::FilePath::StringType GetNativeLibraryPattern() {
-  const base::string16 library_pattern = base::GetNativeLibraryName(
-      base::UTF8ToUTF16("*"));
+  const std::string library_pattern = base::GetNativeLibraryName("*");
```
mrunalk: The upstream patch you have referenced above has string16 but looks like there was another [change](https://codereview.chromium.org/2048523002) made upstream later which has std::string. Anyway, this looks OK to me.

extensions/common/xwalk_external_extension.h
```diff
@@ -53,7 +54,7 @@ class XWalkExternalExtension : public XWalkExtension {
 
   // Variables from the browser process. Usually things like currently-running
   // application ID.
-  base::ValueMap runtime_variables_;
+  base::DictionaryValue::Storage runtime_variables_;
```
mrunalk: I'm still not sure about the use of DictionaryValue::Storage as I don't see it being used anywhere upstream. May I know why you chose DictionaryValue::Storage to replace ValueMap.

extensions/common/xwalk_external_extension.h
```diff
@@ -39,8 +40,8 @@ class XWalkExternalExtension : public XWalkExtension {
 
   bool Initialize();
 
-  void set_runtime_variables(const base::ValueMap& runtime_variables) {
-    runtime_variables_ = runtime_variables;
+  void set_runtime_variables(base::DictionaryValue::Storage* runtime_variables) {
+      runtime_variables_.swap(*runtime_variables);
```
mrunalk: Logically it is ok to use swap but I don't think we need to make setter function that complicated. Wouldn't simply assigning the value or using std::move work here?

msisov: Neither of them work. The thing is that there is base::ValueMap now. There is a using Storage = std::map<std::string, std::unique_ptr<Value>>; now. That is why I am using base::DictionaryValue::Storage now. 

What is more, it cannot be assigned (../../xwalk/extensions/common/xwalk_external_extension.h:45:23: error: no viable overloaded '='
   runtime_variables_ = runtime_variables;) or moved as long as it is not known how to copy unique_ptr unless operator= is overloaded.

extensions/common/xwalk_extension_server.cc
msisov: Oh, I considered that one and mentioned in comments another one.

extensions/common/xwalk_external_extension.h
msisov: https://codereview.chromium.org/2000803003/diff/160001/base/values.h lines 46 and 210

mrunalk:
LGTM. Thanks for fixing all these issues.

PR #3857, opened 2016-08-18
[Android] Decouple XWalkView from the activity

lincsoon:
Originally, XWalkView is designed to be tightly bound to the activity
that creates and holds it, and to listen the state change and activity
result, which will cause following problems:

- With the change of the container activity's life cycle, XWalkView will
  pause all timers and other components like videos when activity
  paused, which is not suited to the usage in background.
- Single XWalkView can't be shared across multiple activities.
- XWalkView must be created in an activity, it can't be used for a
  service.
- A XWalkView object will be destroyed after its container activity
  died and can't be used anymore.

To solve the problems above, we decided to decouple XWalkView from the
activity and give the developer more freedom to handle it. This patch
does following things:

- XWalkView can be created and held in any context, including a service.
- XWalkInitializer and XWalkUpdate don't have to be created in an
  activity, they support to initialize Crosswalk environment in any context.
- XWalkExternalExtensionManager doesn't handle the activity result anymore.
- The implementation of the file chooser in XWalkView is separated to
  a individual class called XWalkFileChooser. Because it needs to receive
  an activity, it can't be integrated into internal library as default
  file chooser. So move it from org.xwalk.core.internal to org.xwalk.core
  as a new API, the developer has to invoke it manually in
  XWalkUIClient.openFileChooser.
- Use the application context instead of the activity in
  XWalkPresentationHot and BuiltinXWalkExtensions to avoid memory leak.

BUG=XWALK-7002
BUG=XWALK-7188
BUG=XWALK-7189
BUG=XWALK-7190
BUG=XWALK-7191
BUG=XWALK-7192
BUG=XWALK-7193
BUG=XWALK-7197
BUG=XWALK-7198
BUG=XWALK-7199


rakuco:
I don't understand why this was merged so quickly and without any review if we're just going to start Crosswalk 23's life cycle. #3821 was reverted because it caused a lot of regressions, yet the pull request message is identical to the previous one and says zero about how this new version avoids breaking API and does not cause the regressions previously reported.

baleboy:
Given that the branch is today, this is too risky, can we merge it in master after 22 has branched? Also, please don't merge PRs without reviews, thanks!

rakuco:
22.52.561.0 will likely not include it anyway, but even if it does we can just revert the commit in the branch (with an explanation), so branching is still going to work.

lincsoon:
@rakuco this pull request has been reviewed previously, but then caused some bugs and was reverted. This update is just to fix those bugs.

rakuco:
> @rakuco this pull request has been reviewed completely previously, but then it caused some bugs and was reverted. The difference between this update and previous pull request is just bug fixing.

That's one of the things I'm complaining about: how is anyone supposed to know what changed if you send a pull request with the exact same message and merge it shortly after it has been uploaded? It's impossible to know how you fixed those bugs (which were not minor) and whether the fixes make sense and how they interact with the rest of the code. Just look at any sane open source project in the wild: whenever a patch series is sent more than once, there **must** be a list of changes between each version for other people to understand what's going on.

> About the risk, this feature has been being developed from Crosswalk-21, and I think it's already stable enough after the improvement throughout Crosswalk-22. Even if it may cause new regressions, it's safe to fix it during Crosswalk-22 beta.

Every time we delay our releases it's because something is wrong with Android and it takes forever to fix and verify, be it because of bugs caused by a rebase or in the code itself. Introducing such a big change (whose impact in terms of API breaks and whatnot looks uncertain) at this stage of Crosswalk 22's life cycle (when it should only be receiving bug fixes) sounds too risky to me.

lincsoon:
OK, I understand your concern. Let's skip Crossawlk-22 and verify it on Crossawlk-23.

For some big commits like this, just reviewing is not enough to find problems. It can help to check the code style and code logic, but can not examine potential bugs. About how other ones interact with the rest of the code, I was shareing my update at the internal sync meeting every week, so at least Crosswalk Android team in PRC is aware of the status and the difference from previous pull request, so it's not a surprise for them to see this commit was merged shortly. I know it may be a surprise to the ones outside of PRC team, but I really haven't seen anybody helped to review every of our pull request actively before. 

And as I said before, this commit is totally a feature implementation and not a bug fix. If previous pull reqeust hasn't been reverted, I must list all changes and related bugs in the message of this commit. But after previous pull reqest was reverted, all bugs it caused have gone from the master branch, it doesn't make sense to fix unexisted bugs. I really cann't see the point to mention how I make a feature stable in the commit message of the feature itself.

Anyway, thanks for your attention to this pull request. I'll make sure more standard development process and more communication across teams before critical changes in the future.

rakuco:
> About how other ones interact with the rest of the code, I was shareing my update at the internal sync meeting every week, so at least Crosswalk Android team in PRC is aware of the status and the difference from previous pull request, so it's not a surprise for them to see this commit was merged shortly.

This exactly the kind of thing that has no reason to be kept behind closed doors: there are no public records of those discussions and basically everybody is kept in the dark. Updating people in private sync meetings is fine as long as the actual code review and code discussions happen where they should happen: in the pull request.

> I know it may be a surprise to the ones outside of PRC team, but I really haven't seen anybody helped to review every of our pull request actively before.

The Android code is usually maintained by people in PRC, but that does not mean everybody else is oblivious to the work. This is the case here for me at least, especially since I had to revert the original pull request and was also taking part in the other ticket about API compatibility. I've just been kept in the dark here.

> I really cann't see the point to mention how I make a feature stable in the commit message of the feature itself.

I'm not sure I understand what you're trying to say, so let me try to rephrase what I said before: in the end, the commit message itself does not need to contain "a description of how to make itself stable", but the pull request message (or a comment etc) needs to say "hey guys, I'm aware that the previous PR was reverted because of regressions, here's how this patch differs from the previos one". Basically, other people need to be aware somehow of the iterations on your code.

darktears:
@lincsoon to put in simple word, right now from this commit and the one that was reverted (which introduces breakage which is ok and not the point of the discussion here) there is no way one can figure out what changed to make the feature works better. You need to compare manually the diff to figure out and this is the point @rakuco is making here. "Internal" sync meetings are not enough because someone may pass after you on this commit and commit message needs to be explanatory so you don't need to rely on "internal sync meetings" to figure out the changes. 

And it's not because the whole team doesn't review Android PR that nobody touches the Android code in Crosswalk, keep that in mind.

lincsoon:
Thanks all the people for the concerns for this pull request. I didn't realized that this commit will attract so much attentions and got it done quickly without giving everybody enough notice. I'm sorry for that. Please trust me that I don't want to keep any work in the dark, instead I really want to make it public to meet the challenge so that our result would be more solid. But when we are not getting attention, I'm a bit sloppy with sharing information. I'll be careful and make sure enough communication in any ways including the commit message in the future.

xdevs23:
Is the drop down menu and date picker issue related to this? I guess it is...

PR #3859, opened 2016-08-19
Update release notes for Crosswalk 20 stable

baleboy:
SSIA. Note that cherry-pick alone wasn't enough because of some conflicts


baleboy:
@rakuco

PR #3861, opened 2016-08-25
[Android] Stop passing -Wl,--no-fatal-warnings when linking libxwalkcore

rakuco:
This was added back in 2013 with commit dd2dbba7 ("[Android] Merge
initial files for native library (libxwalkcore) build") and is no longer
necessary.


rakuco:
@darktears @lincsoon

darktears:
LGTM

PR #3862, opened 2016-08-25
Remove the WebUI File Picker code.

rakuco:
This was only enabled by default on Tizen, which means this code has not
been built or maintained for at least a year (it currently does not
build).

Drop all the code following the discussion in
https://lists.crosswalk-project.org/pipermail/crosswalk-dev/2016-August/003883.html

BUG=XWALK-950


rakuco:
@darktears @joone

darktears:
LGTM

PR #3865, opened 2016-08-29
Add GN build files of web_notification, xwalk_installer for Crosswalk-Linux

heke123:
Add web_notification, xwalk_installer part of Crosswalk-Linux.
With this PR, GN conversion for Crosswalk-Linux is done.

BUG = XWALK-7180

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -22,6 +22,15 @@ group("xwalk_builder") {
   }
 }
 
+if (is_linux) {
+  group("xwalk_installer") {
+    testonly = true
+    deps = [
+      "//xwalk/tools/installer:xwalk_installer_package_deb",
+    ]
+  }
+}
+
 executable("xwalk") {
   sources = [
     "runtime/app/xwalk_main.cc",
@@ -198,6 +207,12 @@ source_set("xwalk_runtime") {
     "runtime/browser/xwalk_content_settings.h",
     "runtime/browser/xwalk_form_database_service.cc",
     "runtime/browser/xwalk_form_database_service.h",
+    "runtime/browser/xwalk_notification_manager_linux.cc",
+    "runtime/browser/xwalk_notification_manager_linux.h",
+    "runtime/browser/xwalk_notification_manager_win.cc",
+    "runtime/browser/xwalk_notification_manager_win.h",
+    "runtime/browser/xwalk_notification_win.cc",
+    "runtime/browser/xwalk_notification_win.h",
     "runtime/browser/xwalk_permission_manager.cc",
     "runtime/browser/xwalk_permission_manager.h",
     "runtime/browser/xwalk_platform_notification_service.cc",
@@ -354,11 +369,9 @@ source_set("xwalk_runtime") {
   ]
 
   if (is_linux) {
+    configs += [ "//xwalk/build/libnotify" ]
     deps += [ "//xwalk/resources:xwalk_locales" ]
   }
-
-  # TODO(heke123): Add WebNotification for Linux
-  # WebNotification needs libgtk2ui and libnotify.
 
   if (disable_bundled_extensions) {
     defines += [ "DISABLE_BUNDLED_EXTENSIONS" ]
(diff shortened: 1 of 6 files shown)


heke123:
@rakuco   I have two problems in this draft PR:
1) I added the v8_use_external_startup_data back temporally.  please see comments in` [Backport] gn: Remove unnecessary v8 defaults #378` first. 
2) I add  `#!/bin/bash`  in the first line of /xwalk/tools/installer/common/generate-changelog.sh.  Without this line the `subprocess.call` or `subprocess.Popen` cannot execute `generate-changelog.sh` correctly.
I will split it into another PR when we finished discussion.
So it is a draft,  please have a look first,  Thanks:)

BUILD.gn
```diff
@@ -476,6 +497,13 @@ action("generate_upstream_blink_version") {
   ]
 }
 
+if (is_linux) {
+  pkg_config("libnotify_config") {
```
rakuco: I think it is much cleaner to move the detection of libnotify to a separate file, just like Chromium itself does with `src/build/config/linux/{atk,dbus,gconf,...}`. This way, the only changes you need to make here is the `configs +=` part, and all the rest is done in another file with much fewer `if (is_linux)` checks.

tools/installer/BUILD.gn
```diff
@@ -0,0 +1,119 @@
...
+    "//xwalk",
+    "//xwalk:generate_upstream_blink_version",
+  ]
+
+  if (is_component_ffmpeg) {
+    deps += [ "//third_party/ffmpeg" ]
+  }
+}
+
+action("xwalk_installer_linux_prepare") {
```
rakuco: Since the gyp version of `xwalk_installer_linux_prepare` was split into several separate actions, I don't think the name makes sense any longer: in the GN version, it's better to call it something like "generate_changelog" (you already depend on each of these actions in `xwalk_installer_package_deb` anyway).

tools/installer/BUILD.gn
```diff
@@ -0,0 +1,119 @@
...
+  ]
+  deps = [
+    ":common_packaging_files",
+    ":deb_packaging_files",
+    ":xwalk_installer_linux_prepare",
+    "//xwalk",
+    "//xwalk:generate_upstream_blink_version",
+  ]
+
+  if (is_component_ffmpeg) {
```
rakuco: It makes sense to merge this block with the check for `is_component_ffmpeg` in the beginning of the target.

tools/installer/BUILD.gn
```diff
@@ -0,0 +1,119 @@
...
+    build_script_arch = "ia32"
+  } else {
+    build_script_arch = current_cpu
+  }
+
+  if (current_cpu == "x86") {
+    deb_arch = "i386"
+  } else if (current_cpu == "x64") {
+    deb_arch = "amd64"
+  } else if (current_cpu == "arm") {
```
rakuco: I wouldn't worry about the "arm" case for now: it's not in the gyp version and we've never tried to create an ARM Debian package so it's not even clear if it works.

rakuco:
The PR looks quite good; the things I've commented inline are minor.

As for the `v8_use_external_startup_data`, I'm fixing it separately in crosswalk-project/chromium-crosswalk#381 and crosswalk-project/v8-crosswalk#166. In any case, it's unrelated to this PR and a fix for it should be made separately.

tools/installer/BUILD.gn
heke123: it is `deps +=` in below block, so I moved the whole `deps = [...]` block on top after merging `is_component_ffmpeg`

heke123:
I add `#!/bin/bash `in the first line of `/xwalk/tools/installer/common/generate-changelog.sh`, it is so trivial that maybe we don't need to split it out into another PR.

rakuco:
The `bash` bit is part of the required changes in this PR, I don't see a reason to move it to another pull request.

BUILD.gn
heke123: Yes It is much cleaner so I moved the libnotify into separate file.
I checked how chrome use the `libgtkui`, (line 329-335 of chrome/browser/ui/BUILD.gn), so I didn't delete the `if(is_linux)`.

heke123:
Add a "TODO" in `tools/installer/common/installer.include`.

heke123:
@rakuco  PR updated. Thanks:)

rakuco:
lgtm!

PR #3868, opened 2016-08-31
[Android] Crash when parent construct using child reflected method

fujunwei:
The order of constructing object is parent member variables => parent constructor => child member variables => child constructor, so it will be crash when parent constructor using child override reflected method.


fujunwei:
@lincsoon @wuhengzhi PTAL

wuhengzhi:
LGTM.

PR #3869, opened 2016-08-31
Enable GN build for crosswalk-android.

heke123:
In this patch:
    1) Inits the GN build configuration for Crosswalk-Android.
    2) Build out all of the APKs.
    3) Make sure Instrumentation tests pass in GN.
    4) Generate xwalk_core_librarys, javadocs, AARs.
    
    TODO: xwalk_core_unittest. lzma support for download mode.
    
    BUG = XWALK-7285

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -20,6 +20,24 @@ group("xwalk_builder") {
       ":xwalk_all_tests",
     ]
   }
+  if (is_android) {
+    deps += [
+      # For internal testing.
+      "app/android/runtime_client_embedded_shell:xwalk_runtime_client_embedded_shell_apk",
+      "app/android/runtime_client_shell:xwalk_runtime_client_shell_apk",
+      "runtime/android/core_internal_shell:xwalk_core_internal_shell_apk",
+      "runtime/android/core_shell:xwalk_core_shell_apk",
+      "test/android/core/javatests:xwalk_core_test_apk",
+      "test/android/core_internal/javatests:xwalk_core_internal_test_apk",
+      "test/android/runtime_client/javatests:xwalk_runtime_client_test_apk",
+      "test/android/runtime_client_embedded/javatests:xwalk_runtime_client_embedded_test_apk",
+
+      # For external testing.
+      "app/android/app_hello_world:xwalk_app_hello_world_apk",
+      "runtime/android/runtime_lib:xwalk_runtime_lib_apk",
+      "runtime/android/sample:xwalk_core_sample_apk",
+    ]
+  }
 }
 
 if (is_linux) {
@@ -363,14 +381,25 @@ source_set("xwalk_runtime") {
     "//xwalk/resources:xwalk_resources_100_percent_pak",
     "//xwalk/resources:xwalk_resources_200_percent_pak",
     "//xwalk/resources:xwalk_resources_300_percent_pak",
-    "//xwalk/resources:xwalk_strings",
     "//xwalk/sysapps",
     "//xwalk/sysapps:xwalk_sysapps_resources",
   ]
 
+  if (is_android) {
+    deps += [
+      "//components/cdm/browser",
+      "//xwalk/runtime/android/core_internal:xwalk_core_jar_jni",
+      "//xwalk/runtime/android/core_internal:xwalk_core_native_jni",
+    ]
+  } else {
+    deps += [
+      "//xwalk/resources:xwalk_locales",
+      "//xwalk/resources:xwalk_strings",
+    ]
+  }
+
   if (is_linux) {
     configs += [ "//xwalk/build/libnotify" ]
-    deps += [ "//xwalk/resources:xwalk_locales" ]
   }
 
   if (disable_bundled_extensions) {
--- a/app/android/app_hello_world/BUILD.gn
+++ b/app/android/app_hello_world/BUILD.gn
@@ -1,0 +1,35 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+android_apk("xwalk_app_hello_world_apk") {
+  apk_name = "XWalkAppHelloWorld"
+  android_manifest = "AndroidManifest.xml"
+  java_files = [ "src/org/xwalk/app/hello/world/HelloWorldActivity.java" ]
+  deps = [
+    ":xwalk_app_hello_world_apk_assets",
+    ":xwalk_app_hello_world_apk_resources",
+    "//xwalk/app/android/runtime_client:xwalk_app_runtime_java",
+    "//xwalk/runtime/android/core:xwalk_core_java",
+  ]
+}
+
+android_resources("xwalk_app_hello_world_apk_resources") {
+  visibility = [ ":*" ]
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.app.hello.world"
+}
+
+android_assets("xwalk_app_hello_world_apk_assets") {
+  visibility = [ ":*" ]
+  renaming_sources = [
+    "//xwalk/test/android/data/index.html",
+    "//xwalk/test/android/data/sampapp-icon-helloworld.png",
+  ]
+  renaming_destinations = [
+    "www/index.html",
+    "www/sampapp-icon-helloworld.png",
+  ]
+}
--- a/app/android/app_template/BUILD.gn
+++ b/app/android/app_template/BUILD.gn
@@ -1,0 +1,33 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+android_apk("xwalk_app_template_apk") {
+  apk_name = "XWalkAppTemplate"
+  android_manifest = "AndroidManifest.xml"
+  java_files = [ "src/org/xwalk/app/template/AppTemplateActivity.java" ]
+  deps = [
+    ":xwalk_app_template_apk_assets",
+    ":xwalk_app_template_apk_resources",
+    "//xwalk/app/android/runtime_client:xwalk_app_runtime_java",
+    "//xwalk/runtime/android/core:xwalk_core_java",
+  ]
+}
+
+android_assets("xwalk_app_template_apk_assets") {
+  renaming_sources = [
+    "//xwalk/test/android/data/index.html",
+    "//xwalk/test/android/data/sampapp-icon-helloworld.png",
+  ]
+  renaming_destinations = [
+    "www/index.html",
+    "www/sampapp-icon-helloworld.png",
+  ]
+}
+
+android_resources("xwalk_app_template_apk_resources") {
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.app.template"
+}
--- a/app/android/runtime_client/BUILD.gn
+++ b/app/android/runtime_client/BUILD.gn
@@ -1,0 +1,20 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+android_library("xwalk_app_runtime_java") {
+  java_files = [
+    "//xwalk/app/android/runtime_activity/src/org/xwalk/app/XWalkRuntimeActivityBase.java",
+    "src/org/xwalk/app/runtime/extension/XWalkRuntimeExtensionLoader.java",
+    "src/org/xwalk/app/runtime/XWalkCoreProviderImpl.java",
+    "src/org/xwalk/app/runtime/XWalkRuntimeTestHelper.java",
+    "src/org/xwalk/app/runtime/XWalkRuntimeView.java",
+    "src/org/xwalk/app/runtime/XWalkRuntimeViewProvider.java",
+    "src/org/xwalk/app/runtime/XWalkRuntimeViewProviderFactory.java",
+  ]
+  deps = [
+    "//xwalk/runtime/android/core:xwalk_core_java",
+  ]
+}
--- a/build/android.gni
+++ b/build/android.gni
@@ -1,0 +1,25 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file contains feature flags for Crosswalk Android build. Users should
+# import it into their build folder by run:
+#   echo 'import("//xwalk/build/android.gni")' > out/your_dir/args.gn'
+# Add your build args in file args.gn like: target_os = "android"
+# See more in "https://chromium.googlesource.com/chromium/src/+/master/docs/
+# android_build_instructions.md"
+# Then run:
+#   gn gen out/your_dir
+#   ninja -C out/your_dir xwalk_builder
+# to generate Android Apks of Crosswalk.
+
+import("//xwalk/build/common.gni")
+
+# Make release builds smaller by omitting stack unwind support for backtrace()
+# TODO(rakuco): determine if we only want this in official builds.
+# From //build/config/compiler/BUILD.gn
+exclude_unwind_tables = true
+
+# Temporarily disable use of external snapshot files (XWALK-3516).
+# From //build_overrides/v8.gni
+v8_use_external_startup_data = false
--- a/build/version.gni
+++ b/build/version.gni
@@ -46,3 +46,32 @@ _result = exec_script("//build/util/version.py",
                       "scope",
                       [ xwalk_version_file ])
 xwalk_version = _result.full
+
+if (is_android) {
+  import("//build/config/android/config.gni")
+
+  # Get xwalk API version
+  _xwalk_api_version_template = "api = \"@API@\" min_api = \"@MIN_API@\""
+  xwalk_api_version_file = "//xwalk/API_VERSION"
+  _result = exec_script("//build/util/version.py",
+                        [
+                          "-f",
+                          rebase_path(xwalk_api_version_file, root_build_dir),
+                          "-t",
+                          _xwalk_api_version_template,
+                        ],
+                        "scope",
+                        [ xwalk_api_version_file ])
+  api_version = _result.api
+  min_api_version = _result.min_api
+
+  # Get xwalk_version_code
+  _result = exec_script("//xwalk/build/android/generate_version_code.py",
+                        [
+                          "--abi-name=$android_app_abi",
+                          "--version=$xwalk_version",
+                        ],
+                        "trim string",
+                        [ xwalk_version_file ])
+  xwalk_version_code = _result
+}
--- a/build/xwalk.gni
+++ b/build/xwalk.gni
@@ -7,4 +7,13 @@ declare_args() {
   # Whether to build and use Crosswalk's internal extensions (device
   # capabilities, sysapps etc).
   disable_bundled_extensions = false
+
+  # Android: whether the integrity of the Crosswalk runtime library should
+  # be verified when Crosswalk is run in shared mode.
+  verify_xwalk_apk = false
+
+  # Name of Crosswalk Maven artifacts used to generate their respective POM
+  # files.
+  xwalk_core_library_artifact_id = "xwalk_core_library_canary"
+  xwalk_shared_library_artifact_id = "xwalk_shared_library_canary"
 }
--- a/extensions/BUILD.gn
+++ b/extensions/BUILD.gn
@@ -14,10 +14,6 @@ source_set("extensions") {
     "browser/xwalk_extension_process_host.h",
     "browser/xwalk_extension_service.cc",
     "browser/xwalk_extension_service.h",
-    "common/android/xwalk_extension_android.cc",
-    "common/android/xwalk_extension_android.h",
-    "common/android/xwalk_native_extension_loader_android.cc",
-    "common/android/xwalk_native_extension_loader_android.h",
     "common/xwalk_extension.cc",
     "common/xwalk_extension.h",
     "common/xwalk_extension_messages.cc",
@@ -67,6 +63,15 @@ source_set("extensions") {
     "//url",
     "//v8",
   ]
+  if (is_android) {
+    sources += [
+      "common/android/xwalk_extension_android.cc",
+      "common/android/xwalk_extension_android.h",
+      "common/android/xwalk_native_extension_loader_android.cc",
+      "common/android/xwalk_native_extension_loader_android.h",
+    ]
+    deps += [ "//xwalk/extensions/android:xwalk_core_extensions_native_jni" ]
+  }
 }
 
 # TODO(heke123): Remove this config by putting the files in the right place
--- a/extensions/android/BUILD.gn
+++ b/extensions/android/BUILD.gn
@@ -1,0 +1,24 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+android_library("xwalk_core_extensions_java") {
+  deps = [
+    "//base:base_java",
+    "//content/public/android:content_java",
+  ]
+  java_files = [
+    "java/src/org/xwalk/core/internal/extensions/XWalkExtensionAndroid.java",
+    "java/src/org/xwalk/core/internal/extensions/XWalkNativeExtensionLoaderAndroid.java",
+  ]
+}
+
+generate_jni("xwalk_core_extensions_native_jni") {
+  jni_package = "xwalk"
+  sources = [
+    "java/src/org/xwalk/core/internal/extensions/XWalkExtensionAndroid.java",
+    "java/src/org/xwalk/core/internal/extensions/XWalkNativeExtensionLoaderAndroid.java",
+  ]
+}
--- a/extensions/test/BUILD.gn
+++ b/extensions/test/BUILD.gn
@@ -101,11 +101,23 @@ action("generate_jsapi_extensions_test") {
 }
 
 loadable_module("echo_extension") {
-  visibility = [ ":*" ]
+  visibility = [
+    ":*",
+
+    # These test Android targets also package this library.
+    "//xwalk/app/android/runtime_client_shell:*",
+    "//xwalk/app/android/runtime_client_embedded_shell:*",
+  ]
   sources = [
     "echo_extension.c",
   ]
-  output_dir = "$root_out_dir/tests/extension/echo_extension"
+
+  # This library is used as in |native_libs| in the Android targets mentioned
+  # above, and as of M52 these libraries are required to be in $root_out_dir.
+  # TODO(heke123): Check if we can get rid of this once we move to M53.
+  if (!is_android) {
+    output_dir = "$root_out_dir/tests/extension/echo_extension"
+  }
 }
 
 loadable_module("echo_extension_messaging_2") {
--- a/runtime/android/core/BUILD.gn
+++ b/runtime/android/core/BUILD.gn
@@ -1,0 +1,64 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+android_library("xwalk_core_java") {
+  deps = [
+    ":xwalk_core_java_resources",
+    "//xwalk/runtime/android/core_internal:xwalk_core_reflection_layer_java_gen",
+    "//xwalk/third_party/lzma_sdk:lzma_sdk_java",
+  ]
+  srcjars =
+      [ "$root_gen_dir/xwalk_core_reflection_layer/wrapper/wrapper.srcjar" ]
+  java_files = [
+    "src/org/xwalk/core/extension/BindingObject.java",
+    "src/org/xwalk/core/extension/BindingObjectAutoJS.java",
+    "src/org/xwalk/core/extension/BindingObjectStore.java",
+    "src/org/xwalk/core/extension/EventTarget.java",
+    "src/org/xwalk/core/extension/ExtensionInstanceHelper.java",
+    "src/org/xwalk/core/extension/JsApi.java",
+    "src/org/xwalk/core/extension/JsConstructor.java",
+    "src/org/xwalk/core/extension/JsContextInfo.java",
+    "src/org/xwalk/core/extension/JsStubGenerator.java",
+    "src/org/xwalk/core/extension/MessageHandler.java",
+    "src/org/xwalk/core/extension/MessageInfo.java",
+    "src/org/xwalk/core/extension/ReflectionHelper.java",
+    "src/org/xwalk/core/extension/XWalkCoreExtensionBridge.java",
+    "src/org/xwalk/core/extension/XWalkExtensionContextClient.java",
+    "src/org/xwalk/core/extension/XWalkExternalExtension.java",
+    "src/org/xwalk/core/extension/XWalkExternalExtensionBridge.java",
+    "src/org/xwalk/core/extension/XWalkExternalExtensionBridgeFactory.java",
+    "src/org/xwalk/core/extension/XWalkExternalExtensionManagerImpl.java",
+    "src/org/xwalk/core/JavascriptInterface.java",
+    "src/org/xwalk/core/XWalkActivity.java",
+    "src/org/xwalk/core/XWalkActivityDelegate.java",
+    "src/org/xwalk/core/XWalkApplication.java",
+    "src/org/xwalk/core/XWalkEnvironment.java",
+    "src/org/xwalk/core/XWalkFileChooser.java",
+    "src/org/xwalk/core/XWalkCoreWrapper.java",
+    "src/org/xwalk/core/XWalkDialogManager.java",
+    "src/org/xwalk/core/XWalkInitializer.java",
+    "src/org/xwalk/core/XWalkDecompressor.java",
+    "src/org/xwalk/core/XWalkLibraryInterface.java",
+    "src/org/xwalk/core/XWalkLibraryLoader.java",
+    "src/org/xwalk/core/XWalkMixedResources.java",
+    "src/org/xwalk/core/XWalkUpdater.java",
+  ]
+}
+
+android_resources("xwalk_core_java_resources") {
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.core"
+  deps = [
+    ":xwalk_app_strings",
+  ]
+}
+
+java_strings_grd("xwalk_app_strings") {
+  grd_file = "strings/xwalk_app_strings.grd"
+  outputs = [
+    "values/xwalk_app_strings.xml",
+  ]
+}
--- a/runtime/android/core_internal/BUILD.gn
+++ b/runtime/android/core_internal/BUILD.gn
@@ -1,0 +1,175 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+import("//xwalk/build/version.gni")
+import("//xwalk/build/xwalk.gni")
+
+reflection_gen_dir = "$root_gen_dir/xwalk_core_reflection_layer/"
+android_library("xwalk_core_internal_java") {
+  deps = [
+    ":xwalk_core_internal_java_resources",
+    ":xwalk_core_reflection_layer_java_gen",
+    "//base:base_java",
+    "//components/navigation_interception/android:navigation_interception_java",
+    "//components/web_contents_delegate_android:web_contents_delegate_android_java",
+    "//content/public/android:content_java",
+    "//media/base/android:media_java",
+    "//net/android:net_java",
+    "//ui/android:ui_java",
+    "//xwalk/extensions/android:xwalk_core_extensions_java",
+  ]
+  srcjars = [ "$reflection_gen_dir/bridge/bridge.srcjar" ]
+  java_files = [
+    "src/org/xwalk/core/internal/AndroidProtocolHandler.java",
+    "src/org/xwalk/core/internal/ClientCertLookupTable.java",
+    "src/org/xwalk/core/internal/ClientCertRequestHandlerInternal.java",
+    "src/org/xwalk/core/internal/ClientCertRequestInternal.java",
+    "src/org/xwalk/core/internal/CustomViewCallbackHandlerInternal.java",
+    "src/org/xwalk/core/internal/CustomViewCallbackInternal.java",
+    "src/org/xwalk/core/internal/ErrorCodeConversionHelper.java",
+    "src/org/xwalk/core/internal/InMemorySharedPreferences.java",
+    "src/org/xwalk/core/internal/MixedContext.java",
+    "src/org/xwalk/core/internal/PageLoadListener.java",
+    "src/org/xwalk/core/internal/ReflectConstructor.java",
+    "src/org/xwalk/core/internal/ReflectField.java",
+    "src/org/xwalk/core/internal/ReflectMethod.java",
+    "src/org/xwalk/core/internal/SslUtil.java",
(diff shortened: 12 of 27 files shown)


heke123:
@rakuco   I submit the "Init patch".  Please help to have a look, Thanks!
@PeterWangIntel, @halton  FYI.

BUILD.gn
```diff
@@ -6,6 +6,9 @@ import("//build/config/features.gni")
 import("//build/config/ui.gni")
 import("//xwalk/build/version.gni")
 import("//xwalk/build/xwalk.gni")
+if (is_android) {
+  import("//build/config/android/config.gni")
```
rakuco: This doesn't seem to be required in this file.

BUILD.gn
```diff
@@ -360,17 +371,29 @@ source_set("xwalk_runtime") {
     "//xwalk/extensions",
     "//xwalk/resources:generate_xwalk_resources",
     "//xwalk/resources:xwalk_pak",
+    "//xwalk/resources:xwalk_resources",
```
rakuco: This is not really Android-specific and was fixed independently in #3872

BUILD.gn
```diff
@@ -360,17 +371,29 @@ source_set("xwalk_runtime") {
...
     "//xwalk/resources:xwalk_resources_100_percent_pak",
     "//xwalk/resources:xwalk_resources_200_percent_pak",
     "//xwalk/resources:xwalk_resources_300_percent_pak",
-    "//xwalk/resources:xwalk_strings",
     "//xwalk/sysapps",
     "//xwalk/sysapps:xwalk_sysapps_resources",
   ]
-
```
rakuco: It makes sense to keep this empty line.

build/android/android.gni
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
```
rakuco: I don't see why this file is not being added to the same directory as `linux.gni`.

build/version.gni
```diff
@@ -1,27 +1,11 @@
-# Copyright 2015 The Chromium Authors. All rights reserved.
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
```
rakuco: You're supposed to add copyright notices, not replace existing ones -- even if you change it, the code came from Chromium.

build/version.gni
```diff
@@ -1,27 +1,11 @@
...
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-# This exposes the Chrome version and Crosswalk version as GN variables for
```
rakuco: I think it's valid to keep the first half of this file as-is, and only extend it at the end with the Android bits.

build/xwalk.gni
```diff
@@ -7,4 +7,13 @@ declare_args() {
   # Whether to build and use Crosswalk's internal extensions (device
   # capabilities, sysapps etc).
   disable_bundled_extensions = false
+
+  # Android: whether the integrity of the Crosswalk runtime library should
+  # be verified when Crosswalk is run in shared mode.
+  verify_xwalk_apk = 0
```
rakuco: `false`, not `0`.

extensions/BUILD.gn
```diff
@@ -58,6 +58,9 @@ source_set("extensions") {
     "renderer/xwalk_v8tools_module.cc",
     "renderer/xwalk_v8tools_module.h",
   ]
+  if (is_android) {
+    include_dirs = [ "$root_gen_dir/xwalk" ]
```
rakuco: This is a red flag. If you see yourself needing to use a hack like this there's a 99% chance you're doing something wrong, such as missing a dependency.

It's exactly what's happening here: you need this directory because you are not depending on `//xwalk/extensions/android:xwalk_core_extensions_native_jni` here even though the android files here depend on the headers generated there (the `generate_jni` template takes care of adding the right directories to the list of include dirs). In fact, it's possible that dependency is not even needed by `xwalk_runtime` itself. And it's also possible that other targets have the same problem.

The cleanest solution here is this:

```
if (is_android) {
  sources += [
    "common/android/xwalk_extension_android.cc",
    "common/android/xwalk_extension_android.h",
    "common/android/xwalk_native_extension_loader_android.cc",
    "common/android/xwalk_native_extension_loader_android.h",
  ]
  deps += [ "//xwalk/extensions/android:xwalk_core_extensions_native_jni" ]
}
```

Adding the Android files inside the `if` also brings the whole file closer to the spirit of GN's style guide, and is something already done in places like `ui/events/BUILD.gn`.

extensions/android/BUILD.gn
```diff
@@ -0,0 +1,22 @@
...
+    "//content/public/android:content_java",
+  ]
+  java_files = [
+    "java/src/org/xwalk/core/internal/extensions/XWalkExtensionAndroid.java",
+    "java/src/org/xwalk/core/internal/extensions/XWalkNativeExtensionLoaderAndroid.java",
+  ]
+}
+
+generate_jni("xwalk_core_extensions_native_jni") {
+  jni_package = rebase_path(
```
rakuco: Why complicate this assignment with this `rebase_path` call. Using something as simple as "xwalk" or "extensions" works just fine (again, look at how `generate_jni` is implemented).

runtime/android/core_internal/BUILD.gn
```diff
@@ -0,0 +1,210 @@
...
+    "//xwalk/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/presentation/PresentationViewNull.java",
+    "//xwalk/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/presentation/XWalkPresentationContent.java",
+    "//xwalk/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/wifidirect/WifiDirect.java",
+    "//xwalk/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/DisplayManagerJBMR1.java",
+    "//xwalk/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/DisplayManagerNull.java",
+    "//xwalk/runtime/android/core_internal/src/org/xwalk/core/internal/extension/api/XWalkDisplayManager.java",
+  ]
+}
+
+copy("bridge") {
```
rakuco: This likely needs `visibility = [ ":*" ]`.

rakuco:
I've added some inline comments, but have not read the entire patch in detail yet.

Overall, I've noticed none of the new targets have `visibility = [ ":*" ]` when they're internal, a lot of `rebase_path` calls seem to be entirely unnecessary, your dependencies (at least some of them) are in the wrong targets and you haven't explained why the change to `reflection_generator.py` is needed.

BUILD.gn
heke123: Yes, I saw your PR has already correct this,  Thanks.

build/xwalk.gni
heke123: It is on purpose set as 0,  see line 124 of `//xwalk/tools/reflection_generator/reflection_generator.py`

extensions/android/BUILD.gn
heke123: In GYP, all the jni headers are generated into `$root_out_dir/jni/`
If I set "jni_package=xwalk" here,  the jni headers will be generated into `$target_out_dir/xwalk/jni/`, so I use a rebase_path to make sure all the jni headers are generated same path as in GYP.

heke123:
I replied some comments,  I'll make correction in my following entire patch.
the explaination of changing reflection_generator.py:
      I found those java files generated by reflection_generator cannot be assigned directly into target Android_Library.  I have to zip them as srcjar first, then add the srcjar to Andorid_Library.
See my post:  https://groups.google.com/a/chromium.org/forum/#!topic/gn-dev/GokFp6TorgI

extensions/android/BUILD.gn
rakuco: There's no need to generate the headers in the same location as gyp, this is all internal to the build. As long as the right `-I<path>` is being passed (which `generate_jni` takes care of), there's no problem if the path is different.

build/xwalk.gni
rakuco: This is another case of what I've been saying about following "the spirit of law" (in this case, trying to do things the GN way instead of just copying gyp). In other words, fix the script instead of not using GN the right way :-)

The script accepts ints because gyp uses 1 and 0 instead of true or false, but it should be easy to change the behavior. I can write the patch myself.

rakuco: #3873

rakuco:
> I found those java files generated by reflection_generator cannot be assigned directly into target Android_Library. I have to zip them as srcjar first, then add the srcjar to Andorid_Library.
> See my post: https://groups.google.com/a/chromium.org/forum/#!topic/gn-dev/GokFp6TorgI

Got it. It makes sense to land this separately with a proper explanation. I'll see if I can work on this tomorrow.

rakuco:
> I'll see if I can work on this tomorrow.

#3875

extensions/test/BUILD.gn
```diff
@@ -106,6 +106,16 @@ loadable_module("echo_extension") {
     "echo_extension.c",
   ]
   output_dir = "$root_out_dir/tests/extension/echo_extension"
+
+  if (is_android) {
```
rakuco: This is overly complicated. GN doesn't seem to complain about entries in `visibility` that do not exist, so you can do something like this instead:

```
loadable_module("echo_extension") {
  visibility = [
    ":*",

    # These test Android targets also package this library.
    "//xwalk/app/android/runtime_client_embedded_shell:*",
    "//xwalk/app/android/runtime_client_shell:*",
  ]
  sources = [
    "echo_extension.c",
  ]

  # This library is used as in |native_libs| in the Android targets mentioned
  # above, and as of M52 these libraries are required to be in $root_out_dir.
  # TODO(heke123): Check if we can get rid of this once we move to M53.
  if (!is_android) {
    output_dir = "$root_out_dir/tests/extension/echo_extension"
  }
}
```

app/android/app_hello_world/BUILD.gn
```diff
@@ -0,0 +1,39 @@
...
+  sources = []
+  renaming_sources = [
+    "//xwalk/test/android/data/index.html",
+    "//xwalk/test/android/data/sampapp-icon-helloworld.png",
+  ]
+  renaming_destinations = [
+    "www/index.html",
+    "www/sampapp-icon-helloworld.png",
+  ]
+  disable_compression = true
```
rakuco: Are you sure this is needed? The only occurrences I could find in Chromium were for assets containing .pak or .dex files, which is not the case here.

app/android/app_hello_world/BUILD.gn
```diff
@@ -0,0 +1,39 @@
...
+
+android_resources("xwalk_app_hello_world_apk_resources") {
+  visibility = [ ":*" ]
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.app.hello.world"
+}
+
+android_assets("xwalk_app_hello_world_apk_assets") {
+  visibility = [ ":*" ]
+  sources = []
```
rakuco: You don't need to set this if it's an empty list.

app/android/app_template/BUILD.gn
```diff
@@ -0,0 +1,67 @@
...
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.app.template"
+}
+
+action("prepare_xwalk_app_template") {
+  script = "//xwalk/build/android/gn/generate_app_packaging_tool.py"
+  outputs = [
+    "$root_out_dir/xwalk_app_template",
+  ]
+  src_dir = rebase_path("//", "$root_out_dir")
```
rakuco: This block is needlessly confusing. This is easier to read:

```
  _build_dir = rebase_path(root_build_dir)
  _src_dir = rebase_path("//xwalk")
  args = [
    "--build-dir", _build_dir,
    "--build-mode", android_configuration_name,
    "--output-dir", "${_build_dir}/xwalk_app_template",
    "--source-dir", _src_dir,
  ]
```

app/android/app_template/BUILD.gn
```diff
@@ -0,0 +1,67 @@
...
+
+android_resources("xwalk_app_template_apk_resources") {
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.app.template"
+}
+
+action("prepare_xwalk_app_template") {
+  script = "//xwalk/build/android/gn/generate_app_packaging_tool.py"
+  outputs = [
+    "$root_out_dir/xwalk_app_template",
```
rakuco: Is this enough to make this action always run?

app/android/app_template/BUILD.gn
```diff
@@ -0,0 +1,67 @@
...
+  disable_compression = true
+}
+
+android_resources("xwalk_app_template_apk_resources") {
+  resource_dirs = [ "res" ]
+  custom_package = "org.xwalk.app.template"
+}
+
+action("prepare_xwalk_app_template") {
+  script = "//xwalk/build/android/gn/generate_app_packaging_tool.py"
```
rakuco: You forgot to declare `common_function.py` a part of `inputs`.

app/android/app_template/BUILD.gn
```diff
@@ -0,0 +1,67 @@
...
+  ]
+}
+
+action("xwalk_app_template") {
+  script = "//xwalk/tools/tar.py"
+  outputs = [
+    "$root_out_dir/xwalk_app_template.tar.gz",
+  ]
+  inputs = [
+    "AndroidManifest.xml",
```
rakuco: From `gn help action`:

> It is recommended you put inputs to your script in the "sources"
> variable, and stuff like other Python files required to run your
> script in the "inputs" variable.

app/android/app_template/BUILD.gn
```diff
@@ -0,0 +1,67 @@
...
+  ]
+  args = [ "./xwalk_app_template" ]
```
rakuco: Relative paths like these are ugly and dangerous: try `args = [ rebase_path(root_build_dir) + "/xwalk_app_template" ]`.

app/android/runtime_client/BUILD.gn
```diff
@@ -0,0 +1,57 @@
...
+    "src/org/xwalk/app/runtime/XWalkRuntimeViewProvider.java",
+    "src/org/xwalk/app/runtime/XWalkRuntimeViewProviderFactory.java",
+  ]
+  deps = [
+    "//xwalk/runtime/android/core:xwalk_core_java",
+  ]
+  srcjar_deps = [ ":xwalk_runtime_client_version" ]
+}
+
+copy("xwalk_runtime_client_version") {
```
rakuco: These two actions are also too complicated for no reason and another instance of "look at the bigger picture and don't just copy gyp".

First of all, `generate_runtime_client_version.py` is an ugly, useless script that has no reason to exist. I'm removing it in #3878.

Secondly, GN has a much better infrastructure to help you not reinvent the wheel: `chrome/version.gni` and `build/config/zip.gni`. Look at `components/cronet/android/BUILD.gn` to see what I'm talking about. The whole file could be rewritten like this:

```
import("//build/config/zip.gni")
import("//chrome/version.gni")

android_library("xwalk_app_runtime_java") {
# ...
}

_version_dir =
    "$target_gen_dir/templates/xwalk_runtime_client_version_java"
_version_file =
    "$_version_dir/org/xwalk/app/runtime/XWalkRuntimeClientVersion.java"

process_version("runtime_version_java") {
  visibility = [ ":*" ]
  process_only = true
  # sources = [ "//xwalk/VERSION" ]
  extra_args = [ "-f", rebase_path("//xwalk/VERSION") ]
  template_file = "src/templates/XWalkRuntimeClientVersion.template"
  output = _version_file
}

zip("runtime_version_srcjar") {
  visibility = [ ":*" ]
  inputs = [ _version_file ]
  output = "$target_gen_dir/$target_name.srcjar"
  base_dir = _version_dir
  deps = [ ":runtime_version_java" ]
}

```

Note that there's a commented out line in `runtime_version_java`. That's because `version.gni` passes some files such as `chrome/VERSION` to version.py after the files listed in `sources`, so `chrome/VERSION` will override the values in `xwalk/VERSION`. I'm trying to change the order in https://codereview.chromium.org/2308583002, but until then you can use the version that is not commented out that achieves the same final result.

build/android.gni
```diff
@@ -0,0 +1,23 @@
...
+# import it into their build folder by run:
+#   echo 'import("//xwalk/build/android.gni")' > out/your_dir/args.gn'
+# Then run:
+#   gn gen out/your_dir
+#   ninja -C out/your_dir xwalk_builder
+# to generate Android Apks of Crosswalk.
+
+import("//xwalk/build/common.gni")
+
+target_os = "android"
```
rakuco: I'd rather not do this here and have users explicitly do this in `args.gn`. It makes it easier to see what's going on with `gn args`, is more in line with existing Chromium GN instructions for Android and also allows us to get rid of `android.gni` in the future more easily.

build/android.gni
```diff
@@ -0,0 +1,23 @@
...
+
+# Make release builds smaller by omitting stack unwind support for backtrace()
+# TODO(rakuco): determine if we only want this in official builds.
+# From //build/config/compiler/BUILD.gn
+exclude_unwind_tables = true
+
+# Temporarily disable use of external snapshot files (XWALK-3516).
```
rakuco: Which .gn file is this from?

test/android/BUILD.gn
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+# TODO(heke123): implement xwalk_core_unittests.
+
+shared_library("xwalk_core_unittests") {
```
rakuco: Nothing seems to depend on this target, so I can't build it.

test/android/BUILD.gn
```diff
@@ -0,0 +1,21 @@
...
+  deps = [
+    "//base/base:test_support_base",
```
rakuco:
```
ERROR at //xwalk/test/android/BUILD.gn:11:5: Can't load input file.
    "//base/base:test_support_base",
    ^------------------------------
Unable to load:
  /data/src/crosswalk/crosswalk/src/base/base/BUILD.gn
```

test/android/BUILD.gn
```diff
@@ -0,0 +1,21 @@
...
+    "//net/net:net_test_support",
```
rakuco:
```
/data/src/crosswalk/crosswalk/src/buildtools/linux64/gn --root=/data/src/crosswalk/crosswalk/src -q gen //gn-android/Release/
ERROR at //xwalk/test/android/BUILD.gn:12:5: Can't load input file.
    "//net/net:net_test_support",
    ^---------------------------
Unable to load:
  /data/src/crosswalk/crosswalk/src/net/net/BUILD.gn
```

test/android/BUILD.gn
rakuco: And you're missing `testonly = true`.

test/android/BUILD.gn
```diff
@@ -0,0 +1,21 @@
...
+    "//testing/android/native_test:native_test_native_code",
+    "//testing/gmock:gmock",
+    "//testing/gtest:gtest",
+  ]
+  include_dirs = [ "//" ]
```
rakuco: Not needed.

build/version.gni
```diff
@@ -2,8 +2,7 @@
...
-# use in build files.
+# This exposes the Chrome version as GN variables for use in build files.
```
rakuco: I commented a while ago about how changing the existing parts of this file was not necessary, and I still do.

build/android/BUILD.gn
```diff
@@ -0,0 +1,340 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//xwalk/build/xwalk.gni")
+import("//xwalk/build/version.gni")
+
+declare_args() {
```
rakuco: ?!

build/android/BUILD.gn
```diff
@@ -0,0 +1,340 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
```
rakuco: I don't see why this file was added here. `build/android` contains helper scripts used during the build, there's nothing to be built _from_ here.

rakuco:
I've added more inline comments in the files I've managed to review (I haven't reached the big ones in `runtime` yet).

General comments that apply to many files:
* All .gn and .gni files must have a copyright header.
* Assets without sources don't need `sources = []`.
* Assets without .pak or .dex files likely do not need `disable_compression = true`.
* Single-item lists should be put in a single line:
````
# correct
foo = [ "bar" ]

# wrong
foo = [
  "bar"
]
````

We've spoken about the Python scripts in build/android/gn; they obviously must be removed, but I still need to see what they do differently to figure out what needs to be fixed in the existing ones in build/android.

File: test/android/BUILD.gn (hunk 22 above)
Comment (heke123) [inline test/android/BUILD.gn]: Sorry, this part is not finished yet.  In xwalk_android_test.gypi:
```
'target_name': 'xwalk_core_unittests_apk',
       'type': 'none',
       'dependencies': [
         'xwalk_core_unittests',
         'xwalk_core_unittests_java',
         'xwalk_core_unittests_jni',
       ],
```
I'll add this part as soon as possible, please ignore this temporarily.

File: build/version.gni (hunk 6 above)
Comment (heke123) [inline build/version.gni]: Sorry, I'm still not clear with this.  My understanding: I keep the existing copyright notices, and add comments for android part. like below:
```
# Copyright 2015 The Chromium Authors. All rights reserved.
# Use of this source code is governed by a BSD-style license that can be
# found in the LICENSE file.

# This exposes the Chrome and Crosswalk version as GN variables for use in
# build files. This also calculates xwalk_version_code for Crosswalk-Android.
```
Am I correct?

File: build/android/BUILD.gn (hunk 27 above)
Comment (heke123) [inline build/android/BUILD.gn]: And there are lots of Relative paths you mentioned too.  I'll modify this.

File: build/android/BUILD.gn (hunk 28 above)
Comment (heke123) [inline build/android/BUILD.gn]: I think the operation of generating xwalk_core_library/aar operation has nothing to do with the source code and looks like a build-target wrapper of python scripts, that's why I put them here.
Maybe need to find a better place, //xwalk/BUILD.gn?  or //xwalk/runtime/android/BUILD.gn?

**Comment by heke123** [comment]
> Assets without sources don't need sources = []. 

I'm not sure if below code is ok:  (without sources = [], the following sources+= complains error) 
```
  sources = []
  deps = []
  if (icu_use_data_file) {
    sources += [ "$root_out_dir/icudtl.dat" ]
    deps += [ "//third_party/icu:icudata" ]
  }
  if (v8_use_external_startup_data) {
    sources += [
      "$root_out_dir/natives_blob.bin",
      "$root_out_dir/snapshot_blob.bin",
    ]   
  }
```

> Single-item lists should be put in a single line:

No problem, I'll have a check.  
By the way, I find the `inputs` and `outputs` of action will be formatted by "gn format" cmd as
```
inputs = [
  "bar"
]
```
even if it is single line.  so we can ignore this exception situation. 

I'll have a full check on the code to follow the general comments you mentioned. Thanks very much.

**Comment by rakuco** [comment]
> I'm not sure if below code is ok: (without sources = [], the following sources+= complains error)

Yeah, if you're manipulating it later you need a `sources = []`; I was referring to the cases where you just declared it there without using it later.

> By the way, I find the inputs and outputs of action will be formatted by "gn format" cmd as

I'm not entirely sure about this anymore either: `gn format` also broke some single-line statements I had into multiple lines, so I'd go with whatever `gn format` produces.

File: build/android/BUILD.gn (hunk 28 above)
Comment (rakuco) [inline build/android/BUILD.gn]: Well, they do have to do with the source code: they all depend on other targets and files and generate something out of them afterwards. I'd put them wherever the original targets that generate libxwalkcore & co are.

File: runtime/android/runtime_lib/BUILD.gn [hunk 29]
```diff
@@ -0,0 +1,63 @@
...
+import("//xwalk/build/version.gni")
+import("//third_party/icu/config.gni")
+
+android_apk("xwalk_runtime_lib_apk") {
+  apk_name = "XWalkRuntimeLib"
+  android_manifest = "AndroidManifest.xml"
+  java_files = []
+  native_libs = [ "$root_out_dir/libxwalkcore.so" ]
+  version_name = xwalk_version
+  version_code = "$xwalk_version_code"
```
rakuco: you can just use `xwalk_version_code` here without the quotes and the `$`.

runtime/android/core_shell/BUILD.gn
```diff
@@ -0,0 +1,100 @@
...
+    "$root_out_dir/libxwalkcore.so",
+    "$root_out_dir/libxwalkdummy.so",
+  ]
+  deps = [
+    ":xwalk_core_shell_apk_assets",
+    ":xwalk_core_shell_apk_java",
+    ":xwalk_core_shell_apk_pak",
+    ":xwalk_core_shell_apk_resources",
+    "//base:base_java",
+    "//third_party/android_tools:android_support_v13_java",
```
rakuco: I don't think this one is necessary here, it's `xwalk_core_shell_apk_java` that uses it.

runtime/android/core_internal_shell/BUILD.gn
```diff
@@ -0,0 +1,76 @@
...
+  apk_name = "XWalkCoreInternalShell"
+  android_manifest = "AndroidManifest.xml"
+  native_libs = [ "$root_out_dir/libxwalkcore.so" ]
+  deps = [
+    ":xwalk_core_internal_shell_apk_assets",
+    ":xwalk_core_internal_shell_apk_java",
+    ":xwalk_core_internal_shell_apk_pak",
+    ":xwalk_core_internal_shell_apk_resources",
+    "//base:base_java",
+    "//third_party/android_tools:android_support_v13_java",
```
rakuco: Nothing in this directory seems to depend on this.

runtime/android/core_internal_shell/BUILD.gn
```diff
@@ -0,0 +1,76 @@
...
+android_library("xwalk_core_internal_shell_apk_java") {
+  java_files = [
+    "src/org/xwalk/core/internal/xwview/shell/XWalkViewInternalShellActivity.java",
+    "src/org/xwalk/core/internal/xwview/test/TestContentProvider.java",
+    "src/org/xwalk/core/internal/xwview/test/XWalkViewInternalTestRunnerActivity.java",
+  ]
+  deps = [
+    ":xwalk_core_internal_shell_apk_resources",
+    "//base:base_java",
+    "//third_party/android_tools:android_support_v13_java",
```
rakuco: Ditto.

runtime/android/core_internal_empty/BUILD.gn
```diff
@@ -0,0 +1,34 @@
...
+# "xwalk_core_empty_embedder_apk" is used for generating the R.txt.
+android_apk("xwalk_core_empty_embedder_apk") {
+  apk_name = "XWalkCoreEmptyEmbedder"
+  android_manifest = "AndroidManifest.xml"
+  deps = [
+    "//components/web_contents_delegate_android:web_contents_delegate_android_java_resources",
+    "//content/public/android:content_java_resources",
+    "//ui/android:ui_java_resources",
+    "//xwalk/runtime/android/core:xwalk_core_java_resources",
+    "//xwalk/runtime/android/core_internal:xwalk_core_internal_java_resources",
```
rakuco: I don't think it's needed (which also invalidates the comment above the target), see #3880.

runtime/android/core_internal/BUILD.gn
```diff
@@ -0,0 +1,215 @@
...
+    "//base:base_java",
+    "//components/navigation_interception/android:navigation_interception_java",
+    "//components/web_contents_delegate_android:web_contents_delegate_android_java",
+    "//content/public/android:content_java",
+    "//media/base/android:media_java",
+    "//net/android:net_java",
+    "//ui/android:ui_java",
+    "//xwalk/extensions/android:xwalk_core_extensions_java",
+  ]
+  srcjar_deps = [ ":bridge" ]
```
rakuco: Once crosswalk-project/chromium-crosswalk#384 lands, you should be able to get rid of the `bridge` target altogether and just add a dependency on `xwalk_core_reflection_layer_java_gen` and add the path to `bridge.srcjar` as `srcjars` instead of `srcjar_deps`.

runtime/android/core/BUILD.gn
```diff
@@ -0,0 +1,76 @@
...
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+android_library("xwalk_core_java") {
+  deps = [
+    ":xwalk_core_java_resources",
+    "//xwalk/third_party/lzma_sdk:lzma_sdk_java",
+  ]
+  srcjar_deps = [ ":wrapper" ]
```
rakuco: The same thing I mentioned about getting rid of `bridge` in `runtime/android/core_internal/BUILD.gn` also applies here.

build/android/BUILD.gn
```diff
@@ -0,0 +1,340 @@
...
+  args = [ "./xwalk_shared_library" ]
+  outputs = [
+    "$root_out_dir/xwalk_shared_library.tar.gz",
+  ]
+  deps = [
+    ":xwalk_shared_library",
+  ]
+}
+
+action("xwalk_core_library_documentation") {
```
rakuco: I suggest you take a look at `components/cronet/android:generate_javadoc`: it first packages all the sources into a single source jar and passes that to its [javadoc wrapper](https://github.com/crosswalk-project/chromium-crosswalk/blob/master/components/cronet/tools/generate_javadoc.py). You could take some inspiration from it, and it's fair to add a `build/android/generate_javadoc.py` adapted from the upstream one if you need to.

build/android/BUILD.gn
```diff
@@ -0,0 +1,340 @@
...
+  ]
+  deps = [
+    ":copy_xwalk_core_library_src",
+    ":xwalk_core_library",
+  ]
+}
+
+action("pack_xwalk_shared_library") {
+  script = "//xwalk/tools/tar.py"
+  args = [ "./xwalk_shared_library" ]
```
rakuco: Paths starting with "./" should be avoided; see one of the comments I added last week about getting absolute paths via `rebase_path(root_build_dir)`.

build/android/BUILD.gn
```diff
@@ -0,0 +1,340 @@
...
+    "//xwalk/runtime/android/core_internal_empty:xwalk_core_empty_embedder_apk",
+  ]
+}
+
+# Remove "build/android/maven_pom.gypi" and call "build/util/version.py" directly
+action("xwalk_core_library_pom_gen") {
+  script = "//build/util/version.py"
+  args = [
+    "-i",
+    "$src_dir/xwalk/runtime/android/maven/xwalk_core_library.pom.xml.in",
```
rakuco: You're not declaring this as an input to the action, so if it changes the action will not be run. In fact, you should try using the `process_version` template in chrome/version.gni instead.

build/android/BUILD.gn
```diff
@@ -0,0 +1,340 @@
...
+    "-e",
+    "ARTIFACT_VERSION='$xwalk_version'",
+  ]
+  outputs = [
+    "$root_out_dir/xwalk_shared_library.pom.xml",
+  ]
+}
+
+# TODO(heke123): do we really need the core_internal_empty to get R.class files?
+# Is it possible to get R.class files from core_shell or core_internal_shell?
```
rakuco: I think the easiest way to get answers to your questions is by testing it.

runtime/android/runtime_lib/BUILD.gn
heke123: It is set on purpose to convert integer to string.
In //xwalk/build/version.gni, Line60,  I tried set "string" as the input_conversion of exec_script, but found it returns "8066002/". so I set "value" as the input_conversion and get integer value 8066002.  So I convert it into string here. 
Maybe we need to modify //xwalk/build/android/generate_version_code.py if want to make it totally clean.

rakuco: Using "trim string" in the `exec_script` call seems to work fine here.

build/android/BUILD.gn
heke123: > it first packages all the sources into a single source jar and passes that to its javadoc wrapper.
Seems it doesn't package sources into srcjar before passes to javadoc wrapper?  The sources are set by `--lib-java-dir`. So I have to extract it first(like "extract_cronet_jars" action does)
Or is it possible to generate both java files and srcjars in script "reflection_generator.py" with flag "--use-srcjars"? 
Which way do you prefer?

runtime/android/runtime_lib/BUILD.gn
heke123: exec_script works, thanks.

build/android/BUILD.gn
heke123: sorry, it is a reminder to myself,  I'll delete this comments, and have a whole check on the usage of core_internal_empty_embedder and core_empty_embedder.

build/android/BUILD.gn
rakuco: My bad, I was looking at the other targets that use `jar_src` template (which might be useful for other targets here). What cronet seems to do is use a combination of a javadoc Python wrapper (which you will need anyway, and where you can move most of the arguments) and a custom Ant build.xml to actually call javadoc. Please take a look at those and determine if something similar to that is useful.

app/android/app_template/BUILD.gn
heke123: another unsure point here:  I add the "//xwalk/build/android/gn/generate_app_packaging_tool.py", which is slightly different with original "//xwalk/build/android/generate_app_packaging_tool.py".
I remember last time you mentioned that the new created scripts in gn/ folder is not acceptable, any suggestions?

app/android/app_template/BUILD.gn
heke123: I'll remove xwalk_app_template target in this patch because it depends on xwalk_core_library which is still in discussing.

runtime/android/core_internal/BUILD.gn
heke123: I found `android_library` doesn't allow user to insert "action" and "copy" into its deps. The only way I found is to use the `srcjars_deps `.  If I change the code as below:
```
android_library("xwalk_core_internal_java") {
  deps = [ 
    ":xwalk_core_reflection_layer_java_gen",  #outputs "$reflection_gen_dir/bridge/bridge.srcjar"
    ......
  ]
  srcjars = [ "$reflection_gen_dir/bridge/bridge.srcjar" ]
  ......
}

```
"gn gen" complains:
```
ERROR Input to target not generated by a dependency.
The file:
  //out/an/gen/xwalk_core_reflection_layer/bridge/bridge.srcjar
is listed as an input or source for the target:
  //xwalk/runtime/android/core_internal:xwalk_core_internal_java__compile_java__javac
but this file was not generated by any dependencies of the target. The target
that generates the file is:
  //xwalk/runtime/android/core_internal:xwalk_core_reflection_layer_java_gen

If you have generated inputs, there needs to be a dependency path between the
two targets in addition to just listing the files. For indirect dependencies,
the intermediate ones must be public_deps. data_deps don't count since they're
only runtime dependencies. If you think a dependency chain exists, it might be
because the chain is private. Try "gn path" to analyze.
```
Then I run the "gn path", it shows:
```
gn path out/an //xwalk/runtime/android/core_internal:xwalk_core_reflection_layer_java_gen //xwalk/runtime/android/core_internal:xwalk_core_internal_java__compile_java__javac
//xwalk/runtime/android/core_internal:xwalk_core_internal_java__compile_java__javac --[private]-->
//xwalk/runtime/android/core_internal:xwalk_core_internal_java__build_config --[private]-->
//xwalk/runtime/android/core_internal:xwalk_core_reflection_layer_java_gen

1 non-data path found. It is not public.
heke@heke-workstation:~/master/crosswalk-src/src$ gn path out/an //xwalk/runtime/android/core_internal:xwalk_core_reflection_layer_java_gen //xwalk/runtime/android/core_internal:xwalk_core_internal_java__compile_java__javac
//xwalk/runtime/android/core_internal:xwalk_core_internal_java__compile_java__javac --[private]-->
//xwalk/runtime/android/core_internal:xwalk_core_internal_java__build_config --[private]-->
//xwalk/runtime/android/core_internal:xwalk_core_reflection_layer_java_gen

1 non-data path found. It is not public.
```

build/android/BUILD.gn
heke123: So far I didn't find any options can take srcjar as input in cmd of "javadoc -help", neither in Ant. 
Anyway, let's postpone it when implementing "xwalk_core_library_andorid.gypi"

runtime/android/core_internal/BUILD.gn
rakuco: Like I said, this depends on crosswalk-project/chromium-crosswalk#384. Your branch doesn't seem to include aae862914235da65a06df5081aebad2aff71b49e.

heke123: Thanks, I'll make change on this.

heke123:
PR updated.  But I'm surprised to find several crashed test cases. I remember all the test cases were pass in my first version. Please have a look at this PR, at the same time I'll figure out what causes the crash.
Crash found in xwalk_core_shell:
C  564.254s Main  [  FAILED  ] org.xwalk.core.xwview.test.CookieManagerTest#testAcceptCookie (CRASHED)
C  564.254s Main  [  FAILED  ] org.xwalk.core.xwview.test.CookieManagerTest#testCookieExpiration (CRASHED)
C  564.254s Main  [  FAILED  ] org.xwalk.core.xwview.test.CookieManagerTest#testRemoveAllCookie (CRASHED)
C  564.254s Main  [  FAILED  ] org.xwalk.core.xwview.test.DownloadListenerTest#testOnDownloadStart (CRASHED)
C  564.254s Main  [  FAILED  ] org.xwalk.core.xwview.test.FindAsyncTest#testOnFindResultReceived (CRASHED)
Crash found in xwalk_core_internal_shell:
org.xwalk.core.internal.xwview.test.XWalkJavascriptResultTest#testOverrideBeforeUnloadHandling(CRASHED)

heke123:
@rakuco  All the test cases passed(in release build).  The previous fails are because I mistakenly run the debug version, they are broken by some DCHECKs.  
Please have a review, thanks.

BUILD.gn
```diff
@@ -363,14 +381,26 @@ source_set("xwalk_runtime") {
...
   ]
 
+  if (is_android) {
+    deps += [
+      "//components/cdm/browser",
+      "//xwalk/runtime/android:xwalk_core_jar_jni",
+      "//xwalk/runtime/android:xwalk_core_native_jni",
+    ]
+  } else {
+    # for windows and linux build:
```
rakuco: This comment is redundant and should be removed.

runtime/android/BUILD.gn
```diff
@@ -0,0 +1,54 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+shared_library("libxwalkcore") {
+  sources = [
+    "//xwalk/runtime/app/android/xwalk_entry_point.cc",
```
rakuco: If all the files are in a different directory, the entire target should probably be in `xwalk/runtime/app/android/BUILD.gn` instead.

rakuco:
There are only some minor things to fix in addition to what I mentioned inline:
* Some BUILD.gn files are still missing their copyright headers.
* There are unnecessary changes to version.gni -- I've commented about this twice before.
* You should probably remove `test/android/BUILD.gn` from this pull request, as it's still broken.

heke123:
@rakuco  PR updated.

- Make sure all BUILD.gn files have copyright header.
- I keep the version.gni's copyright/comments in the header totally unchanged(copy from build/utils/version.gni), 
- test/android/BUILD.gn removed, will implement it in next PR.

And, the Java targets in the deps list of `android_library` won't be included into the generated library. So in the `android_apk` target of app_hello_world.apk, only deps on `xwalk_app_runtime_java` is not enough, `xwalk_core_java` should also be depended.  So I removed the comments like: "#TODO(heke123): remove xwalk_core_java, it should be embedded in xwalk_app_runtime_java."

runtime/android/BUILD.gn
heke123: done.

rakuco:
lgtm!!!!!!!!!!!!!!!!
:+1: :smile:

heke123:
@rakuco,  Thanks very much!

halton:
Congratulations! Thanks a lot, both of you. @rakuco and @heke123

PR #3870, opened 2016-08-31
rebase next on top of master

msisov:


rakuco:
Note that merging a rebased branch does not work, as you can see from the list of commits that are going to be pulled in.
What you can do is explain if you had any merge conflicts and, if everything was fine, you can just force-push your branch or ask @mrunalk to take a look at your branch and do it himself.

mrunalk:
Yep @rakuco is right. It's easier if I rebase myself and force push to the next. I have done the same now and updated the cr-cw and v8-cw repos with latest from master as well as upstream. Sorry, if you had to wait for latest changes to land in next. Thanks for working on this anyway.

PR #3871, opened 2016-08-31
[gn] Rename xwalk_installer_package_deb to xwalk_installer_linux_debian

rakuco:
Use the same target name for the gyp and the GN versions, otherwise this
needlessly complicates the release build process and confuses everyone
trying to build Debian packages.

While here, add Chromium's missing copyright to `tools/installer/BUILD.gn`.

BUG=XWALK-7180

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -26,7 +26,7 @@ if (is_linux) {
   group("xwalk_installer") {
     testonly = true
     deps = [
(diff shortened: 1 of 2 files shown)


rakuco:
@heke123 fyi

pozdnyakov:
rs=me

PR #3872, opened 2016-08-31
[gn] Make xwalk_runtime depend on xwalk_resources

rakuco:
We need to depend on `xwalk_resources`, not `generate_xwalk_resources`:
the former copies certain headers produced by the latter to the location
`xwalk_resources` expects.

Without this dependency, the build can fail like this:

    ../../xwalk/runtime/browser/ui/desktop/exclusive_access_bubble.cc:12:10: fatal error: 'xwalk/grit/xwalk_resources.h' file not found
    #include "xwalk/grit/xwalk_resources.h"

BUG=XWALK-7180


pozdnyakov:
lgtm

rakuco:
@heke123 fyi

heke123:
Yes, I found this problem when build crosswalk-android too. Thanks to correct this.

PR #3873, opened 2016-09-01
reflection_generator: Stop requiring an int for a boolean option.

rakuco:
`reflection_generator.py` expects to receive an integer for the
`--verify-xwalk-apk` argument, which does not make sense. It was likely
added as a shortcut because gyp uses 1 and 0 instead of booleans, so one
could just do `--verify-xwalk-apk=<(my_int_var)` in gyp. Unfortunately,
that does not integrate so well with GN, so revamp how the argument is
handled:
* Turn the `--verify-xwalk-apk` into a boolean argument that does not
  require a value; if it is passed in the command line, its value is
  true, otherwise it is false.
* Adapt xwalk_android.gypi to follow suit and only pass that argument
  when `verify_xwalk_apk` is on.


rakuco:
@heke123 @darktears @pozdnyakov

pozdnyakov:
rs=me

heke123:
Thanks, I'll change in GN files accordingly.

PR #3874, opened 2016-09-02
[Android] Fix crash on file chooser

lincsoon:
- Fixed crash when no file is uploaded after the file chooser is launched.
- Do not invoke RuntimeFileSelectHelper when the uploaded file is null.

BUG=XWALK-7278
BUG=XWALK-7271

### DIFF
--- a/runtime/browser/android/xwalk_web_contents_delegate.cc
+++ b/runtime/browser/android/xwalk_web_contents_delegate.cc
@@ -121,18 +121,15 @@ void XWalkWebContentsDelegate::RunFileChooser(
     return;
   }
   int mode = static_cast<int>(params.mode);
-  jboolean overridden =
-      Java_XWalkWebContentsDelegate_shouldOverrideRunFileChooser(env,
-          java_delegate.obj(),
-          web_contents->GetRenderProcessHost()->GetID(),
-          web_contents->GetRenderViewHost()->GetRoutingID(),
-          mode,
-          ConvertUTF16ToJavaString(env,
-              base::JoinString(params.accept_types,
-                               base::ASCIIToUTF16(","))).obj(),
-          params.capture);
-  if (overridden == JNI_FALSE)
-    RuntimeFileSelectHelper::RunFileChooser(web_contents, params);
+  Java_XWalkWebContentsDelegate_shouldOverrideRunFileChooser(env,
+      java_delegate.obj(),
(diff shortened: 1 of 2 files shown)


lincsoon:
@fujunwei @xzhan96 Please help to review

fujunwei:
LGTM

runtime/browser/android/xwalk_web_contents_delegate.cc
```diff
@@ -121,18 +121,15 @@ void XWalkWebContentsDelegate::RunFileChooser(
...
+  Java_XWalkWebContentsDelegate_shouldOverrideRunFileChooser(env,
+      java_delegate.obj(),
+      web_contents->GetRenderProcessHost()->GetID(),
+      web_contents->GetRenderViewHost()->GetRoutingID(),
+      mode,
+      ConvertUTF16ToJavaString(env,
+          base::JoinString(params.accept_types,
+                           base::ASCIIToUTF16(","))).obj(),
+      params.capture);
 }
```
xzhan96: I did some investigation:
1. The `RuntimeFileSelectHelper` is the equivalent for chrome/browser/file_select_helper.cc. And in the chrome implementation, `RunFileChooser` will call `selectFile(...)` in SelectFileDialog.java by JNI, and then launch the select activity.

2. The root cause of the crash is that SelectFileDialog.java use class FileProviderHelper to get file's uri, which is set during the initialization in chrome/android/java/src/org/chromium/chrome/browser/init/ChromeBrowserInitializer.java, but there is not such logic in xwalk. I tried to implement it, but not successful.
3. So, it seems RuntimeFileSelectHelper is **never** used, even before the implementation of XWalkFileChooser in XWalkView and the commit #3857.

**Above all,**  my point is that, since xwalk nerver used RuntimeFileSelectHelper, and we already expose  XWalkFileChooser to user, **do you think it's better to delete RuntimeFileSelectHelper in this commit**?

lincsoon: `RuntimeFileSelectHelper` is still used in `xwalk/runtime/browser/runtiime.cc`, but I'm not sure the code over there is stil active.

lincsoon: We'll remove redundant files by another commit in the future.

xzhan96: LGTM, thanks

PR #3875, opened 2016-09-02
Clean up reflection_generator.py and make it work with GN.

rakuco:
This pull request contains several commits, but the main goal is to make
it work correctly with GN.

With gyp, we are just generating a lot of .java files in the build
directory and depending on the directory where they are written in
`xwalk_core_internal_java`. Depending on a directory does not work with
GN, depending on all those new files directly also fails [1], so we now
follow Chromium's approach of writing those files into zip files called
`.srcjar`s and making `xwalk_core_internal_java` depend on that instead.
This is done by introducing a `--use-srcjars` option (off by default)
while we have to support both gyp and GN.

[1] https://groups.google.com/a/chromium.org/forum/#!topic/gn-dev/GokFp6TorgI

All the commits before the one described above are cleanups to make the
script look saner and more Pythonic.

RELATED BUG=XWALK-3674

### DIFF
--- a/tools/reflection_generator/reflection_generator.py
+++ b/tools/reflection_generator/reflection_generator.py
@@ -26,7 +26,7 @@ from util import build_utils
 
 
 # Classes list that have to generate bridge and wrap code.
-CLASSES_TO_BE_PROCESS = [
+CLASSES_TO_PROCESS = [
     'ClientCertRequestHandlerInternal',
     'ClientCertRequestInternal',
     'CustomViewCallbackHandlerInternal',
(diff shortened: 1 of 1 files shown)


rakuco:
@heke123 @lincsoon @darktears @pozdnyakov

heke123:
LGTM

heke123:
@rakuco   I just found in target "`xwalk_core_library_documentation`" of `xwalk_core_library_android.gypi`, it references the `wrapper` java files directly. I'm not familiar with the `javadoc` usage, and I didn't find how to javadoc the srcjar files directly. So I have to unzip srcjar into java files first, then do the javadoc, it is not good.
How about to generate both srcjar and "filesystem files" when user set the "--use-srcjar" flag to reflection_generator.py?  Any ideas on this?  thanks!

PR #3878, opened 2016-09-02
Android: Drop generate_runtime_client_version.py in favor of version.py

rakuco:
It makes zero sense for `generate_runtime_client_version.py` to exist;
it's just a poor-man's version of upstream's `version.py`.

Just use the latter and simplify our code.


rakuco:
@heke123 @lincsoon @darktears @pozdnyakov

darktears:
Sure. LTGM

PR #3880, opened 2016-09-05
[Android] Clean up xwalk_core_{internal_}empty_embedder_apk

rakuco:
* `generate_resource_maps`: Use `direct_dependent_settings` to set
  variables that should be available to targets, there is no reason to
  make the variables available to targets that depend on it indirectly.
  While here, define `generated_src_dirs` instead of leaving it up to
  targets that depend on it to take care of it.
* `xwalk_core_empty_embedder_apk`: Stop depending on
  `generate_resource_maps`, it's not used.


rakuco:
@heke123 @lincsoon @darktears @wuhengzhi

lincsoon:
LGTM

PR #3882, opened 2016-09-06
[Android] Remove generate_xwalk_runtime_client_version

rakuco:
Usage of this target and the file it creates was removed in commit
b67a9909 ("[Android] Use reflection layer in xwalk core for runtime
usage") from 2 years ago, yet we were still producing this useless file
on every build.


rakuco:
@heke123 @lincsoon @darktears

darktears:
LGTM

PR #3883, opened 2016-09-06
[Android] Stop depending on xwalk_core_shell from xwalk_core_library

rakuco:
This makes `xwalk_core_library_android.gypi` stop depending on
`xwalk_android_tests.gypi`; the latter contains just test targets and
thus should not be required by anything not related to tests.

Instead, we now obtain the list of native libraries we want to copy from
the build system itself, which makes a lot more sense than having to
figure out what's in gyp/GN and what's hardcoded in an obscure Python
script.

`gdbserver` is also no longer copied (even in a debug build): the whole
xwalk_core_library thing targets a deprecated IDE (Eclipse) and I
honestly doubt people are debugging their debug Crosswalk builds with
the gdbserver binary from xwalk_core_library.

### DIFF
--- a/build/android/generate_xwalk_core_library.py
+++ b/build/android/generate_xwalk_core_library.py
@@ -11,10 +11,17 @@ import os
 import shutil
 import sys
 import zipfile
-from common_function import RemoveUnusedFilesInReleaseMode
 from xml.dom.minidom import Document
 
-XWALK_CORE_SHELL_APK = 'xwalk_core_shell_apk'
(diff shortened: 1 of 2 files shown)


rakuco:
@heke123 @lincsoon @darktears

rakuco:
Ugh, the libraries from `lib/` are not stripped and look gigantic... I'll have to fix that first.

rakuco:
@heke123 @lincsoon @darktears everything's working now, PTAL

darktears:
lgtm

PR #3888, opened 2016-09-08
[Android] Remove the generate_resource_maps target.

rakuco:
It is really difficult to figure out why it is necessary from commits
6f1f4c58 ("[Android] Fix that R.class in core library is set with
incorrect value") and e5bc61ac ("[Android] Generate resource maps for
internal packages"), but so much has changed in the Android code in
Crosswalk and Chromium compared to 2014 that it no longer seems to be
necessary.

Making `xwalk_core_internal_empty_embedder_apk` depend directly on
`xwalk_core_internal_java` does create many R.java files in the former's
`gen/` directory with several values being assigned (instead of just
being declared as was the case with `genereate_resource_maps`), but
creating Android apps with crosswalk-app-tools in both embedded and
shared mode with code that causes some of those resources to be
used (e.g. <input type="color /"> uses resources from `ui/android`)
resulted in no crashes, which seem to be the reason for
e5bc61ac ("[Android] Generate resource maps for internal packages") at
the time.

BUG=XWALK-2746
BUG=XWALK-2750
BUG=XWALK-2752


rakuco:
@lincsoon @heke123 @darktears @wuhengzhi

lincsoon:
LGTM

wuhengzhi:
LGTM.
In R.java, there are some variables, the type is int[] instead of int:

>     public static final class styleable {
        public static final int[] ButtonCompat = { 0x7f010002, 0x7f010003 };
        public static final int[] TextViewWithLeading = { 0x7f010004 };
    }

It could not be set in XWalkInternalResources.java:

> try {
                            int value = generatedInnerClazz.getField(field.getName()).getInt(null);
                            field.setInt(null, value);
                        }

rakuco:
> It could not be set in XWalkInternalResources.java:

Do you mean this is something that needs to be fixed regardless of this PR?

wuhengzhi:
Yes. Currently the exception(IllegalArgumentException) was caught In XWalkInternalResources.java.

rakuco:
I see; can you please file a bug or work on this then?

wuhengzhi:
Sure.

PR #3891, opened 2016-09-09
[Android] Remove tar.py and the pack_* targets from gyp

rakuco:
`pack_xwalk_core_library` was created years ago when we had our own
Cordova 3 fork and suggested people extracted the xwalk_core_library
tarball into its checkout.

This has not been the case for a long time, and our release build system
does not use any of those `pack_*` targets to create its zip files.
Anyone wishing to create a tarball out of the directories in `out/` can
do so manually.


rakuco:
@lincsoon @heke123 @wuhengzhi @darktears

wuhengzhi:
Should ```pack_xwalk_core_library_src``` also be removed?

rakuco:
As far as I can see, `pack_xwalk_core_library_src` is an action that's being removed as part of `pack_xwalk_core_library`.

wuhengzhi:
Oh, LGTM, thanks.

heke123:
LGTM

PR #3892, opened 2016-09-09
gn: Use existing targets to manipulate icudtl.dat and V8 snapshots

rakuco:
Instead of repeating the same checks for `icu_use_data_file` and
`v8_use_external_startup_data` along with changes to `sources` and
`deps` in many different files, we can leverage the existing upstream
targets whose purpose is to do the exact same thing.


rakuco:
@heke123

heke123:
LGTM

PR #3893, opened 2016-09-09
gn: Rename a few targets to prepare for M53.

rakuco:
Starting with M53, https://codereview.chromium.org/2109293003 enforces a
naming convention for certain Android targets, and if a target name
matches it's assumed to have a `.build_config` file with it.

* Rename `xwalk_app_strings` to `xwalk_app_strings_grd` so GN accepts it
  as a valid target that generates a `.build_config`.

* Rename `xwalk_core_reflection_layer_java_gen` to
  `xwalk_core_reflection_layer_gen` to achieve the opposite and have GN
  _stop_ assuming the target produces a `.build_config` (the "java" part
  was making it fall into a naming whitelist).

### DIFF
--- a/runtime/android/core/BUILD.gn
+++ b/runtime/android/core/BUILD.gn
@@ -7,7 +7,7 @@ import("//build/config/android/rules.gni")
 android_library("xwalk_core_java") {
   deps = [
     ":xwalk_core_java_resources",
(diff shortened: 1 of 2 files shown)


rakuco:
@heke123

rakuco:
@heke123: gn style issues fixed. please take a look

heke123:
LGTM

PR #3894, opened 2016-09-12
gn: Unify packing of xwalk{_100_percent}.pak into a single target

rakuco:
Get rid of a lot of repetition in the Android `BUILD.gn`s. There were 4
targets doing the exact same thing: calling `android_assets()` to pack
`xwalk.pak` and `xwalk_100_percent.pak` along with ICU and V8 blobs.
Some of those targets were then being depended on in other files,
leading to a needless entanglement in the build system.

Additionally, many `*_apk` targets were duplicating the dependency on
the `xwalk_pak` and `xwalk_resources_100_percent_pak` targets that were
already part of the `*_assets` targets.

We now unify all those targets into `//xwalk/runtime:pak_file_assets`,
and only depend on it from the `*_assets` Android targets (which is also
what upstream does with its targets).


rakuco:
@heke123

heke123:
LGTM

PR #3895, opened 2016-09-12
[Android] Fix the issue that date picker is not work

xzhan96:
In the patch #3857(decouple XWalkView from the activity), we need
use WindowAndroid instead of ActivityWindowAndroid as the top window
of XWalkView, but that will prevent DateTimeChooserAndroid activity
to to launched.

Here we create WindowAndroid or ActivityWindowAndroid according to
the context type.

BUG=XWALK-7269

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
@@ -56,6 +56,7 @@ import org.chromium.content_public.browser.NavigationController;
 import org.chromium.content_public.browser.WebContents;
 import org.chromium.content_public.browser.navigation_controller.UserAgentOverrideOption;
 import org.chromium.media.MediaPlayerBridge;
+import org.chromium.ui.base.ActivityWindowAndroid;
 import org.chromium.ui.base.WindowAndroid;
 import org.chromium.ui.gfx.DeviceDisplayInfo;
 import org.json.JSONArray;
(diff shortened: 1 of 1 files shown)


xzhan96:
@lincsoon @fujunwei please help to review, thanks

runtime/android/core_internal/src/org/xwalk/core/internal/XWalkContent.java
```diff
@@ -124,6 +125,17 @@ public void run() {
...
+        WindowAndroid wrapper;
+        boolean contextWrapsActivity = activityFromContext(context) != null;
+        if (contextWrapsActivity) {
+            wrapper = new ActivityWindowAndroid(context);
+        } else {
+            wrapper = new WindowAndroid(context);
+        }
+        return wrapper;
+    }
+
```
lincsoon: I think the code below is much simpler:

```
private static WindowAndroid getWindowAndroid(Context context) {
    if (WindowAndroid.activityFromContext(context) != null) {
        return new ActivityWindowAndroid(context);
    } else {
        return new WindowAndroid(context);
    }
}
```

xzhan96: Nice. 
But I don't understand what's the purpose of the code in AwContents.java as below. Is there any special consideration to save the WindowAndroidWrapper? Do we need porting this part totally?

   ```
private static WeakHashMap<Context, WindowAndroidWrapper> sContextWindowMap;
    @SuppressFBWarnings("LI_LAZY_INIT_STATIC")
    private static WindowAndroidWrapper getWindowAndroid(Context context) {
        if (sContextWindowMap == null) sContextWindowMap = new WeakHashMap<>();
        WindowAndroidWrapper wrapper = sContextWindowMap.get(context);
        if (wrapper != null) return wrapper;

        boolean contextWrapsActivity = activityFromContext(context) != null;
        if (contextWrapsActivity) {
            final boolean listenToActivityState = false;
            wrapper = new WindowAndroidWrapper(
                    new ActivityWindowAndroid(context, listenToActivityState));
        } else {
            wrapper = new WindowAndroidWrapper(new WindowAndroid(context));
        }
        sContextWindowMap.put(context, wrapper);
        return wrapper;
    }
```

xzhan96:
@lincsoon @fujunwei

xzhan96:
@lincsoon PTAL

PR #3896, opened 2016-09-13
[Android] Remove the xwalk_core_library_src gyp target

rakuco:
This target is not built by default and has thus been unmaintained for a
long time. It requires one to maintain a hardcoded list of directories
in gyp that needs to be updated on every rebase cycle, needs an ugly
script like `merge_java_srcs.py` and is just a hurdle to port to GN that
is not really justified these days (if ever).

BUG=XWALK-2587


rakuco:
@lincsoon @wuhengzhi @heke123

lincsoon:
LGTM. @rakuco Hengzhi has left from Intel

PR #3897, opened 2016-09-14
[Android] Rewrite generate_xwalk_core_library.py

rakuco:
Finally do something I've wanted for years: a completely rewrite of the
horrendous `generate_xwalk_core_library.py`. Whether it really needs to
exist is still up for debate.

Even though this commit changes a lot of lines, it is basically just
simplifying the script by making it dumber. All the file names and paths
have been moved to where they should have been in the first place: the
build system. We now properly depend on all targets that produce files
and directories we use, and pass them to the script via arguments. The
script is now only responsible for creating the right directories and
copying things around (doing processing on resources in some cases).

One of the largest changes is in `CopyResources()`: instead of acting on
the processed resource zip files, we now copy the original `res/`
directories from the source tree and do some minimal processing on them
ourselves (by avoiding file name conflicts and renaming some image
directories). This makes more sense than fully processing everything,
since this is actually something the user's build should do later.

These changes also make it easier to port the Android gyp code to GN, as
the file locations are different there.


rakuco:
@lincsoon @heke123

rakuco:
I've rewritten parts of the patch and instead of just copying resource directories directly from the source tree, I'm working on the generated zipped files again -- while it looks more impractical at the moment than the first iterations of this patch, this will make porting to GN easier.

build/android/generate_xwalk_core_library.py
```diff
@@ -21,324 +27,174 @@
...
+
+
+def CopyBinaryData(output_dir, binary_files):
+  res_raw_dir = os.path.join(output_dir, 'res', 'raw')
+  res_value_dir = os.path.join(output_dir, 'res', 'values')
+  build_utils.MakeDirectory(res_raw_dir)
+  build_utils.MakeDirectory(res_value_dir)
 
+  # Poor man's XML writer. It is safe to assume there are only ASCII characters
+  # in the entries we are going tow rite.
```
heke123: going tow rite  >>> going to write.

rakuco: Fixed, thanks.

heke123:
LGTM

PR #3900, opened 2016-09-20
[Android] Temporary disable the unified media pipeline by default

xzhan96:
This workaround will not work for long once Chromium makes this
the only code path. So after finding the solution, this patch should
be reverted.

BUG=XWALK-7231


xzhan96:
@lincsoon @rakuco please help to review, thanks.

PR #3901, opened 2016-09-20
[Android] Rewrite remaining generation scripts

rakuco:
This should finish all the required rewrites to the `generate_*.py` scripts in `build/android`.

Not only do they make more sense now and look cleaner, but using them from GN is a lot easier.

This follows what was done in #3897: all the logic and decisions about file paths are now the responsibility of the build system, and the scripts just copy files around and put them in the right directories.


rakuco:
@lincsoon @heke123

xwalk_core_library_android.gypi
```diff
@@ -441,7 +441,13 @@
...
           'action': [
             'python', '<(DEPTH)/xwalk/build/android/generate_xwalk_core_library_aar.py',
-            '-t', '<(PRODUCT_DIR)',
+            '--aar-path', '<(PRODUCT_DIR)/xwalk_core_library.aar',
+            '--android-manifest', '<(DEPTH)/xwalk/build/android/xwalkcore_library_template/AndroidManifest.xml',
+            '--classes-jar', '<(PRODUCT_DIR)/lib.java/xwalk_core_library_java.jar',
+            '--jni-abi', '<(android_app_abi)',
+            '--jni-dir', '<(PRODUCT_DIR)/xwalk_core_library/libs/<(android_app_abi)',
+            '--res-dir', '<(PRODUCT_DIR)/xwalk_core_library/res',
+            '--r-txt', '<(PRODUCT_DIR)/xwalk_core_internal_empty_embedder_apk/gen/R.txt',
```
heke123: In original logic it is "xwalk_core_empty_embedder_apk/gen/R.txt" but here it is "xwalk_core_**internal**_empty_embedder_apk"?

xwalk_core_library_android.gypi
```diff
@@ -466,8 +472,11 @@
           ],
           'action': [
             'python', '<(DEPTH)/xwalk/build/android/generate_xwalk_core_library_aar.py',
-            '-t', '<(PRODUCT_DIR)',
-            '--shared',
+            '--aar-path', '<(PRODUCT_DIR)/xwalk_shared_library.aar',
+            '--android-manifest', '<(DEPTH)/xwalk/build/android/xwalk_shared_library_template/AndroidManifest.xml',
+            '--classes-jar', '<(PRODUCT_DIR)/lib.java/xwalk_core_library_java_app_part.jar',
+            '--res-dir', '<(PRODUCT_DIR)/xwalk_shared_library/res',
+            '--r-txt', '<(PRODUCT_DIR)/xwalk_core_internal_empty_embedder_apk/gen/R.txt',
```
heke123: ditto

xwalk_core_library_android.gypi
rakuco: Fixed, thanks.

xwalk_core_library_android.gypi
rakuco: Also fixed.

heke123: by the way, I didn't understand well on the meaning of                     
```
 'outputs': [
             '<(PRODUCT_DIR)/xwalk_core_library_aar_intermediate/always_run',
            ],
```
when to set it as "stamp" and when to "always_run"? thanks.

rakuco: The "always_run" version is a hack that the previous people who worked on this file came up with -- you can see it's not present in Chromium itself. It basically points to a file that is never created, so the target is always run regardless of whether it needs to.
The stamp version is much more commonly used -- you pass a file that will be touched when the script finishes running, and the build system uses that to determine whether the inputs are more recent that it and only call the script if something changed.

heke123: I know it has nothing to do with this PR.
Why can't we just remove the 'output' if no stamp is needed?  Is the 'output' is mandatory in "action" of GYP?

rakuco:
Patch rebased on top of #3902.

xwalk_core_library_android.gypi
rakuco: `output` is mandatory for actions, just like in GN.

heke123:
LGTM

PR #3902, opened 2016-09-20
[Android] Do not include BUILD.gn in xwalk_app_template/

rakuco:
Since we do a simple shutil.copytree() to copy `app/android/app_template`
to `xwalk_app_template/template`, we end up including the `BUILD.gn`
present in the former.

Explicitly ignore `*.gn` files when copying to avoid that.


rakuco:
@heke123 
I'd like to land this before #3901 to prevent this fix from being lost in the rewrite.

heke123:
LGTM

PR #3904, opened 2016-09-20
gn: Implement project and AAR generation targets for Android

rakuco:
In gyp, they are `xwalk_core_library`, `xwalk_shared_library`,
`xwalk_core_library_aar` and `xwalk_shared_library_aar`.

The GN implementation requires more lines of code, but is hopefully a
lot easier to understand: developers only have to care about the
`xwalk_{core,shared}_library` targets. Everything else is internal to
their implementation.

Thanks to GN's architecture, we can use `build_config` files to write
down a lot of dependency information for each target and use that file's
contents when specifying which resources we want or certain JARs we want
to work or generate. For example, we do not need to manually specify any
of the zipped resource files we use, nor their corresponding paths in
the source tree; we also do not have to hardcode paths to certain JARs,
such as `xwalk_app_runtime_java.jar`.

The biggest piece of code is the `generate_android_project` template in
the newly-added `build/android/generate_android_project.gni. It performs
all the steps required to have a working project directory:
* Processing resources (`res/`) of all the specified dependencies.
* Generating a `NativeLibraries.java` wrapper for any native libraries
  listed as dependencies (e.g. `libxwalkcore.so`) to be loaded at
  runtime.
* Building a separate JAR with the code generated in the above steps.
* Merging this JAR with all the other JARs generated by the target's
  dependencies.
* Actually calling `generate_xwalk_core_library.py`.
* Generating an AAR file (via `generate_xwalk_core_library_aar.py`).

In gyp, the maze made of `xwalk_core_empty_embedder_apk`,
`xwalk_core_internal_empty_embedder_apk`, `xwalk_core_library_java`,
`xwalk_core_library_java_app_part` and
`xwalk_core_library_java_library_part` was used only to perform the
first 2 steps described above by creating empty, unused APKs and using
some of the byproducts of the APK's build.

With GN, we no longer need to pass either `--resource-zip-sources` or
`--resource-strings` to `generate_xwalk_core_library.py`, as all the
resource files are specified together in the build_config and the paths
are all derived from the original source locations. We can also separate
string and non-string resources based on their file names, as all string
targets end in `_grd`, and so do their zips.

Compared to `xwalk_core_library_java.jar` and `xwalk_core_library.aar`
from gyp, the GN versions have a few differences in what is packaged:
* We now ship `org.chromium.base.BuildConfig`, which is not really used
  by any of our code paths but may become required in the future.
* There are fewer `R$foo.class` files in both
  `org/chromium/components/web_contents_delegate_android` and
  `org/xwalk/core/internal` (seemingly with no adverse effects).
* We ship fewer Mojo classes, simply because the targets in `content`
  do not depend on all the Mojo targets the gyp version does.
* We no longer ship `org/xwalk/core/library/empty/R*.class`, as they are
  not used anywhere and are just a byproduct of the AndroidManifest.xml
  we pass when processing resources.

Compared to `xwalk_core_library_java_app_part.jar` and
`xwalk_shared_library.aar` from gyp, the GN versions differ in one aspect:
* They do not contain any resources coming from Chromium targets: they
  are also present in `XWalkRuntimeLibrary`, and that is where they
  should come from in shared mode (which is what uses the aforementioned
  JAR).

Finally, the existing Python scripts had to be modified slightly:
* Since some arguments to the scripts are passed in GN via the
  "@FileArg" expansion, we need to call `build_utils.ExpandFileArgs()`.
  There are no side-effects to gyp.
* `merge_jars.py` now takes the JARs to merge as positional arguments
  due to a limitation in the `ExpandFileArgs()` function mentioned
  above (it cannot expand multiple `@FileArg()` calls as a list value to
  a single argument).

### DIFF
--- a/build/android/generate_android_project.gni
+++ b/build/android/generate_android_project.gni
@@ -1,0 +1,302 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/config/android/rules.gni")
+
+# Runs generate_xwalk_core_library.py.
+#
+# Args:
+#   binary_files [optional]
+#     List of additional binary data (icudtl.dat, for example) that will be
+#     added to res/raw.
+#   build_config [required]
+#     Path to a .build_config file with resource and native libraries
+#     information (generally, a build_config with the "android_apk" type).
+#   has_native_libraries [optional]
+#     Boolean indicating whether there are architecture-dependent shared
+#     libraries to copy to the project.
+#   js_bindings [required]
+#     List of Crosswalk extension bindings .js files to add to res/raw.
+#   output_dir [required]
+#     The project directory to be created. If it already exists, it will be
+#     erased.
+#   template_dir [required]
+#     Directory to the template directory containing a ant.properties,
+#     AndroidManifest.xml and others.
+template("generate_xwalk_core_library") {
+  assert(defined(invoker.build_config))
+  assert(defined(invoker.js_bindings))
+  assert(defined(invoker.output_dir))
+  assert(defined(invoker.template_dir))
+
+  _build_config = invoker.build_config
+  _js_bindings = invoker.js_bindings
+  _output_dir = invoker.output_dir
+  _template_dir = invoker.template_dir
+
+  action(target_name) {
+    forward_variables_from(invoker, [ "deps" ])
+
+    script = "//xwalk/build/android/generate_xwalk_core_library.py"
+    inputs = [
+      "//build/android/gyp/package_resources.py",
+      "//build/android/gyp/util/build_utils.py",
+    ]
+    outputs = [
+      "$target_gen_dir/${target_name}__generate.stamp",
+    ]
+
+    _stamp = rebase_path("$target_gen_dir/${target_name}__generate.stamp",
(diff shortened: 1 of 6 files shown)


rakuco:
@heke123 @lincsoon

rakuco:
I have a patch ready to add `xwalk_app_template` to GN, but chose to send it after this one to avoid creating an even bigger patch.
That leaves us with `xwalk_core_library_documentation` and the Maven bits, which can be implemented later.

build/android/generate_android_project.gni
```diff
@@ -0,0 +1,294 @@
...
+#   js_bindings [required]
+#     List of Crosswalk extension bindings .js files to add to res/raw.
+#   output_dir [required]
+#     The project directory to be created. If it already exists, it will be
+#     erased.
+#   template_dir [required]
+#     Directory to the template directory containing a ant.properties,
+#     AndroidManifest.xml and others.
+template("generate_xwalk_core_library") {
+  _build_config = invoker.build_config
```
heke123: For all "required" parameters, how about to add assert like `assert(defined(invoker.build_config))`?  just like `src/build/config/android/rules.gni` does.

rakuco: Done.

heke123:
LGTM

heke123:
LGTM, thanks:)

PR #3905, opened 2016-09-21
gn: Implement the xwalk_app_template target

rakuco:
This is fairly simple after commit cadbd11 ("[Android] Rewrite
generate_app_packaging_tool.py"): we just need to depend on the
recently-added `xwalk_core_library` and `xwalk_shared_library` targets
and specify the paths to everything we want to copy.

We also leverage GN's build_configs to determine the path to the runtime
JAR without having to hardcode it anywhere.

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -34,6 +34,7 @@ group("xwalk_builder") {
 
       # For external testing.
       "app/android/app_hello_world:xwalk_app_hello_world_apk",
+      "app/android/app_template:xwalk_app_template",
       "runtime/android/core:xwalk_core_library",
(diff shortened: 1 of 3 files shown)


rakuco:
@heke123

heke123:
After this PR, the only thing left is "xwalk_core_library_documentation" for GN Crosswalk-Android, right?

rakuco:
There's the documentation target and the POM ones (dealt with in #3906).

rakuco:
Oops, I thought there was an lgtm in your comment so I clicked the merge button :(

heke123:
LGTM :)

PR #3908, opened 2016-09-21
gn: Add the xwalk_core_library_docs target.

rakuco:
Unfortunately we still need to manually list all the files we want to
process as per
https://lists.crosswalk-project.org/pipermail/crosswalk-dev/2016-September/003927.html

We cannot call `javadoc` directly in GN, so a small wrapper,
`generate_javadoc.py`, has been added. It also knows how to deal with
`.srcjar` files, which is how the reflection wrapper code is currently
packaged.

The gyp version of the target has also been changed to use the wrapper
so that the targets in both build systems do not differ too much.

There is no `xwalk_core_library_java_app_part` target in GN, so we use
`xwalk_core_java` instead (and adjust gyp accordingly): the `app_part`
target was just merging `xwalk_core_java.jar` and the JARs it depends
on, but the documentation only needs the classes from
`xwalk_core_java.jar` itself, so there is no difference in the output.


rakuco:
@lincsoon @heke123

runtime/android/core/BUILD.gn
```diff
@@ -146,3 +146,73 @@ generate_android_project("xwalk_shared_library") {
...
+    rebase_path(android_sdk_jar, root_build_dir),
+    "--classpath",
+    "@FileArg($_core_java_build_config:deps_info:jar_path)",
+    "--input-srcjar",
+    rebase_path(_srcjar, root_build_dir),
+    "--java-files",
+    "$_rebased_java_files",
+    "--output-dir",
+    rebase_path("$root_out_dir/xwalk_core_library_docs", root_build_dir),
+    "--srcjar-files",
```
heke123: I remember we have a `TODO` in` reflection_generator.py` says something like "it will only output a wrapper.srcjar when get rid of GYP and no wrapper_java_files anymore".  I guess maybe we don't need to provide the option `--srcjar-files` in `generate_javadoc.py`? 
In original xwalk_core_library_android.gypi, it doesn't use the --srcjar-files option.
In our new GN logic, the `wrapper.srcjar` is assigned to `--input-srcjar` already, why to set the `_wrapper_java_files` to `--srcjar-files` again?

heke123: @rakuco   Understand now, thanks.  Still have something to discuss:
1) If we have to use the `_wrapper_java_files` to keep a part file list of the `wrapper.srcjar`,  we have to also keep `reflection_generator.py` outputs both srcjars and java files even after we get rid of GYP.(conflict with our TODO in reflection_generator.py).
2) I suppose the "gn gen" would complain "don't know who generates the java files in `_wrapper_java_files`".  But apparently success of the trybot tells me GN didn't complain anything. I guess maybe it is because the `_wrapper_java_files` is not in the `inputs` or `sources` of this `action`. It is ok,  only set the `wrapper.srcjar ` in the `sources` is enough.

LGTM.   Above discussions shouldn't block you merge this PR I think.

rakuco: I think you're getting confused -- the files in `_wrapper_java_files` are not tracked by GN at all. Think of it just as a list of strings. They only have meaning inside `generate_javadoc.py`, which uses them to reference files that are extracted from the srcjar into a temporary directory. Only the srcjar itself is tracked by GN.

heke123: >  just as a list of strings. They only have meaning inside generate_javadoc.py
Exactly true.  totally understand now. thanks:)

PR #3910, opened 2016-09-26
gn: Implement the "xwalk_runtime_lib_lzma_apk" target

rakuco:
This was hopefully the last missing Android target in the gyp->GN
conversion, and it is an ugly one: we are forced to keep a hardcoded
list of all the files we want to compress and ship as assets, which
makes the change to `runtime/android/runtime_lib/BUILD.gn` bigger than
usual.

Since `android_assets()` in GN requires a list of all files being
shipped as assets, we can simplify `lzma_compress.py` and get rid of its
different operation modes: instead of passing `--mode=show-output-names`
we just list all files that are being compressed in gyp.

While here, also use a temporary directory to perform the compressions
in `lzma_compress.py`, as it's then removed automatically in case
something goes wrong and we are not left with dangling files in the
build tree.

BUG=XWALK-7285

### DIFF
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -41,6 +41,7 @@ group("xwalk_builder") {
       "runtime/android/core:xwalk_shared_library",
       "runtime/android/core:xwalk_shared_library_pom_gen",
       "runtime/android/runtime_lib:xwalk_runtime_lib_apk",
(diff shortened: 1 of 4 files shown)


rakuco:
@heke123

heke123:
@rakuco Thanks!
LGTM

PR #3912, opened 2016-09-28
[Windows] wifidirect: Stop checking for Windows.md

rakuco:
Synchronously running a command to check if `Windows.winmd` is present
makes it more difficult to port Crosswalk on Windows to GN, as we would
at the very least need to do the same thing through a Python script and
a synchronous script call.

We can safely assume that `Windows.md` is always present though, as it
is installed by the Windows 10 SDK required by Chromium as a build
dependency.

RELATED BUG=XWALK-7336


rakuco:
@darktears @astojilj @heke123

darktears:
Sounds ok to me but I let @astojilj to do the final review.

darktears:
LGTM

PR #3914, opened 2016-09-29
ApplicationProtocolHandler: Remove |is_authority_match|.

rakuco:
Ever since commit 2bdc24a1 ("[Runtime] Return a "NOT FOUND" message if
the resource is invalid when using application scheme"), we are bailing
out early in `ApplicationProtocolHandler::MaybeCreateJob()` when there
is no `ApplicationData`.

This means we can remove all the further checks for `application.get()`
and get rid of `is_authority_match`, since by the time a
`UrlRequestApplicationJob` is created its value will always be true.

Not only does this make the code more readable, but it avoids a build
warning turned error in MSVC2015 when building the code with stricter
flags in GN:

    application_protocols.cc(307): warning C4800: 'xwalk::application::ApplicationData *': forcing value to bool 'true' or 'false' (performance warning)

RELATED BUG=XWALK-7336

### DIFF
--- a/application/browser/application_protocols.cc
+++ b/application/browser/application_protocols.cc
@@ -52,14 +52,11 @@ namespace {
 net::HttpResponseHeaders* BuildHttpHeaders(
     const std::string& content_security_policy,
     const std::string& mime_type, const std::string& method,
-    const base::FilePath& file_path, const base::FilePath& relative_path,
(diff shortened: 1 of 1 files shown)


rakuco:
@darktears @pozdnyakov

pozdnyakov:
lgtm

PR #3915, opened 2016-09-30
[Windows] Add a msbuild.gypi and clean up the MSBuild Python wrapper

rakuco:
Instead of manually calling `msbuild_dotnet.py` and then renaming
`xwalk_dotnet_bridge.dll` every time one wants to build a .NET
extension, do that only once in `msbuild.gypi`: the other targets just
need to define a few variables and include the new gypi file, which then
takes care of doing all the rest.

This considerably reduces the size of the other gyp files and avoids a
lot of duplication.

Clean up `msbuild_dotnet.py` itself a little:
* Stop invoking `MSBuild` from `//xwalk/extensions/test/win`, which
  creates some needless complications in the WiFi Direct gyp code.
  Instead, just invoke it from the root of the build directory (the
  default).
* Use `argparse` instead of `optparse` to avoid having to manually
  validate required arguments.
* Pass `/nologo` and `/verbosity:quiet` to `MSBuild` to make it quieter.
* Split MSBuild's parameters into a separate variable to make the code
  easier to understand.

RELATED BUG=XWALK-7336

### DIFF
--- a/build/msbuild.gypi
+++ b/build/msbuild.gypi
@@ -1,0 +1,81 @@
+# Copyright (c) 2016 Intel Corporation. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This file is meant to be included into a target in order to invoke MSBuild,
+# generate an extension .dll and its respective accompanying bridge DLL from a
+# number of C# source files.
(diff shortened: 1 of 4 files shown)


rakuco:
@darktears @heke123

darktears:
look a lot nicer thanks.

LGTM

PR #3916, opened 2016-09-30
[Windows] Remove unreferenced variables in the .NET bridge code

rakuco:
This avoids the following type of error when building the bridge code
with stricter flags (on by default with GN):

    xwalk_dotnet_bridge.cc(388): error C2220: warning treated as error -
    no 'object' file generated
    xwalk_dotnet_bridge.cc(388): warning C4189: 'returnValue': local
    variable initialized but not referenced

RELATED BUG=XWALK-7336


rakuco:
@darktears @heke123

darktears:
LGTM

PR #3917, opened 2016-09-30
gn: Use the AndroidManifest.xml from the template when creating AARs

rakuco:
We were previously creating the Android AARs using the
`AndroidManifest.xml` specified via `invoker.android_manifest`.

While that is correct for all other usages in the
`generate_android_project` template, using it in the AARs we create is
wrong, as the manifest specified via `invoker.android_manifest` is
generally the one in `//xwalk/runtime/android/core_internal_empty`
instead of the one
`//xwalk/build/android/xwalk{core,_shared}_library_template`. The ones
in `build/` specify a different package name and don't have an
`<application>` tag. This tag conflicts with the ones in the manifest
generated by Cordova when building an app, meaning Android GN builds
were not working with Cordova at all.

BUG=XWALK-7348

### DIFF
--- a/build/android/generate_android_project.gni
+++ b/build/android/generate_android_project.gni
@@ -262,6 +262,12 @@ template("generate_android_project") {
   action(_aar_target) {
     _aar_path = "$root_out_dir/${_target_name}.aar"
 
+    # Curiously enough, the AARs should use the AndroidManifest.xml present in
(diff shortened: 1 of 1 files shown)


rakuco:
@darktears @heke123

darktears:
LGTM

PR #3919, opened 2016-09-30
gn: Add support for Windows

rakuco:
The first commit just adjusts `msbuild_dotnet.py` to work with both gyp
and GN (its commit message is a lot more explanatory).

The second one includes everything required to build Crosswalk for
Windows using GN. Thankfully not many changes were needed: the work was
more similar to porting Linux than the dreadful task of porting Android.

Most of the new files are just GN versions of some .gypi wrappers, and
existing BUILD.gn files were augmented to include everything required by
the Windows code.

`generate_crosswalk_zip.py` was slightly modified to use gyp/GN lists
instead of separate arguments for the list of files and directories to
compress, as that works with gyp but not GN.

Big thank-you to Ke He <ke.he@intel.com> for starting the work!

BUG=XWALK-7336


rakuco:
@darktears @heke123

darktears:
LGTM

PR #3920, opened 2016-10-04
gn: Make xwalk_shared_library depend on additional resource targets

rakuco:
It is not clear if this is a bug or due to some odd design, but the
xwalk_shared_library Android project needs resources from both
`runtime/android/core_internal` (due to its usage of the reflection
layer wrapper classes) as well as resources from Chromium itself that
the core_internal code depends on.

These same dependencies are also part of `XWalkRuntimeLib` itself, so it
means we need to duplicate them.

For now, introduce a new variable, `core_internal_java_chromium_deps` to
make it clear that this is hopefully something provisional and add some
informative comments instructing people to not add non-//xwalk
dependencies to `xwalk_core_internal_java`s `deps` list.

BUG=XWALK-7350
BUG=XWALK-7358
RELATED BUG=XWALK-7366

### DIFF
--- a/runtime/android/core/BUILD.gn
+++ b/runtime/android/core/BUILD.gn
@@ -8,6 +8,7 @@ import("//third_party/icu/config.gni")
 import("//xwalk/build/android/generate_android_project.gni")
 import("//xwalk/build/android/maven_pom.gni")
 import("//xwalk/build/version.gni")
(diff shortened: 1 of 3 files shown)


rakuco:
@darktears @lincsoon @heke123

darktears:
LGTM

PR #3926, opened 2016-10-06
[Windows] Fix gyp build after f4a0f97

rakuco:
The sanity check introduced in f4a0f97 ("gn: Copy //xwalk/VERSION to the
build dir for generate_crosswalk_win_zip") broke the gyp build.

gyp passes relative paths to the script and mixes "/"s and "\"s, which
broke the os.path.commonprefix check. Fix it by normalizing all paths
first.

RELATED BUG=XWALK-7336

### DIFF
--- a/build/win/generate_crosswalk_zip.py
+++ b/build/win/generate_crosswalk_zip.py
@@ -24,7 +24,12 @@ from util import build_utils
 
 
 def PathInZipArchive(fs_path, root_dir):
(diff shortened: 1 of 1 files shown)


baleboy:
LGTM. Out of curiosity, why didn't the bots catch this?

rakuco:
The Windows Buildbot slave did catch this, but since we don't have a Windows trybot we only found out after the original commit was merged.

PR #3928, opened 2016-10-07
[Android] Sync find_helper.{cc,h} copies with upstream

rakuco:
Incorporate changes that went into the upstream versions of
`find_helper.{cc,h}`:
* 1cfca29 Fix for problem with activating find-in-page match via find
  tickbar.
* c0b762b Implement a shell of FindRequestManager, and hook it up to
  WebContentsImpl.

In addition to reducing the delta to upstream and making updating the
code in the future easier, it actually makes the code work in debug
mode (we were hitting some DCHECKs because the code had not been
updated).

RELATED BUG=XWALK-7325

### DIFF
--- a/runtime/browser/android/find_helper.cc
+++ b/runtime/browser/android/find_helper.cc
@@ -24,6 +24,7 @@ FindHelper::FindHelper(WebContents* web_contents)
       async_find_started_(false),
       find_request_id_counter_(0),
       current_request_id_(0),
+      current_session_id_(0),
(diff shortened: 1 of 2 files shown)


rakuco:
@lincsoon @xzhan96 
Reason number 28430928492912313563 not to copy entire files from upstream.

rakuco:
@mrunalk

rakuco:
@chuan9

chuan9:
LGTM!

PR #3929, opened 2016-10-07
[Android] Dispatch a mouse event when faking a click in the tests

rakuco:
This is a backport of the same android_webview commit from
M53 (afc3aed86 in Chromium, "Fix android webview unit tests to dispatch
correct click event"). It is being landed separately into M52 to reduce
the size of the M53 pull request and to make it easier to identify in
the history that this is being done.

Starting with M53, creating a generic event will not cause a click to be
generated as per Chromium bug 520519, so adjust the code to create a
`MouseEvent` instead.

RELATED BUG=XWALK-7325


rakuco:
@lincsoon @chuan9 @xzhan96 @mrunalk

chuan9:
LGTM!

PR #3930, opened 2016-10-07
[Android] Remove DCHECK from CookieManager

rakuco:
Another file copied from upstream, another crash in debug mode. This
`DCHECK` was removed from the android_webview version of `CookieManager`
in 2013 (commit 23dc0bb, "Android WebView: Allow CookieManager to be
used without starting Chromium").

Since then the upstream version has undergone several modifications
while ours remained mostly unchanged. After so many years it's very
likely the `DCHECK` is just wrong and can be removed, otherwise the
following the `CookieManager` Android tests crash in debug mode.


rakuco:
@lincsoon @chuan9 @xzhan96

xzhan96:
LGTM

chuan9:
LGTM

PR #3931, opened 2016-10-07
M53 gardening: adapt to https://codereview.chromium.org/2067843003

rakuco:
Add some missing verifiers for TLS/QUIC sockets, following what upstream
has done.

This makes
org.xwalk.core.xwview.test.ClearSslPreferenceTest#testDeniedRequestForSslError
as well as regular SSL connections to websites on any platform (Android
and others) stop crashing.

RELATED BUG=XWALK-7325


rakuco:
I'm sending this PR just so that it is registered in XWALK-7325.

PR #3932, opened 2016-10-07
Depend on blink, not blink_headers, in the extensions target

rakuco:
blink_headers is supposed to be used when targets only need access to
Blink structs and enums, which is not the case here, as
`xwalk_v8tools_module.cc` actually calls Blink code.

This was preventing `xwalk_extensions_unittest` from linking in debug
mode with GN and `is_component = true`. gyp was not affected because the
`xwalk_extensions` target there is always built as a static library, but
the code was updated there too for completeness.

### DIFF
--- a/extensions/BUILD.gn
+++ b/extensions/BUILD.gn
@@ -59,7 +59,7 @@ source_set("extensions") {
     "//base",
     "//content",
     "//ipc",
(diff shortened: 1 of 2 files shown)


rakuco:
@darktears @heke123 @pozdnyakov

darktears:
LGTM

PR #3934, opened 2016-10-10
[Android] Temporary disable the unified media pipeline by default

xzhan96:
Chromium start use desktop media pipeline on Android from version 52,
see the background: http://crbug.com/507834 and
https://docs.google.com/document/d/1QB4RVu0zT54Bys7dUHU6QeR3DyjOs-wk-MN9tjEzquQ/edit?usp=sharing

Currently, we still can use this patch to disable unified media pipeline,
that means use WebMediaPlayerAndroid instead of WebMediaPlayer_Impl, and
not use FFmpeg for demuxing and audio decoding.

However, this is just a temporary workaground, need to continue working
on find the final solution before Chromium make unified media pipeline
as the only code path.

BUG=XWALK-7231


xzhan96:
@rakuco please help to merge

PR #3933, opened 2016-10-09
    Add XWalkViewAsynchronousApisTest refering to upstream's     implementation. Remove the original FindAynsc test accordingly.

chuan9:
Add XWalkViewAsynchronousApisTest refering to upstream's
    implementation. Remove the original FindAynsc test accordingly.
    RELATED BUG=XWALK-7325

PR #3935, opened 2016-10-11
Add XWalkViewAsynchronousApisTest referring to upstream's implementation. 

chuan9:
Add XWalkViewAsynchronousApisTest referring to upstream's implementation. Remove the original FindAynsc test accordingly.
RELATED BUG=XWALK-7325


test/android/core/javatests/BUILD.gn
```diff
@@ -42,7 +42,8 @@ android_library("xwalk_core_test_java") {
     "src/org/xwalk/core/xwview/test/EvaluateJavascriptTest.java",
     "src/org/xwalk/core/xwview/test/ExtensionEcho.java",
     "src/org/xwalk/core/xwview/test/ExtensionEchoTest.java",
-    "src/org/xwalk/core/xwview/test/FindAsyncTest.java",
+    "src/org/xwalk/core/xwview/test/XWalkViewFindApisTestBase.java",
```
rakuco: This is no longer alphabetically sorted.

test/android/core/javatests/src/org/xwalk/core/xwview/test/FindAsyncTest.java
```diff
@@ -1,59 +0,0 @@
...
-        super.setUp();
-
-        mOnFindResultReceivedHelper = mTestHelperBridge.getOnFindResultReceivedHelper();
-        setFindListener();
-    }
-
-    @SmallTest
-    @Feature({"FindAllAsync"})
-    public void testOnFindResultReceived() throws Throwable {
-        String fileContent = getFileContent("find.html");
```
rakuco: There's nothing referencing `find.html` any longer.

rakuco:
> Add XWalkViewAsynchronousApisTest refering to upstream's implementation.

s/refering/referring/

You're also missing an "[Android]" in the title.

rakuco:
> It's also not totally clear to me if you've decided to replace the existing tests altogether because you think the upstream ones are better (for one reason or another) or because I mentioned them in JIRA and #3933: it's important that you understand why the previous tests were prone to failure and how the upstream infrastructure, with findAllAsyncOnUiThread() in particular, do not have the same issues. Whether the things you are testing themselves also need to change is a separate issue.

I am still waiting for a comment here.

> s/refering/referring/

This is still wrong.

chuan9:
@rakuco Sorry for late response. I replace the existing tests for three resons. 1. Upstream's test case is the right case. They have better hierarchy. Our existing patch in XWalkViewBase.java hava many specific function which is used only for find test, so it's better to split XWalkViewFindApisTestBase.java from XWalkViewTestBase.java.  Also, as you said, they have elegant functions to wait until finding is over while not hack function(Thread.sleep) which is in my first patch. 2.You mentioned it in the JIRA. 3. Remove existing test and resource(find.html) and keep same with upstream can make us find issue easier in the later test.
As for root cause of this test failure, from the log, I can see that callback "onFindResultReceived" is called four times when invoke mXwalkView.findAllAsync() while it's called three times in the original test. so original test can past when wait for 3 callbacks. Upstream changed somewhere to cause this failure, but I can't find where because XWalkCoreShell's debug version can't work now, so It may take a long time to debug by adding log. But we can ignore these changes because our tests are same with upstream and can pass, so our embedding apis can work same as android webview.

chuan9:
@rakuco  Thanks for your view, really appreciate your work!

PR #3936, opened 2016-10-13
gn: Add more dependencies to the xwalk_core_reflection_layer_gen target

rakuco:
We were previously not declaring any `sources` or `inputs`, which means
the target was not being run when any of the files it processes was
modified. In practical terms, it means changing files in `core_internal`
did not run the reflection scripts, which lead to a disconnect between
`core` and `core_internal` and build failures.

Fix it by manually declaring all files this target depends on (the
authors of the gyp side cheated by calling `find(1)`). Since these files
are also used by `xwalk_core_internal_java`, some new variables were
introduced to avoid duplication and to make sure the list is kept
up-to-date.

BUG=XWALK-7385


rakuco:
@darktears @heke123

darktears:
LGTM

PR #3937, opened 2016-10-13
[Android] Deprecate XWalkView.load() and split it into multiple methods

rakuco:
Users are now expected to use `loadUrl()`, `loadData()` or
`loadDataWithBaseURL()`, which are the same names used by the default
Android WebView. `load()` has been deprecated, not removed, to keep
backwards compatibility (the `@Deprecated` tags will only show up in
`XWalkView.java` once XWALK-7384 is fixed though). `API_VERSION` has
been bumped accordingly.

`XWalkContent.load()` tried to be too smart and implement those 3
methods without having enough information to decide which one to run
when called. Specifically, there was no equivalent to `loadData()`,
which loads a request with a "data:" URL, so whenever one needed to load
some data into a XWalkView the code path mimicked the Android WebView's
`loadDataWithBaseURL()` instead. Not to mention that most users just
want to load a URL, and `load(<address>, null)` is more tedious and less
clean to write than `loadUrl(<address>)`.

The lack of a `loadData()` equivalent in `load()` was causing
`org.xwalk.core.internal.xwview.test.XWalkJavascriptResultTest#testOverrideBeforeUnloadHandling`
to fail in the upcoming Chromium M53 because of
http://crrev.com/1890493002.

We relied on `blink::FrameLoader::startLoad()` calling `shouldClose()`
to trigger the machinery that delivers the "onbeforeunload" JavaScript
event we depended on in that test. While
`blink::FrameLoader::shouldContinueForNavigationPolicy()` could end up
calling `decidePolicyForNavigation()` which would trigger this same
machinery, that function was bailing out earlier in the check for an
empty request URL, as `XWalkContent.doLoadUrl()` was calling
`LoadUrlParams.createLoadDataParamsWithBaseUrl()` to create a data
request. That function creates a request with an empty URL and the data
as Blink's `SubstituteData`. What we need is
`LoadUrlParams.createLoadDataParams()`, which just creates a request
with the data in the URL.

The previous `XWalkContent.doLoadUrl()` implementation was moved to
`XWalkViewInternal.load()`, as we need to preserve compatibility with
the old code in the deprecated functions. All the new code comes from
Chromium's `android_webview/`, which also inspired the changes to
`XWalkViewInternalTestBase.java` and `XWalkViewTestBase.java`.

RELATED BUG=XWALK-7325

### DIFF
--- a/API_VERSION
+++ b/API_VERSION
@@ -1,2 +1,2 @@
-API=7
+API=8
 MIN_API=5
--- a/app/android/runtime_client/src/org/xwalk/app/runtime/XWalkCoreProviderImpl.java
+++ b/app/android/runtime_client/src/org/xwalk/app/runtime/XWalkCoreProviderImpl.java
@@ -112,7 +112,7 @@ class XWalkCoreProviderImpl implements XWalkRuntimeViewProvider {
 
     @Override
     public void loadAppFromUrl(String url) {
-        mXWalkView.load(url, null);
+        mXWalkView.loadUrl(url);
     }
 
     @Override
@@ -146,6 +146,6 @@ class XWalkCoreProviderImpl implements XWalkRuntimeViewProvider {
 
     @Override
     public void loadDataForTest(String data, String mimeType, boolean isBase64Encoded) {
-        mXWalkView.load("", data);
+        mXWalkView.loadData(data, mimeType, isBase64Encoded ? "base64" : null);
     }
 }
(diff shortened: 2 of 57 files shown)


rakuco:
@lincsoon @xzhan96 @chuan9 
This patch is fairly big because a lot of files are being touched, but the focus should be on XWalkContent.java, XWalkViewInternal.java, XWalkViewInternalTestBase.java and XWalkViewTestBase.java.

chuan9:
LGTM! Thanks!

lincsoon:
LGTM

PR #3939, opened 2016-10-17
tools/installer: Import ffmpeg_options.gni

rakuco:
`is_ffmpeg_component` is defined there, but we were not importing the
file, which leads to an error in M54.


rakuco:
@heke123

pozdnyakov:
rs=me

PR #3940, opened 2016-10-18
[Android] Depend on the native library targets when generating a project

rakuco:
This fixes a regression introduced in 02104a7 ("Roll Chromium
53.0.2785.143"). Specifically, when the code in `build/android` was
updated due to https://codereview.chromium.org/2082453003 upstream
callers of `generate_android_project` no longer needed to have targets
such as `libxwalkcore` in `deps`, but the `generate_android_project`
itself was not depending on them at all. In practice, this could lead to
failures where `generate_xwalk_core_library.py` would try to operate on
shared libraries that had not yet been built.


pozdnyakov:
lgtm

PR #3943, opened 2016-10-21
Import ffmpeg_options.gni in more locations

rakuco:
The fix in ba159bc ("tools/installer: Import ffmpeg_options.gni") needs
to be extended to avoid breaking the upcoming M54 build -- that file
needs to be imported in all locations that make use of
`is_component_ffmpeg`.


kenchris:
rs=me

PR #3945, opened 2016-10-24
gn: Import //v8/gni/v8.gni, not //build_overrides/v8.gni

rakuco:
We were importing the latter to access `v8_use_external_startup_data`.
However, we also had to import http://crrev.com/2058033002 in order to
override its value in `build/android.gni`, which means the value is
actually coming from `//v8/gni/v8.gni` instead.

Import the latter now to avoid errors in the M54 build.


kenchris:
rs=me

PR #3948, opened 2016-10-26
[Android] Fixed crash of XWalkPreferences

lincsoon:
Everytime XWalkPreference.sevValue() is invoked, it checks if previous
listeners exist, if exist and the key is ANIMATABLE_XWALK_VIEW, if
will throw an runtime exception.

The listener is registered when an XWalkView is created and unregistered
when an XWalkView is destroyed. But after XWalkView is decoupled from
the activity, XWalkView can not be destroyed automatically when an
activity is closed. So previous listeners are still remaining  and have
the app crashed.

This patch changs the behavior on this situation, warning that this
setting is not effective to previous XWalkViews instead of throwing the
exception.

BUG=XWALK-7309

### DIFF
--- a/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkPreferencesInternal.java
+++ b/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkPreferencesInternal.java
@@ -4,6 +4,8 @@
 
 package org.xwalk.core.internal;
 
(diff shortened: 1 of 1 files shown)


lincsoon:
@xzhan96 @fujunwei Please help to review

xzhan96:
lgtm

PR #3949, opened 2016-11-02
[Android] Fixed test errors

lincsoon:
The test case of OnShowOnHideCustomViewTest is to test whether the video
is played after click the play button. Bug sometimes the page is not
loaded completely when the click action is performing. By this patch, it
tries to click for more times to make sure the click action is effective.

BUG=XWALK-7197


halton:
LGTM

PR #3950, opened 2016-11-04
[Android] Reflection generator supports the @Deprecated annotation

lincsoon:
When generating the reflection layer, if a method that is marked as
@XWalkAPI is markd as @Deprecated as well, the @Deprecated annotation
will also appears in the new generated wrapper class. This will help to
make the javadoc display more precise method information.

BUG=XWALK-7384


lincsoon:
@xzhan96 @fujunwei Please help to review

xzhan96:
LGTM

PR #3951, opened 2016-11-08
[Android] Move user data to profile directory

fujunwei:
CookieManger isn't used at starting crosswalk, so the datas need to be moved early.
https://codereview.chromium.org/71583002

BUG=XWALK-7368


fujunwei:
@halton @lincsoon @mrunalk PTAL.

lincsoon:
LGTM

PR #3952, opened 2016-11-10
[Android] Reflection generator supports the @Deprecated annotation

lincsoon:
When generating the reflection layer, if a method that is marked as
@XWalkAPI is markd as @Deprecated as well, the @Deprecated annotation
will also appears in the new generated wrapper class. This will help to
make the javadoc display more precise method information.

BUG=XWALK-7384

(cherry picked from commit 96e0ad92a7ec9c3e7943b04dd97bd8720505de66)


lincsoon:
@rakuco Please help to merge

PR #3954, opened 2016-11-10
[Android] Use depfiles when generating projects and AARs

rakuco:
Contrary to what one would expect, having a dependency on a shared
library target such as `libxwalkcore` does not cause an action to be
re-run whenever the aforementioned shared library changes -- the .ninja
file only depends on the library's .TOC file, so the action is called
only if the library's symbol list changes.

In practice, this means rebuilding `libxwalkcore.so` would not cause
`xwalk_core_library` and its respective AAR to be regenerated, leading
to all sorts of problems.

We work around the issue by adding `depfile`s to the GN actions, and
then adding those native libraries (as well as some other files like the
.js ones in xwalk_core_library) to them. GN can then use their contents
and trigger the actions when the files there change (in addition to the
ones regularly tracked by the build system).

BUG=XWALK-7433

### DIFF
--- a/build/android/generate_android_project.gni
+++ b/build/android/generate_android_project.gni
@@ -39,10 +39,7 @@ template("generate_xwalk_core_library") {
     forward_variables_from(invoker, [ "deps" ])
 
     script = "//xwalk/build/android/generate_xwalk_core_library.py"
-    inputs = [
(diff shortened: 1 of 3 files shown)


darktears:
rs=me

PR #3957, opened 2016-11-14
Fix some android rebasing issues

msisov:
please take a look


rakuco:
In general we tend to use full declarations instead of `using` clauses; would your patch be much bigger if you did that instead?

msisov:
Got it!

PR #3959, opened 2016-11-25
[Android] Backport presented user data to Crosswalk-23

fujunwei:
CookieManger isn't used at starting crosswalk, so the datas need to be
moved early.
https://codereview.chromium.org/71583002

BUG=XWALK-7368

(cherry picked from commit 22170137e73de48c0750810f663c9d0cf484e6e9)


fujunwei:
@lincsoon

PR #3960, opened 2016-11-28
Implement the XWalkViewDatabase

xzhan96:
This class is the equivalent of Android WebViewDatabase, but it only has
two APIs, hasFormData and clearFormData, they allow developers to determine
whether there is any saved data for web forms and to clear such data.

BUG=XWALK-7057

(cherry picked from commit 2abd648ef7abf88bcfa9ecf62f2e79e7891c9812)


xzhan96:
@lincsoon

PR #3962, opened 2016-12-05
[Android] Do not require certificate transparency

lincsoon:
Certificate transparency is a unique feature of Chrome but non-standard
SSL protocol. It aims to improve the security aloneside with CA. But
Crosswalk is based on content shell so the network connections initiated
by it do not support this certificate policy. This causes a SSL error
called ERT_STATUS_CERTIFICATE_TRANSPARENCY_REQUIRED and makes the app
continue to pop up the "SSL Certificate Error Alert" dialog.

This patch declares that Crosswalk doesn't require certificate
transparency to bypass the SSL error above.

BUG=XWALK-7398

### DIFF
--- a/runtime/browser/runtime_url_request_context_getter.cc
+++ b/runtime/browser/runtime_url_request_context_getter.cc
@@ -33,6 +33,7 @@
 #include "net/http/http_cache.h"
 #include "net/http/http_network_session.h"
 #include "net/http/http_server_properties_impl.h"
+#include "net/http/transport_security_state.h"
 #include "net/proxy/proxy_service.h"
 #include "net/ssl/channel_id_service.h"
 #include "net/ssl/default_channel_id_store.h"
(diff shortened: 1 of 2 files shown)


fujunwei:
Did chromium upstream fixed this issue?
https://codereview.chromium.org/2495583002

xzhan96:
lgtm

lincsoon:
@fujunwei No, the error code in Crosswalk is not same as this issue.

fujunwei:
LGTM

rakuco:
I'd like a more detailed explanation of how Crosswalk and content shell differ from Chromium in this respect. I tried taking a look at this today and couldn't see what sort of "policies" Chromium implements that we do not.

https://m.baidu.com, which was the original URL causing trouble, has a certificate from Symantec, the CA that started the whole CT work upstream. The public key hashes that `net::TransportSecurityState::ShouldRequireCT()` iterates over are different when using Crosswalk/content shell and Chromium (even on Linux), but it is not clear to me why.

In other words: without a bigger explanation, this patch seems to be disabling legitimate checks and fixing the JIRA issue as a side-effect.

rakuco:
> I'd like a more detailed explanation of how Crosswalk and content shell differ from Chromium in this respect. I tried taking a look at this today and couldn't see what sort of "policies" Chromium implements that we do not.

I've also posted this question upstream: https://groups.google.com/a/chromium.org/d/msg/chromium-dev/_ZILH704AVU/ZD0XqL3JCgAJ

rakuco:
... and trying to get additional clarification here: https://bugs.chromium.org/p/chromium/issues/detail?id=669978#c9

rakuco:
Superseded by #3964.

PR #3964, opened 2016-12-13
Stop enforcing Certificate Transparency verification

rakuco:
Chromium started enforcing some Certificate Transparency policies by default
in M53, and refuses to load certain websites when the CT verification fails
while establishing the TLS connection; this affected Crosswalk because we
were using the default CT verifier and policy enforcer, but we did not add
any of the logs approved by the Chromium project (see
net/cert/ct_known_logs_static-inc.h).

Instead of doing so, stop requiring websites to pass Chromium's Certificate
Transparency checks altogether after getting some clarification from
upstream. The biggest impediment to us following Chromium's policy is that
Crosswalk's release cycle would need to match Chromium's very closely to
avoid two risks:
1) Chromium approves new CT logs, they start being used in new certificates,
   Crosswalk does not recognize the new logs and refuses to load a website
   with a valid certificate.
2) Chromium disqualifies a CT log, while Crosswalk does not and can end up
   trusting a certificate that Chromium no longer trusts.

Additionally, the aforementioned list of built-in logs that Chromium uses
stops being recognized 70 days after the build date, which can lead our
users to a false sense of security if they do not update their own apps with
a more recent Crosswalk (which we might not even have released yet).

See the discussion in http://crbug.com/669978 for more information.

RELATED BUG=XWALK-7398

### DIFF
--- a/runtime/browser/runtime_url_request_context_getter.cc
+++ b/runtime/browser/runtime_url_request_context_getter.cc
@@ -25,7 +25,8 @@
 #include "content/public/common/url_constants.h"
 #include "net/cert/cert_verifier.h"
 #include "net/cert/ct_policy_enforcer.h"
-#include "net/cert/multi_log_ct_verifier.h"
+#include "net/cert/ct_policy_status.h"
+#include "net/cert/ct_verifier.h"
 #include "net/cookies/cookie_monster.h"
 #include "net/dns/host_resolver.h"
 #include "net/dns/mapped_host_resolver.h"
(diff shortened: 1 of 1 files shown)


rakuco:
@xzhan96 @fujunwei @lincsoon @darktears 

This achieves the same thing as #3962 in a different way and with a much better explanation.

rakuco:
@xzhan96 @fujunwei @lincsoon @darktears: ping again

xzhan96:
nice, I had seen the discussion in http://crbug.com/669978.
LGTM for this patch.

PR #3967, opened 2016-12-19
[Android] Preserve websql databases when migrating from webview to cr…

fujunwei:
…osswalk

SQLite Databases that are created using the window.openDatabase with
System WebVieware not accessible in Crosswalk, the databases need to
be moved to Crosswalk profile directory.

BUG=XWALK-7368 XWALK-4488


fujunwei:
@lincsoon PTAL

PR #3968, opened 2016-12-19
[Android] Back port the feature of preserving websql databases to Crosswalk 23

fujunwei:
SQLite Databases that are created using the window.openDatabase with
System WebVieware not accessible in Crosswalk, the databases need to
be moved to Crosswalk profile directory.

BUG=XWALK-7368 XWALK-4488

(cherry picked from commit 136cf6e6b6ffdaa308326366f5096e0853fd007b)


fujunwei:
@lincsoon

PR #3970, opened 2016-12-19
If building with DISABLE_BUNDLED_EXTENSIONS

cmeerw:
features will just be marked as experimental so they can be enabled via a command line option instead of completely disabling them.

BUG=XWALK-7461


rakuco:
looks fine to me. @pozdnyakov @lincsoon can you also take a look?

pozdnyakov:
lgtm

rakuco:
@cmeerw: can you squash your 2 commits into one and use a shorter title with the longer explanation after that?

cmeerw:
I tried updating the commit... something broken with the Android x86-64 buildbot setup now?

rakuco:
Hmm, the error was unrelated to the patch and has been fixed. I think we can merge this one, thanks for the patch!

PR #3974, opened 2017-01-23
all temporary storage can be set external storage???

shuaqq2004:
my webapp 's  appcache /localstorage/ indexeddb is used 50MB+.

but  https://developer.chrome.com/apps/offline_storage#temporary say:
temporary storage size is (available storage space + storage being used by apps) * .333 .

so 1GB of storage is actually only 66MB available...
I want set this cache to external storage...
like XWalkSetting.SetAppcachePath("external storage")....

is there a way???


rakuco:
This is not a bug tracker. Please send your question to the crosswalk-help mailing list over https://lists.crosswalk-project.org/mailman/listinfo/crosswalk-help or file a bug at https://crosswalk-project.org/jira.

PR #3975, opened 2017-02-02
loadUrl always opens in new mobile browser(not in the app)

theDude30:
Hi,
Since version 20.50.533.12 i do not have the option to open links in the same window as before(version 19.49.514.5 and below).
I know that the load function is deprecated and now it is recommended to use loadUrl, but i do not know how to make it open in the same app window.

**Here is what i tried:**
 xWalkWebView.load RESULT: opens in new mobile browser.
 xWalkWebView.evaluateJavascript("window.open('"+url+"', '_blank','location=no');",null); RESULT: loads a white screen and it looks stuck.
 xWalkWebView.evaluateJavascript("window.open('"+url+"','_self');",null); RESULT: opens in new mobile window
 xWalkWebView.loadUrl(url); RESULT: opens in new mobile window. 

**The scenario:**
1. in the main activity i opened a url(using loadUrl), the url loaded fine at the application.
2. from that page i clicked on "HTML Button" that calls JavascriptInterface which tries to loadUrl.
3. loadUrl function opened a new browser window outside the application.


rakuco:
This is not a bug tracker. Please ask your question in the crosswalk-help mailing list or file a bug in https://crosswalk-project.org/jira.

PR #3993, opened 2017-06-01
tools: Disable gclient syntax validation when fetching dependencies.

rakuco:
As per
https://groups.google.com/a/chromium.org/d/msg/infra-dev/Cwc6jsitAEA/Dwy3xweSAAAJ,
gclient has recently started validating the syntax of DEPS files.

This is a problem for us because src/xwalk/buildtools, which we fetch from
upstream, is at a revision that does not pass gclient's validation and
breaks the checkout process.

For now, just disable syntax validation to get things back to where they
were before.

BUG=XWALK-7528


baleboy:
LGTM

